PR #291, opened 2014-10-31
First steps to support ocaml version 4.03.0

mbouaziz:
Added support for projects compiled with ocaml 4.03.0-trunk (more or less current trunk).
At least merlin can read .cmi files properly without crashing my emacs.
There's still a bit of work to do to fully handle the new feature of constructors with records (instead of tuples).


trefis:
Hi!
Sorry for the absence of answer on our part.
The code was fine, it's just outdated since we didn't merge it in time.
You can use the trunk branch if you still need to use merlin with trunk (cf #365 )

PR #364, opened 2015-02-26
patches needed to build OpenBSD port

madroach:


def-lkb:
Thanks. I just tweaked your `ifeq "$(NATIVE)"` condition to prefer native build if nothing specified, that shouldn't break anything in your workflow.

PR #366, opened 2015-03-05
Construct

art-w:
This is a preliminary implementation of an inverse Destruct command: given a type hole, it creates an ocaml value.

product

It should happily construct tuples, records, gadts, functions, lazy, modules, functors, objects, first-class modules, and less usefully module signature and class types.

module

Without surprise, while Destruct is mostly usefull for pattern matching on sum types, Construct is better at products. Still, given a sufficiently restricted sum types, it can infer a precise value -- and in case of ambiguity, it stops on the first available choice, to not overflow the user.

gadt

There's an Apply variation that generates the required arguments of a function:

apply_function

This simplifies functor application, which would require a verbose `((val __ : ExplicitSignature))` otherwise:

apply_functor

---
- Emacs plugin: `C-c C-k` for Construct; `C-c k` for Construct Apply.
I haven't manage to integrate with the various completion popups, so in case of multiple choices, the user can press `k` to cycle between the results (and `C-k` to cycle in the reverse order.)
- Vim plugin: `<Plug>(MerlinConstruct)` and `<Plug>(MerlinConstructApply)`, which a user can map to concrete keys.
It does integrate with the vim completion here, but it has a weird behaviour on newlines (which are temporarely replaced with `^@`). 

If the user previously typed an enclosing, the construct takes place on the selected term. The commands accept a `count` argument to specify the search depth (by default 1), so `5<Plug>(Construct)` would attempt to create a value of height 5 in vim (`C-u 5 C-c C-k` in emacs.) The demos above are all done with a large search depth, but it can produce some surprisignly large output.
The vim doc describes the various behaviours in more details.

The error handling/reporting is also very light right now. So, in case you haven't notice, the editors integration is *rough*: I would love some help if you know elisp or vimscript :) !

---
Known bugs and limitations:
- The value printing is done by `Pprintast`, which rarely produces a pretty output (and is sometimes wrong too).
- The generator creates fresh type holes `_x` for incomplete subexpressions and avoids reusing them. However, it doesn't know if an undefined variable is used elsewhere -- and this can have some unexpected influence on the infered type of the hole (but it's only a problem when doing multiple constructs in a row.)
- To avoid reprinting the expression, `Construct_apply` appends the generated arguments: It won't add parenthesis around the full application if needed. It also doesn't check the context, and blindly generate all the arguments (but I'm not convinced that partial application detection is necessary.)
- `class` declaration generates a lot of additional types that should be removed somehow (?).
- The type enclosing computed by merlin is sometimes confusing: `module M : S = _1`, `let f : type a. a = _1`, ... and results in a strange location for the constructed value.
- The builtin types (int, string, ...) default to a choosen value (`0, "", 'c', true, ...`) rather than generating an unhelpfull `_hole`. It seems more convenient, but it's arguably wrong :) .
- I'm not familiar with the merlin code base, so please check for obvious misusages of the caml libraries! Notably, I had to expose `Typemod.modtype_of_package` in the imported caml sources.

I plan to work on this feature a little bit more. But in the mean time, it would be great to have some testing, bugreports, suggestions, and any help you can provide :) .


rgrinberg:
What is that cool font?

gasche:
Wow, this looks quite exciting.

> The value printing is done by Pprintast, which rarely produces a pretty output (and is sometimes wrong too).

Please report any such bugs in the [OCaml bugtracker](caml.inria.fr/mantis/). It's good that you rely on this and we can and will fix these issues.

art-w:
Oh, yes, I was planning on fixing Pprintast or reporting it, as Construct isn't practical when its output requires manual fixing (it's already a small pain with Destruct.)

(@rgrinberg > the font is called Terminus.)

rgrinberg:
@art-w this looks awesome!

Is it possible to have some sort of tabbing functionality when completing the holes. For example in the case where you're completing function arguments.

Some plugins provide this functionality for fixed templates (snippets):

https://github.com/Shougo/neosnippet.vim

https://github.com/SirVer/ultisnips

This might require crazy vimscript hackery though.

trefis:
> Is it possible to have some sort of tabbing functionality when completing the holes. For example in the case where you're completing function arguments.

We already have something for that (i.e. completing function arguments) in the completion-next branch. It also offers completion for labelled and optional parameters as well as polymorphic variants.
This should be merged soon and be included in the next release.


As for the actual PR, I haven't had time to test it or look at the code yet.
I can only answer to : 
> Notably, I had to expose `Typemod.modtype_of_package` in the imported caml sources.

That's fine, we've already done similar things in a few places.

jordwalke:
> > Is it possible to have some sort of tabbing functionality when completing the holes. For example in the case where you're completing function arguments.

> We already have something for that (i.e. completing function arguments) in the completion-next branch. It also offers completion for labelled and optional parameters as well as polymorphic variants.
This should be merged soon and be included in the next release.

I would strongly suggest relying on ultisnips for this. If so, it will integrate well with all of your other snippet commands, without stepping on ultisnips' toes. Also, it's trivial to build on top of ultisnips.
For reference, here's a snippet that prepopulates a function "docblock" with snippet placeholders (and tab-ability):

https://github.com/jordwalke/VimJSDocSnippets/blob/master/plugin/JSDocSnippets.vim

Are you currently using ultisnips for this merlin feature?

PR #367, opened 2015-03-05
trunk: remove utils/tbl.ml{,i}

chambart:
They are already in ocaml_aux. This avoids an inconsistent asumption.


trefis:
Thanks, I hadn't noticed that one.

PR #380, opened 2015-03-21
Unite wrappers around ML, MLI, Use

rgrinberg:
* Require https://github.com/Shougo/unite.vim
* Can be triggered with `:Unite merlin/{ml,mli,use}`

For merlin users who use unite instead of ctrlp

### DIFF
--- a/vim/merlin/autoload/unite/sources/merlin.vim
+++ b/vim/merlin/autoload/unite/sources/merlin.vim
@@ -1,0 +1,51 @@
+let s:source_ml = {
+      \ "name" : "merlin/ml",
+      \ "description" : "Navigate to OCaml module",
+      \ "default_kind": "command",
+      \}
+
+let s:source_mli = {
+      \ "name" : "merlin/mli",
+      \ "description" : "Navigate to OCaml module signature",
+      \ "default_kind": "command",
+      \}
+
+let s:source_findlib = {
+      \ "name" : "merlin/use",
+      \ "description" : "Use findlib module",
+      \ "default_kind": "command",
+      \}
+
+function! s:source_findlib.gather_candidates(args, context)
+  let l:pkgs = []
+  py merlin.vim_findlib_list("l:pkgs")
+  return map(l:pkgs, '{
+        \ "word": v:val,
+        \ "source": "merlin/use",
+        \ "action__command": "MerlinUse ".v:val,
+        \ }')
+endfunction
+
+function! s:source_ml.gather_candidates(args, context)
+  let l:modules = []
+  py merlin.vim_which_ext([".ml", ".mli"], "l:modules")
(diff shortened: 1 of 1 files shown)


trefis:
Hi!

I meant to get back to you earlier for this time, but never took the time.
The main reason why I'm not just click on that big "Merge" button is the following: as you guessed, I'm not using Unite, and I don't think @def-lkb does either. So if we merge this, your code will never be tested by us.

Another thing is that I'm a bit worried about adding support for every unite-like plugin out there, I'd rather present a united front (pun intended) and tell people : "we have support for X, as we find it superior to Y and Z for reasons foo and bar". I don't have any such reason for the moment, when I looked a bit at the options available I chose ctrl-p because "it's pure vimscript" (not that I really care, but I thought it might facilitate adoption). Do you have a compelling reason for choosing unite over the alternatives?
(ping @art-w , I know he tested/looked at a few of these plugins).

Finally, if we do chose to support multiple plugins:
- it would be nice to offer the same commands for each of them
- would it be possible to have them in a separate repository?

One last thing: writing an "interactive" plugin (i.e. ILocate) for ctrl-p was a dreadful experience, would it be simpler with Unite?

rgrinberg:
@trefis Thanks for the consideration, I understand your reluctance to accept this patch. I don't mind having a separate repository for unite+merlin plugins and in fact there are many unite/ctrlp plugins out there so it's not uncommon. I don't really have a preference on the multiple vs one repo point, and I just tried to follow the existing convention.

As for CtrlP vs Unite, I've moved over to Unite myself but I think both plugins are widely used. I'm not aware of any other similar plugins that also allows for customization. A couple of other notes:
* Unite is pure VimScript as well
* FWIW I find unite easier to extend. In my own repo I've also implemented `:Unite merlin/outline` and I find it very much usable.
* I wanted to see some documentation for ILocate before I tried my hand at it in Unite. I will give it a try and report back though

trefis:
Thanks for the quick answer!

For the separate repository, I was worried there might be "namespace" problems when trying to access the python part of our plugin. (Note that "more vimscript, less python" is a solution to this problem, but it's also the problem of the solution). But it's great to know there weren't any such problem.

I think I'll merge your PR if we can get Unite and Ctrl-p to expose the same commands (this will require work on the ctrl-p side :( ).
If you need any help with the "ILocate" command, don't hesitate to ask (but it's pretty simple: synchronize the buffer, call expand_prefix every time the user input changes, call locate on the selected candidate).

If someone else comes to request support for a similar plugin, then we'll move things in a separate repository.
Sounds good?

rgrinberg:
Sounds like a good idea. I'll try my hand at ILocate and polish my outline unite implementation. I don't think there's any other similar plugins (with meaningful number of users) so we're in the clear there.

PR #383, opened 2015-04-01
Drop text properties from command sent to merlin process.

milanst:
Some of the arguments that are sent to merlin process, in particular
buffer-file-name, might contain text properties. This changes how they
are printed and that causes merlin process to choke on them which in
turn creates bunch of "spurious communication" errors in emacs. Just
dropping text properties fixes the issues.


trefis:
Thanks!

Issue #379, opened 2015-03-21
Prefix Merlin Vim commands with Merlin


rgrinberg:
Many Merlin commands are using rather common names which sometimes cause annoying conflicts. For example `Locate` in vim-eunuch. I propose to rename the following list by prefixing Merlin.

* Destruct
* ILocate
* Outline
* Locate
* Reload
* Rename
* Occurences
* Use
* AddFlags
* SetFlags
* ClearFlags

This is not exhaustive and some of these already have proper Merlin aliases, so the old forms have to be removed.

jordwalke:
:+1:

trefis:
Fixed by 1094f99b .

Issue #381, opened 2015-03-23
Conflicting key mappings.


jordwalke:
It seems some of the `merlin` key mappings conflict with other plugins. For example, `<tab>` in visual mode conflicts with the popular `Ultisnips` plugin. Can `merlin` have these disabled by default and provide an example keybindings file that people can opt into via some global flag, or copy/paste into their own keybindings config?

rgrinberg:
+1 Please no default mappings.

def-lkb:
Addressed by @hcwndbyw 's pull request! Thanks.

PR #384, opened 2015-04-01
Rename vim commands that take very common names

rgrinberg:
Destruct, Locate, ILocate, Outline, Occurences, Rename, Reload all now
take a mandatory Merlin prefix

Mitigates #379 somewhat

### DIFF
--- a/vim/merlin/autoload/merlin.vim
+++ b/vim/merlin/autoload/merlin.vim
@@ -479,22 +479,21 @@ function! merlin#Register()
   command! -buffer -nargs=0 YankLatestType    call merlin#YankLatestType()
   command! -buffer -nargs=0 ToggleTypeHistory call merlin_type#ToggleTypeHistory()
 
-  command! -buffer -nargs=0 Destruct call merlin#Destruct()
+  command! -buffer -nargs=0 MerlinDestruct call merlin#Destruct()
 
 
-  command! -buffer -complete=customlist,merlin#ExpandPrefix -nargs=? Locate call merlin#Locate(<q-args>)
-  command! -buffer -nargs=0 ILocate call merlin#InteractiveLocate()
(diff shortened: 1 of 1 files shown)


trefis:
I agree with #379 and #381 in principle, but now that these are there I am a bit worried about changing them/removing them.
I really like the flag idea mentioned in #381 however, but I would like the bindings to be here by default and the command names not to be prefixed by default; so current users who don't have a problem with them and are used to them, are not bothered at all.
(How about `g:merlin_clean_integration`?)

Also, why did you only rename a subset of the commands?

trefis:
Ping @def-lkb btw, what do you think?

rgrinberg:
I've only done it to the ones are already causing me problems and the ones that will surely cause more. The rest are not too pretty but I can live with for now.

I don't like the flag idea because it sounds kind of confusing and is not really the custom for any vim plugins i'm familiar with. If users want shorter names then they should use some [alias plugin](https://github.com/coot/cmdalias_vim) to accomplish that. The breakage this PR introduces is really minor IMO as well so it really doesn't like it's worth it.

rgrinberg:
@def-lkb ?

def-lkb:
There is a tension between new users wanting out-of-the-box configuration and more seasoned one wanting customization. I get #379, #381, #384 are all about this problem.

What about defining an initialization function setting easy defaults, which you can redefine to opt-out/fine-tune?

rgrinberg:
@def-lkb I'm not quite sure how this particular issue is a "new vs experienced" user issue. I don't quite see how prefixing merlin commands would be anymore difficult. I do understand that this is a breaking change and users would potentially have to look at the docs again and remap tweak their key mappings. As intrusive as that is, I think it's a vastly better option than being incompatible with other common plugins out there.

As for making this configurable, I've already stated my reservations against it but I agree that it's still better than the current situation. I'll update the PR accordingly.

trefis:
Superseded by 1094f99b .

Thanks!

PR #409, opened 2015-06-23
Revised commits for Windows/mingw32

dra27:
Finally found some time for upgrading Merlin from the patched 1.6 I'd been using. Patches very similar to those for #194 (some issues had already been fixed or become irrelevant) only as I'm a bit less unfamiliar with git now, I've broken the changes into multiple commits for easier cherry picking if necessary.
An adapted version of this branch also compiles against v2.2 (the whitespace clean-up of vim/merlin/autoload/merlin.py in 8c28e236 means the patch doesn't apply cleaning against v2.2)
I've tested it with gVim 7.4.752 x64 on Windows 7 with Python 2.7.10/3.4.3 for OCaml 4.02.2 i686-w64-mingw32 and it seems to be working beautifully.
a3b240b is critical for building with native Windows ports as ln isn't available
acc4ceb is critical for Vim support on Windows - since reporting that in #194, I've discovered that it is indeed a long-standing bug in Python (and the code is commented accordingly)
Remaining commits, at least with a very current Cygwin, seem to be cosmetic.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+configure eol=lf
+*.sh eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ emacs/merlin.elc
 *.cm?
 *.cm??
 ocamlmerlin
+ocamlmerlin.exe
 ._d
 ._ncdi
 
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ distclean: clean
 
 install-binary: $(TARGET)
 	install -d $(BIN_DIR)
-	install $(TARGET) $(BIN_DIR)/ocamlmerlin
+	install $(TARGET)$(EXE) $(BIN_DIR)/ocamlmerlin$(EXE)
 
 install-share: $(TARGET_EMACS)
 	install -d $(SHARE_DIR)
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ install-share: $(TARGET_EMACS)
 	install -d $(SHARE_DIR)
 	install -d $(SHARE_DIR)/emacs/site-lisp
 	install -m 644 emacs/merlin.el $(SHARE_DIR)/emacs/site-lisp/merlin.el
-	-install -m 644 emacs/merlin.elc $(SHARE_DIR)/emacs/site-lisp/merlin.elc
+	test -f emacs/merlin.elc && install -m 644 emacs/merlin.elc $(SHARE_DIR)/emacs/site-lisp/merlin.elc || true
 	install -m 644 emacs/merlin-iedit.el $(SHARE_DIR)/emacs/site-lisp/merlin-iedit.el
 
 install-vim:
--- a/configure
+++ b/configure
@@ -14,9 +14,15 @@ fi
 
 EXIT=0
 
-COL_CLEAR=$(tput sgr0)
-COL_RED=$(tput setf 4)
-COL_GREEN=$(tput setf 2)
+if which tput > /dev/null 2>&1 ; then
+  COL_CLEAR=$(tput sgr0)
+  COL_RED=$(tput setf 4)
+  COL_GREEN=$(tput setf 2)
+else
+  COL_CLEAR=
+  COL_RED=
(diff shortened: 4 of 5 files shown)


def-lkb:
Looks good (interestingly, I had no problems using the vim current mode on windows, but I can't tell what my exact configuration was).

dra27:
Were you using console vim or gVim? Thining about it, the Python issue would affect gVim only.

def-lkb:
I took some time to play with windows build... An installer is available here:

https://github.com/the-lambda-church/merlin/releases/download/v2.3/Merlin.installer.for.OCaml.4.02.on.Windows.exe

It targets @protz [OCaml on Windows](http://protz.github.io/ocaml-installer/) distribution, but should work as long as ocaml and findlib paths are set.
I have no idea what windows customs are, it might feel a bit alien :).

protz:
I'm thinking of providing opam with my windows installer (see the beta 
version under "other versions"). Any chance you could make merlin build 
"out-of-the-box" with opam on windows?

Here's what I'm getting in the error log for "opam install merlin":

protz@manaslu:~/Code/fstar/runtime/allocator (master) $ cat 
~/.opam/system/build/merlin.2.2/merlin*.err
Bad -I option: src/ocaml/: Invalid argument
make[2]: *** Deleting file '._ncdi/src/frontend/command.di'

Please note that all dependencies of Merlin work fine on Windows.

Thanks,

Jonathan

def-lkb:
I guess it is a problem with symbolic links.
What is the appropriate way to check for windows in the configure?

protz:
Sys.os_type == "win32" ?

dra27:
`Sys.os_type == "Win32"` (capital 'W') or I check for the presence of `$COMSPEC` if trying to determine it before I know if I have OCaml (it's not perfect, but neither is uname in Cygwin!)
It would indeed be symlink-related - I compile elevated with `CYGWIN=winsymlinks:nativestrict` (so I get real symbolic links, not Cygwin's fakery which `ocamlopt`, or in this case `ocamldep`, can't see)

def-lkb:
Then I will fix the Makefile to avoid symbolic links under windows.
Checking for `$COMSPEC` will do it (I would like to stay in the shell script without spawning ocaml... Although we are sure OCaml is available at that point).
May I expect it to always be set in windows?

def-lkb:
Hmm the script is already checking for windows using:

```shell
SYSTEM=$(ocamlfind c -config | grep system|cut -d' ' -f2)                       
 case "$SYSTEM" in                                                               
   mingw*|win*)      
```

@protz is it possible that this check failed on your setup?

dra27:
Yes - Windows really doesn't run without it set (the Command Processor doesn't work and it affects the shell in other unexpected ways). What I've never been utterly certain is whether you can reasonably depend on COMSPEC not being used on any Unix variant.
Incidentally, having done quite a lot of work on symbolic links on Windows recently, it would be good to see the test correctly checking whether ln works, rather than just simply disabling it completely on Windows - see https://github.com/dra27/dose/blob/windows/configure.ac#L265-L273

def-lkb:
I changed the way type checker version is selected in latest master.
It is passed through a Makefile variable rather than a symlink on windows system (as defined by COMSPEC).

Tell me if it improves the build with opam.

protz:
Yes it works now. Supercool! Feel free to point your users to 
http://protz.github.io/ocaml-installer/ and the "WITH OPAM" version.

def-lkb:
Nice, thanks for your work.

First feedback, on http://ocaml.org/docs/install.html#Windows :
This is potentially confusing for a large part of the audience (students).
When you/we get more confident that your distribution is stable, it should be the only one advertised (a single binary solving all problems).
More involved options should be relegated to an "Advanced section".

dra27:
That is indeed the goal - I'm certainly hoping that it will be a case of downloading a very lean binary installer for OPAM only which will then carry out the rest.

protz:
Yes, I'm actually thinking of just distributing the OPAM version of my 
installer and dropping the other options... that's actually less work 
for me to repackage.

~ jonathan

PR #417, opened 2015-07-20
vim: provide mechanism to disable the default keybindings.

hcwndbyw:
I'd like to address #381 and provide a way to disable default keybindings for vim.

Thanks!


def-lkb:
Looks good, thank you!

PR #420, opened 2015-07-29
Add alias for `merlin-send-command-async`

marsam:
Hi:
I've added a compatibility alias for `merlin-send-command-async` which was renamed in 4817aa94fec4abdbbd06f5c57c117078a2e0eaf4, and this function is currently used in `flycheck-ocaml.el`


def-lkb:
Thanks.

Be aware that there will be more breakage because we are tweaking emacs mode to definitely fix concurrency issues, including https://github.com/flycheck/flycheck-ocaml/issues/3 .

marsam:
Good to know. btw, thanks for your work with merlin.

Issue #414, opened 2015-07-02
Feature request: display module/function at current cursor


thranur:
I couldn't find this feature in merlin, so I'm suggesting and asking if it would be feasible:

I'd like to have a command/indicator that tells me in which function/module the cursor is.

It often happens that when reading code inside a function and/or module that is several buffers long (in the middle of which I arrived e.g. by grepping a certain word), I have to scroll up to identify in which exact function I am, and then scroll down again. And if there is just one empty line between two functions, I may miss it and think I'm inside another function.

I know that the notion of "function name" is not as well defined in OCaml as in imperative languages, but all I'd like to know is the topmost `let` and/or `module` I'm inside of.

Is there already something like this?

If not, would it be relatively simple to implement? Or perhaps this is syntactic enough that could/should be handled by Tuareg?

def-lkb:
I don't think this should be handled by Tuareg.
We have partial support for features really close to that (like outline), it shouldn't be much work to integrate that to some future version.

neojski:
@thranur, I'm working on a very similar feature that I currently just call jump. You can have a look [at my jump branch](https://github.com/neojski/merlin/tree/jump). Basically, it introduces a command (right now works only in vim) called `:MerlinJump`. You can give it an argument like `fun,let,match,module` (default) or subset like `match,module`. It'll then jump to the ancestor node that is of this type.

If you have some additional ideas, please share.

def-lkb:
If you are still interested, there is a proof of concept for this feature.
It is not yet working properly because of some assumptions made in the AST breaking the path, but this shouldn't be too hard to fix.
Though, it won't make it to next release,

thranur:
@neojski , sorry I didn't have much time lately to try things, especially because I don't use Vim.

@def-lkb, I'm still interested, though I'm not sure about the easiest way to try it (I believe I should use the command-line `ocamlmerlin` for that, no? There's no binding yet for Emacs, right?). I'll see if I find instructions on how to use it to be able to test the proof of concept.

def-lkb:
@thranur no binding yet, feature is not finished. I will tell you when proof of concept is ready for test.

neojski:
@thranur, the jump feature is not vim specific. It works in emacs as well.

PR #427, opened 2015-09-02
Add jump command

neojski:
This pull request implements a jump command that lets you quickly move to the logical parent nodes in your code.  It is very useful to quickly check the function you're in or even more useful to check the module you're in. One can also jump to let binding and match statement.

It is somehow related to #414.

I'm pretty sure that using `<localleader>j` in vim is a bad idea but I left it for testing. I'll wait for your suggestion.

I imagine people may want different keybinding for different sub commands. Because you can do `:MerlinJump target` where target is a subset of `fun,let,module,match`. For example @bmillwood was interested in just using `:MerlinJump module` and not really `:MerlinJump fun`. So we should make it easy to customize (or maybe just more detailed documentation would be enough for all vim/emacs magicians).

### DIFF
--- a/Makefile.ocamlmakefile
+++ b/Makefile.ocamlmakefile
@@ -195,6 +195,8 @@ SOURCES = \
 	src/analysis/completion.ml        \
 	src/analysis/outline.mli          \
 	src/analysis/outline.ml           \
+	src/analysis/jump.mli             \
+	src/analysis/jump.ml              \
 	$(TYPER)/tast_helper.ml           \
 	src/analysis/destruct.mli         \
 	src/analysis/destruct.ml          \
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1517,6 +1517,33 @@ loading"
            (select-window (display-buffer (car r)))))
     (when r (goto-char (cdr r)))))
 
+;;;;;;;;;;
+;; JUMP ;;
+;;;;;;;;;;
+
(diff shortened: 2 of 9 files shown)


neojski:
I just noticed that in this case:

```ocaml
let f a b =
  true
```

Assertion in `command.ml` is failing. So it's apparently not the case that every fun is immediately followed by pat. I'll investigate.

neojski:
The issue is fixed now.

neojski:
Ping @trefis, @def-lkb

trefis:
We talked about this offline, @neojski will do some minor modifications then I'll merge.

neojski:
I moved the command to a separate file and did the small changes you asked for. I think it's pretty much ready now.

Ping @trefis

Drup:
Hum, documentation ? Also, why not add default bindings ?

trefis:
> Hum, documentation ?

Good one.

> Also, why not add default bindings ?

Because default bindings are evil.
People are always complaining about bindings. Just choose your own!

Drup:
> Because default bindings are evil.
> People are always complaining about bindings. Just choose your own!

I don't disagree, but then you *really* need the documentation. :D

neojski:
@Drup, sorry about the lack of documentation. A rudimentary one you can find in the merlin-jump function description for emacs. I plan to prepare a screencast of this feature soon!

PR #429, opened 2015-09-05
Remove unprefixed versions of command from help

rgrinberg:
Some commands (e.g. `MerlinLocate`) had unprefixed names mentioned as
aliases. That's incorrect and breaks :help (e.g. :h Locate)

### DIFF
--- a/vim/merlin/doc/merlin.txt
+++ b/vim/merlin/doc/merlin.txt
@@ -7,25 +7,25 @@
 ==============================================================================
 COMMANDS                                                     *merlin-commands*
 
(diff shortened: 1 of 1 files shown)


trefis:
The idea was that people wondering where the command all went could call `:h` on the old name and discover the new name.
I understand the "incorrect part", not the "breaks" one. What does it break?

rgrinberg:
Expanding on my example:

`:h Locate` will jump to the help document for merlin's `:MerlinLocate` where I actually have `:Locate` executing the command documented in the following plugin instead:

https://github.com/tpope/vim-eunuch/blob/master/doc/eunuch.txt#L60

trefis:
Got it, thanks.

PR #472, opened 2015-11-25
Fix keywords separator in emacs

neojski:
I changed the separator from comma to space but forgot to change that in `emacs.el`


def-lkb:
Thanks.

PR #474, opened 2015-12-02
[Merlin] Add support for SUFFIX flags.

jordwalke:
Summary: In the `ocamlc`/`ocamlopt` compiler, there is no restriction on
which file suffixes may be used. To use a file with a non-standard
suffix, you would supply the file as either `-intf` or `-impl` (note,
you would also tell the compiler that a `-impl` file has a corresponding
interface in a particular suffix by specifying `-intf-suffix`). However,
until this diff, merlin only searches for files with the `.ml/.mli`
suffixes. This diff adds support for checking other suffixes when
locating files. Merlin's heuristic will sometimes try appending `.ml` to
the end of a module name, etc - this diff allows you to instruct it to
also try other suffixes. It works great!

Test Plan:

Reviewers:

CC:

### DIFF
--- a/src/analysis/track_definition.ml
+++ b/src/analysis/track_definition.ml
@@ -184,10 +184,35 @@ module Utils = struct
     | Longident.Lident _ -> false
     | _ -> true
 
+  let split_extension file =
+    let dirname = Filename.dirname file in
+    (* First grab basename to guard against directories with dots *)
+    let basename = Filename.basename file in
+    try
+      let last_dot_pos = String.rindex basename '.' in
+      let ext_name = String.sub basename last_dot_pos (String.length basename - last_dot_pos) in
+      let base_without_ext = String.sub basename 0 last_dot_pos in
+      let without_ext = Filename.concat dirname base_without_ext in
+      (without_ext, Some ext_name)
+    with Not_found -> (file, None)
+
+
+  let synonym_extension file (implAlias, intfAlias) =
+    match split_extension file with
+      | (without_ext, None) -> without_ext
+      | (without_ext, Some ext) ->
+        if ext = ".ml" then
+          without_ext ^ implAlias
+        else (
+          if ext = ".mli" then
+            without_ext ^ intfAlias
+          else
+            file
+        )
+
   let file_path_to_mod_name f =
     let pref = Misc.chop_extensions f in
     String.capitalize (Filename.basename pref)
-
 
   (* Reuse the code of [Misc.find_in_path_uncap] but returns all the files
      matching, instead of the first one.
@@ -220,7 +245,7 @@ module Utils = struct
     in
     List.map files ~f:Misc.canonicalize_filename
 
-  let find_all_matches ?(with_fallback=false) file =
+  let find_all_matches ~project ?(with_fallback=false) file =
     let fname = Misc.chop_extension_if_any (File.name file) ^ (File.ext file) in
     let fallback =
       if not with_fallback then "" else
@@ -230,10 +255,18 @@ module Utils = struct
       | _ -> assert false
     in
     let path  = Fluid.get sources_path in
-    let files = find_all_in_path_uncap ~fallback path fname in
+
+    let filesList =
+      List.map (fun synonym_pair -> (
+          let fallback = synonym_extension fallback synonym_pair in
+          let fname = synonym_extension fname synonym_pair in
+          find_all_in_path_uncap ~fallback path fname
+        )) (Merlin_lib.Project.suffixes project)
+    in
+    let files = List.concat filesList in
     List.uniq files ~cmp:String.compare
 
-  let find_file_with_path ?(with_fallback=false) file path =
+  let find_file_with_path ~project ?(with_fallback=false) file path =
     let fname = Misc.chop_extension_if_any (File.name file) ^ (File.ext file) in
     let fallback =
       if not with_fallback then "" else
@@ -243,12 +276,30 @@ module Utils = struct
       | File.CMT f  -> Misc.chop_extension_if_any f ^ ".cmti"
       | File.CMTI f -> Misc.chop_extension_if_any f ^ ".cmt"
     in
-    try Misc.find_in_path_uncap ~fallback path fname
-    with Not_found ->
-      raise (File.Not_found file)
+    let rec attempt_search synonyms =
+      match synonyms with
+        | [] -> raise Not_found
+        | [synonym_pair] ->
+          (* Upon trying the final [synonym_pair], search failure should raise *)
+          let fallback = synonym_extension fallback synonym_pair in
+          let fname = synonym_extension fname synonym_pair in
+          (
+            try Misc.find_in_path_uncap ~fallback path fname with
+                Not_found -> raise (File.Not_found file)
+          )
+        | synonym_pair :: ((rest1 :: rest2) as rest_synonyms) ->
+          (* If cannot find match, continue searching through [rest_synonyms] *)
+          let fallback = synonym_extension fallback synonym_pair in
+          let fname = synonym_extension fname synonym_pair in
+          (
+            try Misc.find_in_path_uncap ~fallback path fname with
+                Not_found -> attempt_search rest_synonyms
+          )
+    in
+    attempt_search (Merlin_lib.Project.suffixes project)
 
-  let find_file ?with_fallback file =
-    find_file_with_path ?with_fallback file @@
+  let find_file ~project ?with_fallback file =
+    find_file_with_path ~project ?with_fallback file @@
         match file with
         | File.ML  _ | File.MLI _  -> Fluid.get sources_path
         | File.CMT _ | File.CMTI _ -> Fluid.get loadpath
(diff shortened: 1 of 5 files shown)


def-lkb:
Looks good. Thank you.

trefis:
Actually, it doesn't look good.
Locate can't locate anything outside the current buffer anymore.

trefis:
I reverted, feel free to submit a new pull request with a fixed version when you have the time and I will review it.

def-lkb:
trefis: could you push -f a master version without the pull request merged? I am working on it on experimental-refactor branch. Thinking about it, I should have merged it only there, not on master. Not reverting but just dropping will make future rebasing easier.

trefis:
Done.

jordwalke:
@trefis Just to confirm what you're saying - it would not locate a module outside of the current buffer?

trefis:
For me it wouldn't locate anything outside the current buffer, always complaining that `"foo" seems to originate from module "Bar" which could not be found`.

jordwalke:
I'm certain I tested that and I will confirm again before sending another PR. Thanks for taking the time to test it out.

trefis:
I tested again just to make sure, before the merge commit calling locate on `String.first_double_underscore_end` in Track_definition (there is only one occurence) would bring me to `std.ml` after I get the error `'String.first_double_underscore_end' seems to originate from 'Std' whose ML file could not be found`.

Thanks again for looking at this!

jordwalke:
It's strange, I just tried the exact scenario you described, and it worked fine. (With parent commit being https://github.com/jordwalke/merlin/commit/f1d7d0cd1d17ac093ddb0219cfd46e0289c23e7f)

I'll try rebasing onto current master and see if it still works for me.

jordwalke:
Just rebased this commit onto Merlin `master` and it also works, both for `.ml` locate commands and alternative suffixes as well. (I built with `make debug` so that the `.cmt` files would be available - is this how you built as well?)

Before my commit, only `.ml` files would be located. After my commit, both `.ml` files and any other extension would be discovered as well.

jordwalke:
I could reopen a pull request, but I feel bad wasting your time if nothing has actually changed.

jordwalke:
Any suggestions here? Like I mentioned, I tried this on the latest master branch, and confirmed everything worked correctly. Perhaps @def-lkb could try this out.

trefis:
I haven't had time to look at this yet, sorry.
I'll try to get back to you before the end of the week.

trefis:
Once again I haven't looked in details, just tried the same example as before with the same result...
Here is what I did:
```
$ git reset --hard origin/master
$ git clean -dfx
$ opam switch 4.02.3 && eval `opam config env` && make clean && ./configure --with-bin-annot --prefix ~/.local --with-version-string "402" && make install
$ echo "FLG -debug locate:debug,/tmp/locate.log"
$ vim src/analysis/trac_definition.ml
```
Then I called `:MerlinLocate` on `String.first_double_underscore_end` which worked fine.
Here is the log:
```
$ cat /tmp/locate.log
{"time":42,"level":"info","section":"locate","title":null,"content":"looking for the source of 'String.first_double_underscore_end' (prioritizing .mli files)"}
{"time":42,"level":"info","section":"locate","title":null,"content":"lookup in value namespace"}
{"time":42,"level":"debug","section":"locate","title":null,"content":"present in the environment, but ghost lock.\nwalking up the typedtree looking for 'Std.String.first_double_underscore_end[val]'"}
{"time":43,"level":"debug","section":"locate","title":null,"content":"cursor is in a leaf, so we look only before the leaf"}
{"time":43,"level":"debug","section":"locate","title":null,"content":"resolves to Std.String.first_double_underscore_end[val]"}
{"time":43,"level":"debug","section":"locate","title":null,"content":"Fallback.setopt None"}
{"time":43,"level":"debug","section":"locate","title":null,"content":"from_path 'Std.String.first_double_underscore_end[val]'"}
{"time":51,"level":"info","section":"locate","title":null,"content":"inspecting /usr/local/home/trefis/merlin/src/utils/std.cmt"}
{"time":51,"level":"debug","section":"locate","title":null,"content":"File_switching.move_to /usr/local/home/trefis/merlin/src/utils/std.cmt"}
{"time":52,"level":"info","section":"locate","title":null,"content":"/usr/local/home/trefis/merlin/src/utils/std.ml"}
```
(yes, the logs are noisy... sorry.)

Now if I do the exact same thing except that I reset to your patch applied on master instead of master, vim tells me:

> 'String.first_double_underscore_end' seems to originate from 'Std' whose ML file could not be found

And the log is:
```
$ cat /tmp/locate.log
{"time":44,"level":"info","section":"locate","title":null,"content":"looking for the source of 'String.first_double_underscore_end' (prioritizing .mli files)"}
{"time":44,"level":"info","section":"locate","title":null,"content":"lookup in value namespace"}
{"time":44,"level":"debug","section":"locate","title":null,"content":"present in the environment, but ghost lock.\nwalking up the typedtree looking for 'Std.String.first_double_underscore_end[val]'"}
{"time":44,"level":"debug","section":"locate","title":null,"content":"cursor is in a leaf, so we look only before the leaf"}
{"time":44,"level":"debug","section":"locate","title":null,"content":"resolves to Std.String.first_double_underscore_end[val]"}
{"time":44,"level":"debug","section":"locate","title":null,"content":"Fallback.setopt None"}
{"time":44,"level":"debug","section":"locate","title":null,"content":"from_path 'Std.String.first_double_underscore_end[val]'"}
{"time":55,"level":"info","section":"locate","title":null,"content":"inspecting /usr/local/home/trefis/merlin/src/utils/std.cmt"}
{"time":55,"level":"debug","section":"locate","title":null,"content":"File_switching.move_to /usr/local/home/trefis/merlin/src/utils/std.cmt"}
{"time":55,"level":"debug","section":"locate","title":null,"content":"failed to find \"Std\" in source path (fallback = false)"}
{"time":55,"level":"debug","section":"locate","title":null,"content":"(for reference: fname = \"src/utils/std.ml\")"}
{"time":55,"level":"debug","section":"locate","title":null,"content":"looking in '/usr/local/home/trefis/merlin/src/utils'"}
```
(yes, the log is silly and doesn't indicate an error but... that's ok.)

I hope that helps.

jordwalke:
Thanks, I wonder if the issue is that I'm using `4.02.1` and you seem to be using `4.02.3`.

trefis:
That shouldn't matter. I can try tomorrow if you haven't done so by then.

bsansouci:
Hey,
I just tried patching this PR on master and it seems to work. 
I ran those commands to set it up
```
$ git reset --hard origin/master
$ git clean -dfx
$ opam switch 4.02.3 && eval `opam config env` 
$ git fetch origin pull/474/head:testing
$ git checkout testing
$ make clean && ./configure --with-bin-annot --prefix ~/.local --with-version-string "402" && make install
$ echo "FLG -debug locate:debug,/tmp/locate.log" >> .merlin
$ vim src/analysis/track_definition.ml
```

Then ran `:MerlinLocate` on `String.first_double_underscore_end` and it opened a new buffer with the cursor on the definition of `first_double_underscore_end` inside `src/utils/Std.ml`.

Here are my logs 
```
{"time":112,"level":"info","section":"locate","title":null,"content":"/Users/sansouci/Dropbox/Desktop/merlin/src/utils/Std.ml"}
{"time":197,"level":"info","section":"locate","title":null,"content":"looking for the source of 'String.first_double_underscore_end' (prioritizing .ml files)"}
{"time":197,"level":"info","section":"locate","title":null,"content":"lookup in value namespace"}
{"time":197,"level":"debug","section":"locate","title":null,"content":"present in the environment, but ghost lock.\nwalking up the typedtree looking for 'Std.String.first_double_underscore_end[val]'"}
{"time":198,"level":"debug","section":"locate","title":null,"content":"cursor is in a leaf, so we look only before the leaf"}
{"time":198,"level":"debug","section":"locate","title":null,"content":"resolves to Std.String.first_double_underscore_end[val]"}
{"time":199,"level":"debug","section":"locate","title":null,"content":"Fallback.setopt None"}
{"time":199,"level":"debug","section":"locate","title":null,"content":"from_path 'Std.String.first_double_underscore_end[val]'"}
{"time":206,"level":"info","section":"locate","title":null,"content":"inspecting /Users/sansouci/Dropbox/Desktop/merlin/src/utils/std.cmt"}
{"time":206,"level":"debug","section":"locate","title":null,"content":"File_switching.move_to /Users/sansouci/Dropbox/Desktop/merlin/src/utils/std.cmt"}
{"time":206,"level":"info","section":"locate","title":null,"content":"/Users/sansouci/Dropbox/Desktop/merlin/src/utils/Std.ml"}
```

trefis:
Hm. Seing `Std.ml` while the file is `std.ml` hints that you are using OS X, is that correct? Since I'm pretty sure @jordwalke is also using that system I wonder if the bug might be related (or rather, the absence of bug in your case).
Can one of you try on another system (with a different fs)?

bsansouci:
Oh good call. @jordwalke I just tried on a linux machine and had the same error:
> 'String.first_double_underscore_end' seems to originate from 'Std' whose ML file could not be found

jordwalke:
Ah - good find @trefis, @bsansouci!

bsansouci:
So it's a case-sensitive/insensitive issue right?

jordwalke:
I think it's a bug that needlessly capitalizes file names, and then that bug is masked on OS X because OS X file system doesn't differentiate between casing.

PR #491, opened 2016-02-19
[Merlin] Add support for SUFFIX flags.

bsansouci:
Copy pasted from #474 with a tiny fix.  Tested on a OSX and Linux machines :D

Summary: In the `ocamlc`/`ocamlopt` compiler, there is no restriction on
which file suffixes may be used. To use a file with a non-standard
suffix, you would supply the file as either `-intf` or `-impl` (note,
you would also tell the compiler that a `-impl` file has a corresponding
interface in a particular suffix by specifying `-intf-suffix`). However,
until this diff, merlin only searches for files with the `.ml/.mli`
suffixes. This diff adds support for checking other suffixes when
locating files. Merlin's heuristic will sometimes try appending `.ml` to
the end of a module name, etc - this diff allows you to instruct it to
also try other suffixes. It works great!

Test Plan:

Reviewers:

CC:

### DIFF
--- a/src/analysis/track_definition.ml
+++ b/src/analysis/track_definition.ml
@@ -187,10 +187,35 @@ module Utils = struct
     | Longident.Lident _ -> false
     | _ -> true
 
+  let split_extension file =
+    let dirname = Filename.dirname file in
+    (* First grab basename to guard against directories with dots *)
+    let basename = Filename.basename file in
+    try
+      let last_dot_pos = String.rindex basename '.' in
+      let ext_name = String.sub basename last_dot_pos (String.length basename - last_dot_pos) in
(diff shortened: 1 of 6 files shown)


jordwalke:
:balloon:

jordwalke:
I think this should be good now that Ben rebased on top of the latest Merlin refactor.

trefis:
I tried it and it seems to be working (i.e. it doesn't break the "normal" case and I assume you tested the case with different suffixes).

Thanks!

trefis:
N.B. I also rebased on top of the beta branch and pushed it there as well.

bsansouci:
Thanks! :D

PR #504, opened 2016-03-16
Add ability to supply custom flags to merlin binary from Vim

jordwalke:
Allows specifying of merlin binary flags via a global config in Vim.

@yunxing


trefis:
Thank you!

jordwalke:
Also, I had to add this because I noticed that the following is not respected in the `.merlin` file:

```
FLG -pp mypreprocessor_binary
```

I supposed I'll make another issue for it.

PR #507, opened 2016-03-17
Reverse history

yunxing:
The 'items' list generated by Lexer is actually in reversed order. (e.g., 'EOF' is the head of the list). 

However the result of the logic in "reconstruct_identifier" assumes that the history is not reversed. 

This patch reverses the items before feeding it into the constructor of history. 

I've tested it with "Locate" command, it seems more correct now. 

cc @jordwalke


def-lkb:
I pushed another implementation of the same feature. It should cover this case and also for_completion.

yunxing:
Sounds good.

Issue #294, opened 2014-11-03
Emacs mode, set the mark on merlin-locate


dbuenzli:
I had forgotten about these nice source code navigation features. In the emacs mode, one thing that was disappointing though is that it seems that you don't set the [mark](https://www.gnu.org/software/emacs/manual/html_node/emacs/Setting-Mark.html) before moving to the location. This makes it hard to come back from where you originated (would you do so this would be one natural `C-u-<SPC>` away). Text search in emacs works exactly that way and it feels natural that `merlin-locate` should do the same.

PR #508, opened 2016-03-24
set the emacs mark on merlin--goto-file-and-point

oandrieu:
as discussed in issue #294.

I'm using it and find it quite convenient.


let-def:
Thanks.

PR #522, opened 2016-04-26
Fixed newlines in architecture documentation

chriskonstad:
Now there is a line break between each item instead of them all being on the same line.


let-def:
Thank you.

PR #528, opened 2016-05-10
Fix the 'Unbound value .()' message

chambart:
Quick and dirty version to fix this issue: https://github.com/the-lambda-church/merlin/issues/496


let-def:
Thanks, I made the full fix and synced with upstream parser :).

PR #550, opened 2016-07-07
Configure script is bash specific

juanchanco:
The configure scirpt fails when /bin/sh isn't bash. Changed the shebang line to use "/usr/bin/env bash"

### DIFF
--- a/configure
+++ b/configure
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 BIN_DIR=/usr/local/bin
 SHARE_DIR=/usr/local/share


dbuenzli:
A better move here is to install [`checkbashisms`](https://packages.debian.org/unstable/devscripts) run it and fix the bashisms.

let-def:
[`checkbashisms`](https://packages.debian.org/unstable/devscripts) didn't find anything to complain about.
I also ran the script successfully with `dash`. Could you point which bashisms you found or at least which shell failed you?

juanchanco:
It seems I was too quick to judge, I apologize. The problem actually seems to be with /bin/sh on NetBSD. The error I get is `./configure: 238: Syntax error: Bad substitution` That section is this:

```
if [ -z ${COMSPEC+x} ]; then

  # Non windows-like system, use symlinks

  rm -f "$PWD/src/ocaml" 2>&1 >/dev/null
  ln -sf "$PWD/src/$OCAML_VERSION" "$PWD/src/ocaml"
  rm -f "$PWD/src/sturgeon" 2>&1 >/dev/null
  ln -sf "$PWD/src/$STURGEON_VERSION" "$PWD/src/sturgeon"

else

  # Assume windows-like system
  # Avoid symlinks, find right version through environment variable

  touch "$PWD/src/ocaml"
  MERLIN_CONFIG_ENV="$MERLIN_CONFIG_ENV
export MERLIN_OCAML_VERSION=${OCAML_VERSION/ocaml/}
export MERLIN_STURGEON_VERSION=${STURGEON_VERSION/sturgeon/}
"
fi
```

Apparently /bin/sh is trying to parse the "else" part even if the test succeeds. Should those export lines actually be this (with the appended path after the variable)?
```
export MERLIN_OCAML_VERSION=${OCAML_VERSION}/ocaml/
export MERLIN_STURGEON_VERSION=${STURGEON_VERSION}/sturgeon/
```

Perhaps NetBSD's /bin/sh is less forgiving. This is from the man page:
```
sh is the standard command interpreter for the system.  The current ver-
sion of sh is in the process of being changed to conform with the POSIX
1003.2 and 1003.2a specifications for the shell.  This version has many
features which make it appear similar in some respects to the Korn shell,
but it is not a Korn shell clone (see ksh(1)).  Only features designated
by POSIX, plus a few Berkeley extensions, are being incorporated into
this shell.  This man page is not intended to be a tutorial or a complete
specification of the shell.

```

let-def:
I removed this bashism from configure in git version.
Is this solving your problem?

juanchanco:
Sorry for my lateness, but yes, it builds perfectly from master on NetBSD now!

PR #557, opened 2016-07-29
Implementing the imenu feature for Merlin-Emacs to view code outline 

tddsg:
Hi,

I want to contribute my implementation of the imenu feature to your excellent Merlin-Emacs mode. The reason I implemented this feature is because:
  - The default imenu of Tuareg-mode is quite slow
  - The Merlin-Emacs mode hasn't supported imenu yet, whereas code outline is ready by Merlin

## Description and usage ##

Structure of the merline-imenu module is as follows: 

* Since the code outline returned by Merlin can be big, so I change two thresholds of Emacs, but I'm really not sure about the side effect.

```
(setq max-specpdl-size 10000)
```

* The core function is  `merlin-imenu-create-index`, which create an index function for imenu

* Users can decide which imenu feature that they want to use by executing one of the two functions:
`merlin-use-merlin-imenu` or `merlin-use-tuareg-imenu`. These functions simply assign the indexing function of either `merlin-imenu` or `tuareg-imenu` to the built-in variable `imenu-create-index-function` of imenu-mode

* For my personal use, I add a hook to always enable `merlin-imenu` when switching to tuareg-mode because `merlin-imenu` is really faster than `tuareg-imenu` (see experiment below).

* Imenu can be invoke as usual. For example, I usually use `helm-semantic-or-imenu` to view the list of outline.

## Experiment ##
I tested both the Tuareg-imenu and Merlin-imenu with the module [config.ml] (https://github.com/facebook/infer/blob/master/infer/src/backend/config.ml)
(~1500 lines of code) from the Facebook's Infer tool in my machine (CPU: Core i5 3.2Ghz, RAM: 8GB): 
 - the default Tuareg-imenu spent 4.7s to read the outline
 - whereas the Merlin-imenu took only 0.05s (Thanks to the excellent Merlin-mode)

Source code that printing executing time is commented out in the file `merline-imenu.el`. However, I also added detail comments how to use it.  For example, the detail log can be printed as follows:
(2 lines: Time spent by ...)

```
Merlin: tuareg-imenu is selected, rescanning buffer...
Saving file /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml...
Wrote /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml
Searching definitions...done
** Time spent by Tuareg-Imenu: 4.713094s
Merlin: merlin-imenu is selected, rescanning buffer...
Saving file /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml...
Wrote /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml
** Time spent by Merlin-Imenu: 0.049508s
```

## Illustration ##

Below are illustration in my Emacs with when invoking `helm-semantic-or-imenu` with `merlin-imenu` and `tuareg-imenu`, subsequently:

Note that `meline-imenu` returns more complete outline with 354 candidates, whereas `tuareg-imenu` found only 121.

(I think the reason that `tuareg-imenu` is slower, and less complete than `merlin-imenu` is because the `tuareg-imenu` uses regular expression to find code outline, whereas `merlin-imenu` inherits the entire information found by Ocamlmerlin parser)

### Merlin-imenu ###
imenu-merlin

### Tuareg-imenu ###
imenu-tuareg

Bests,
Trung.

### DIFF
--- a/emacs/merlin-imenu.el
+++ b/emacs/merlin-imenu.el
@@ -53,6 +53,7 @@
                            (t (concat (visit (cdr xs)) " / " (car xs))))))
     (if (null items) "" (visit items))))
 
+;; go to the location of the item
 (defun merlin-imenu--goto-item (line col item)
   (save-excursion
     ;; go to line
@@ -66,6 +67,7 @@
     ;; return the marker
     (point)))
 
+;; the main indexing function
 (defun merlin-imenu-create-index ()
   "Set imenu function"
   (merlin/sync)
@@ -148,6 +150,7 @@
               (push (cons "Misc" misc-list) index))
             index))))))
 
+;; enable Merlin to use the merlin-imenu module
 (defun merlin-use-merlin-imenu ()
   "Merlin: use the custom imenu feature from Merlin"
   (interactive)
@@ -168,6 +171,7 @@
   ;; (imenu--menubar-select imenu--rescan-item)
   (message "Merlin: merlin-imenu is selected, rescanning buffer..."))
 
+;; enable Merlin to use the default tuareg-imenu module
 (defun merlin-use-tuarge-imenu ()
   "Merlin: use the default imenu feature from Tuareg"
   (interactive)
--- a/emacs/merlin-imenu.el
+++ b/emacs/merlin-imenu.el
@@ -73,8 +73,9 @@
          (outline (merlin/send-command `outline)))
     (when outline
       ;; (message outline)
-      (let ((module-list) (value-list) (type-list) (class-list)
-            (exn-list) (misc-list))
+      (let ((module-list) (value-list) (type-list)
+            (class-list) (exception-list) (constructor-list)
+            (label-list) (misc-list))
         (cl-labels
             ((visit-one (prefix x)
                         (let* ((fstart (nth 1 x))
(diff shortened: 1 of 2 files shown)


tddsg:
Hi, do you have time to consider this pull request?

rgrinberg:
How come this was closed?

tddsg:
I thought it was not interesting since there is no response after 3 weeks...

rgrinberg:
I'd attribute that more to the fact that none of the core maintainers are emacs
users (to my knowledge).

In any case, it seems like there's an equivalent feature to this in the vim mode
(`:MerlinOutline`). I've been trying out emacs lately, so I'll volunteer to at
least test this.

One question: do I need helm to test this or does the feature work fine without
it?

tddsg:
Oh, thanks for your interest! If so, I am willing to reopen it. 
Helm is not needed the implementation does not requires it. 
However, please update if you encounter problems when using it without helm.

let-def:
@tddsg: sorry for not answering earlier. We don't have much time right now to work on Merlin issues & PRs, so not much is happening on GH right now.

If all goes well, I will start reviewing PRs & fixing issues in a bit more than a month. Thanks for your contribution & patience :)

tddsg:
Another reason that I closed the request is that I need to modified the module `merlin.el` for other uses. Now, this PR is messed up with other unrelated commits. Do you think that I should create another PR on other branch rather than master to separate them?

let-def:
Branching looks better.

tddsg:
Thanks for the response. I will create a new branch.

tddsg:
I have created a new PR for to separate the emacs-imenu feature. Can this PR be closed?

let-def:
Sure thanks.

PR #565, opened 2016-08-23
Emacs-imenu feature for Merlin

tddsg:
Hi,

As discussed in the PR [#557](https://github.com/the-lambda-church/merlin/pull/557), I created this new request to merge the emacs-imenu feature to merlin.

Please help to test its functionality and discuss if we can improve it further. 
(The initial code may contain some extra code, just for debugging only)

For convenience, **I copy again the description from the previous PR [#557](https://github.com/the-lambda-church/merlin/pull/557):**
## Description and usage

Structure of the merline-imenu module is as follows: 
- Since the code outline returned by Merlin can be big, so I change two thresholds of Emacs, but I'm really not sure about the side effect.

```
(setq max-specpdl-size 10000)
```
- The core function is  `merlin-imenu-create-index`, which create an index function for imenu
- Users can decide which imenu feature that they want to use by executing one of the two functions:
  `merlin-use-merlin-imenu` or `merlin-use-tuareg-imenu`. These functions simply assign the indexing function of either `merlin-imenu` or `tuareg-imenu` to the built-in variable `imenu-create-index-function` of imenu-mode
- For my personal use, I add a hook to always enable `merlin-imenu` when switching to tuareg-mode because `merlin-imenu` is really faster than `tuareg-imenu` (see experiment below).
- Imenu can be invoke as usual. For example, I usually use `helm-semantic-or-imenu` to view the list of outline.
## Experiment

I tested both the Tuareg-imenu and Merlin-imenu with the module [config.ml](https://github.com/facebook/infer/blob/master/infer/src/backend/config.ml)
(~1500 lines of code) from the Facebook's Infer tool in my machine (CPU: Core i5 3.2Ghz, RAM: 8GB): 
- the default Tuareg-imenu spent 4.7s to read the outline
- whereas the Merlin-imenu took only 0.05s (Thanks to the excellent Merlin-mode)

The source code that printing executing time is commented out in the file `merline-imenu.el`. However, I also added detail comments how to use it.  For example, the detail log can be printed as follows:
(2 lines: Time spent by ...)

```
Merlin: tuareg-imenu is selected, rescanning buffer...
Saving file /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml...
Wrote /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml
Searching definitions...done
** Time spent by Tuareg-Imenu: 4.713094s
Merlin: merlin-imenu is selected, rescanning buffer...
Saving file /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml...
Wrote /home/trungtq/workspace/ocaml/infer/infer/src/backend/config.ml
** Time spent by Merlin-Imenu: 0.049508s
```
## Illustration

Below are illustration in my Emacs with when invoking `helm-semantic-or-imenu` with `merlin-imenu` and `tuareg-imenu`, subsequently:

Note that `meline-imenu` returns more complete outline with 354 candidates, whereas `tuareg-imenu` found only 121.

(I think the reason that `tuareg-imenu` is slower, and less complete than `merlin-imenu` is because the `tuareg-imenu` uses regular expression to find code outline, whereas `merlin-imenu` inherits the entire information found by Ocamlmerlin parser)
### Merlin-imenu

imenu-merlin
### Tuareg-imenu

imenu-tuareg

### DIFF
--- a/configure
+++ b/configure
@@ -155,11 +155,6 @@ case "$MAGIC_VERSION" in
     OCAML_VERSION_MESSAGE="OCaml 4.03.0"
     OCAML_VERSION_VAL='`OCaml_4_03_0'
     ;;
-  #"Caml1999I020")
-  #  OCAML_VERSION="ocaml_trunk"
-  #  OCAML_VERSION_MESSAGE="OCaml 4.03 (trunk)"
-  #  OCAML_VERSION_VAL='`OCaml_4_03_trunk'
-  #  ;;
 esac
 
 # Findlib version check
@@ -167,11 +162,6 @@ if [ "x1.5.1" = "x$(ocamlfind query findlib -format '%v')" ]; then
   printf "\nOld version of findlib detected (1.5.1), patching\n"
   filter_file src/kernel/dot_merlin.ml \
      sed -e 's/Findlib\.resolve_path ~base:d ~explicit:true/Findlib.resolve_path ~base:d/'
-fi
-
-if [ -e "src/${OCAML_VERSION}.git/.git" ]; then
-  OCAML_VERSION="${OCAML_VERSION}.git"
-  OCAML_VERSION_MESSAGE="${OCAML_VERSION_MESSAGE} (git version)"
 fi
 
 if ocamlfind ocamlopt; then
--- a/configure
+++ b/configure
@@ -110,7 +110,7 @@ while [ -n "$1" ]; do
   shift 1
 done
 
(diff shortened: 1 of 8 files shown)


let-def:
Thanks. Finally I have time to test your work.
Again, sorry for the delay.

tddsg:
Thanks for the update!

PR #576, opened 2016-09-21
(emacs) Improvement on retrieving type information

tddsg:
Hello,

I did a bit improvement on displaying type information of OCaml expressions in Merlin Emacs and want to contribute back to your excellent project.

The problems I feel inconvenient are:
- when querying for type information by `C-c C-t`, the type information is nicely displayed in the echo area of Emacs, but is immediately cleared when the Emacs cursor moves, or other event happens. 
- The type information can also be looked up in the `type buffer` of Emacs, but this buffer contains only type information but not the corresponding function or expression. 

What I improve are:
- Allow to display the type information in the echo area for a certain amount of time, even when the Emacs cursor moves or the other event occurs. This amount of time can be adjusted by the variable `merlin-type-display-time`. Currently, I set it to 10 seconds. That is, when typing `C-c C-t` to retrieve type information in the echo area, it will automatically disappear after 10 seconds.
- Add an option to allow always display the last retrieved type information in the echo area as long as the user wants. This can be done by invoking the function `merlin/toggle-display-type-buffer`, which enable or disable the continuity of displaying the last retrieved type information.
- Adding also the corresponding expression in the type information. For example, the below picture shows that the type information is not only `var -> var -> Globals.pos -> pure_form` but also include the expression `mk_neq_var`

show-type

I want to you ask for your opinion about these improvements. Do you think they can be useful?

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1190,7 +1190,7 @@ prefix of `bar' is `'."
                    'at (merlin/unmake-point (point)))
              callback-if-success callback-if-exn)))
 
-(defun merlin--print-type-buffer ()
+(defun merlin--display-type-buffer ()
   "Print type buffer to the echo area."
   (let* ((type (with-current-buffer merlin-type-buffer-name
                 (font-lock-fontify-region (point-min) (point-max))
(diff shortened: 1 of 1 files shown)


rgrinberg:
@tddsg would you mind rebasing this pr so I can try it out?

let-def:
Late update: I looked at the patch some time ago and talked with a few people about the change. There was no consensus about whether this was an improvement or not.
In particular, reliance on a timer was pointed out. I don't have a strong opinion as I am not an heavy emacs user.

tddsg:
@rgrinberg : do you mean rebasing to the newest code of merlin?
@let-def : yeah, looking back, I'm not sure it is a significant improvement either. It's just that it can be convenient, but can also complicate merlin's source code

trefis:
I don't think the situation has evolved over the past year, should we close this?

rgrinberg:
Yeah, closing this is fine.

trefis:
Alright, let's do that then.

Thanks for the work @tddsg!

Issue #638, opened 2017-04-27
invalid column number (-1) returned by merlin/outline


tddsg:
Hi there,

It seems that `merlin` sometimes return an invalid column number (-1), which make the feature `merlin-imenu` not working properly. This happens to this file: https://github.com/facebook/infer/blob/master/infer/src/IR/Localise.ml 

Here is the command of Emacs that I send to get the outline ``(merlin/send-command `outline)``. The output is kind of big, hence I attach it here: [outline.txt](https://github.com/ocaml/merlin/files/960744/outline.txt). You can see the column number `-1` appear quite a number of times. 

Could you double check this issue?
Thanks!

let-def:
I guess it is just the derived function which are not given any position by deriving plugins.

tddsg:
Thank you for the information. The `-1` number appear in this structure:
`((assoc) (start (assoc) (line . 1) (col . -1)) (end (assoc) (line . 1) (col . -1)) (name . compare_t) (kind . Value) (children))`

I searched in the source code but didn't see any text `compare_t`. How can `merlin` find it?

There are several relevant texts like: `[@@deriving compare]`, `[%compare.equal : string list]`, `let equal = [%compare.equal : t]`.

Is `compare_t` from one of them?

let-def:
Yes. `ppx_deriving` is explained here: https://github.com/whitequark/ppx_deriving

ppx is a feature which allows external programs to generate OCaml code.
*deriving* is one of these programs which will generate comparison, printing, etc, functions from the a type declaration.
These programs are free to give a location or not to the generated code. `line 1 col -1` is the convention of OCaml compiler for `no location`.

tddsg:
Thanks for the update! In this case, is it feasible to return the location to one of the lines using `ppx_deriving`? To me, this makes more sense than returning `-1`.

let-def:
Merlin cannot do anything about that. This is up to the implementors of ppx.
This even looks like a bug that might be worth reporting (check which ppx is being applied before reporting :))

PR #639, opened 2017-04-27
merlin-imenu: restructure code & handle exceptions

tddsg:
Hi there,

I improve the `merlin-imenu` module of Merlin-Emacs to handle exceptions, like what happens in the issue: https://github.com/ocaml/merlin/issues/638.

I also restructure the source code of `merlin-imenu`, mostly to improve readability. The previous version of `merlin-imenu` is kind of lengthy, since I wrote it when just starting to learn Emacs-Lisp.

Could you please take a look and advise about the changes?

Thanks!


let-def:
Thanks. It looks good, however merlin api has slightly changed. I will update it.

PR #642, opened 2017-05-10
Make merlin-iedit-occurrences even more friendly

infinity0:
This works even if the user is [autoloading](https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html) iedit; the current version fails with "Symbol's value as variable is void".


let-def:
Thank you.

PR #645, opened 2017-05-11
Add Neomake Maker

statianzo:
Modeled after the Syntastic formatter, the Merlin Neomake maker can be
enabled by setting `let g:neomake_ocaml_enabled_makers = ['merlin']`.

I can also update the vim from scratch wiki if this is merged.

### DIFF
--- a/vim/merlin/autoload/neomake/ocaml.vim
+++ b/vim/merlin/autoload/neomake/ocaml.vim
@@ -1,0 +1,8 @@
+function! neomake#makers#ft#ocaml#merlin()
+  let maker = {'name': 'merlin'}
+  function! maker.get_list_entries(jobinfo)
+    return merlin#ErrorLocList()
+  endfunction
+
+  return maker
+endfunction
--- a/vim/merlin/autoload/neomake/ocaml.vim
+++ b/vim/merlin/autoload/neomake/ocaml.vim
@@ -1,8 +1,7 @@
 function! neomake#makers#ft#ocaml#merlin()
-  let maker = {'name': 'merlin'}
+  let maker = {}
   function! maker.get_list_entries(jobinfo)
     return merlin#ErrorLocList()
   endfunction
-
   return maker
 endfunction


let-def:
I don't have Neomake to test but code looks good.
Is there any way to deal with a missing merlin binary? (as in syntastic plugin).

statianzo:
I haven't found any examples in neomake's built in makers that behaves similar to Syntastic's `IsAvailable` function. I could be wrong, as I haven't written many makers. @blueyed, is there anything concerning to you about this?

vim/merlin/autoload/neomake/ocaml.vim
```diff
@@ -0,0 +1,8 @@
+function! neomake#makers#ft#ocaml#merlin()
+  let maker = {'name': 'merlin'}
```
blueyed: I would assume this to be the default name.. but have not checked.
Try just an empty dict here.

vim/merlin/autoload/neomake/ocaml.vim
```diff
@@ -0,0 +1,8 @@
...
+  function! maker.get_list_entries(jobinfo)
+    return merlin#ErrorLocList()
+  endfunction
+
```
blueyed: Nobody needs this newline.. ;)

blueyed:
> Is there any way to deal with a missing merlin binary? (as in syntastic plugin).

We just use `executable()`.

Is this supposed to be a project maker?

I have plans to actually fix file_mode/project makers, but it seems good for now.

What about adding it to Neomake though instead, so that it would get considered when changing this?

let-def:
Either way is fine for me. I merge if you want :)

statianzo:
Thanks for the review @blueyed. I've implemented the suggestions. Because this plugin is dependent on the other vim code in the project, I think keeping it in the merlin project could help prevent versioning conflicts.

blueyed:
OK.

let-def:
Thanks to both of you.

Issue #632, opened 2017-04-08
Function argument placeholders with LSP's snippets


wokalski:
I really like merlin but as an OCaml newbie I don't know function argument count and labels from the top of my head. Many IDEs and vim completion plugins support jumping over arguments which are a great way to solve this issue. (an example at the bottom)

Implementing this functionality from scratch would be a bit of a pain and would probably go beyond the scope of this project. However there are vim plugins like [neosnippet](https://github.com/Shougo/neosnippet.vim) (used in the example), [ultisnips](https://github.com/SirVer/ultisnips), and probably others which handle the jumping part of it. The only thing that autocompletion provider has to do is certain formatting of text in the buffer.

I would like to contribute this functionality. My contribution would just change the text that's inserted after completing a word. The rest is handled by the aforementioned plugins. It would also include an option to enable/disable this feature.

Would you accept it?

![kwi-08-2017 23-43-36](https://cloud.githubusercontent.com/assets/1693190/24832683/475820e4-1cb5-11e7-9615-bb1715b1c2c8.gif)

wokalski:
Here's how it'd work with OCaml.
![kwi-09-2017 00-11-40](https://cloud.githubusercontent.com/assets/1693190/24832835/320f11da-1cb9-11e7-83c0-5d3954087b4b.gif)

let-def:
Yes definitely, this is a valuable feature. Thanks.

copy:
Out of curiosity:

1. Can arguments also be shown as a visual hint? python-jedi, for example, shows it like this:

    ![screenshot_2017-04-09_19-45-00](https://cloud.githubusercontent.com/assets/1816456/24842457/72fd5c0c-1d5d-11e7-98c9-fb2a37cef0dd.png)

2. Is type information available?

wokalski:
1. AFAIK only in the preview window (could you show the full screenshot?).
2. It is available.

What's your idea?

copy:
> AFAIK only in the preview window (could you show the full screenshot?).

It is implemented using setline: https://github.com/davidhalter/jedi-vim/blob/master/jedi_vim.py#L422

> What's your idea?

It would be nice if the types of parameters were shown (either as part of the placeholder or like in python-jedi). Actually, merlin already does this (`g:merlin_completion_arg_type`), but it is shown only in the omni-completion window and just for the current argument.

wokalski:
[...]

PR #653, opened 2017-05-19
Extend function application infos

copy:
This extends the completion query with the following infos:
- The location, name and signature of the function that is being applied
- The current parameter

This is a prerequisite for #632. I'm also working on extending the vim plugin to show the call signature.

screenshot_2017-05-19_12-49-15
screenshot_2017-05-19_12-50-01

Only tested on 4.04 so far, also missing some code for the other version (which is the reason why CI is failing). I would like to get some feedback first.

### DIFF
--- a/src/analysis/completion.ml
+++ b/src/analysis/completion.ml
@@ -566,7 +566,7 @@ let application_context ~verbosity ~prefix path =
   in
   let context = match path with
     | (_, Expression earg) ::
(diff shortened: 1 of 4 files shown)


copy:
I have tested this for some time, and I don't think I can make it work nicely -- too many unuseful type holes, and the vim overlay is still quite buggy.

PR #656, opened 2017-05-22
Fix display of errors in vim

copy:
The current backend doesn't include a `Warning` or `Error` prefix in the message and specifies the type in the record. This fixes the display of errors, which are currently shown as warnings.

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -389,12 +389,11 @@ def vim_loclist(vimvar, ignore_warnings):
     for error in errors:
         if error['type'] == 'warning' and vim.eval(ignore_warnings) == 'true':
             continue
-        ty = 'w'
+        ty = 'e' if error['type'] == 'error' else 'w'
         msg = re.sub(re_wspaces, " ", error['message']).replace("'", "''")
         if msg.startswith("Warning "):
             msg = msg[8:]
         elif msg.startswith("Error: "):
-            ty = 'e'
             msg = msg[7:]
         lnum = 1
         lcol = 1
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -389,7 +389,7 @@ def vim_loclist(vimvar, ignore_warnings):
     for error in errors:
         if error['type'] == 'warning' and vim.eval(ignore_warnings) == 'true':
             continue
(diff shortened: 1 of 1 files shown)


trefis:
Your patch reverts a change which was introduced "recently" (cf. [eb60625](https://github.com/ocaml/merlin/commit/eb6062522c991b68893b25880da64d165d4d95d6)) so this is not simply "a fix" as the stripping of the "warning"/"error" prefix was a deliberate change on our part.

Obviously we can discuss but I personally would much rather not have the prefix present.
The reason being that vim is not that good at displaying long error messages (they are truncated when shown on the command line and unreadable in the loclist because of a lack of line breaks), so I'd rather limit the "noise" in these messages, to facilitate seeing the actual error.

Can I ask why you would like the prefix being displayed?

copy:
> Can I ask why you would like the prefix being displayed?

I don't want the prefix to be displayed, but the prefix isn't there in the first place (and the type detection fails). I recall not seeing the prefix for both errors and warnings, but currently it's there for warnings and not for errors. I'll push a fix just for the type detection and leave the rest as is.


```
  {
    "start": { "line": 36, "col": 8 },
    "end": { "line": 36, "col": 12 },
    "type": "error",
    "sub": [],
    "valid": true,
    "message": "Unbound value alse"
  },
  {
    "start": { "line": 796, "col": 14 },
    "end": { "line": 796, "col": 17 },
    "type": "warning",
    "sub": [],
    "valid": true,
    "message": "Warning 26: unused variable won."
  }
```

copy:
Just to clarify why I think this should be merged: Currently, when using syntastic with merlin, errors are displayed as warnings, because the latest merlin backend already strips the `Error:` prefix. See the screenshot below for reference.

let-def:
I rewrote error message generation to keep track of the error source.
Does the new implementation have the behavior you are looking for?

copy:
@let-def Warnings still don't get picked up by ALE as the type is expected to be an uppercase `w`, not a lowercase one. If you could you change it to uppercase [here](https://github.com/ocaml/merlin/commit/66a6d93c89b22f744bd24ba2c321b91398398af2#diff-7624189940f55fd967cb31b23ef709e0R403), as the vim documentation says (see `:help setqflist()`) I would be happy :-)

let-def:
Better :) ?

copy:
Yeah, thanks!

PR #657, opened 2017-05-30
Skip cppo directives

### LINKED ISSUE #548: Support for .cppo.ml files
The [C Preprocessor for OCaml](http://mjambon.com/cppo.html) seems to be gaining traction in quite a few projects (UTop, ppx_deriving, etc.). However, CPPO directives seem to break Merlin. I'm going through the source right now to try and throw together a quick fix (just ignoring lines that start with a hashtag should be sufficient). Is this sort of thing already in the works? If not, is it a welcome feature?

maelvalais:
In this PR, I propose a way to skip the [cppo] directives of the form:

```
#if OCAML_MAJOR >= 3 \
      && OCAML_MINOR >= 10
#endif
```

Notes:
- I didn't find a way to give the user the possibility to disable this feature. Would it possible to do that?
- I only did the changes to the 4.04 and 4.05 preprocessors, but if this PR gets good feedbacks, I'll also port that to 4.02 and 4.03.
- it might break things when object method are spread among multiple lines (discussed in #548)

To try the changes, checkout this fork and
1. `make preprocess && make`
2. create a file test.ml with the content:
```
#if (FOO > bar)\
    && XXX
print_endline "a";;
#else
print_endline "a";;
#endif
```
3. check the errors with  
      
       ./ocamlmerlin-server single errors -filename test.ml < test.ml

Fixes #548.

[cppo]: https://github.com/mjambon/cppo


raphael-proust:
Support for `cppo` directives should also be documented. (And probably a short mention about the fact that some method names combined with non-standard line breaks will cause problems.)

maelvalais:
Thanks for your insight!
Should I document this in the README? Or just in the code? Both? Or somewhere in the `doc/`?

raphael-proust:
I'd say the `doc/` and the code. Opinions may vary.

By the way, this is also about documenting support for cppo. It was mentioned as a possible future improvement some time ago in the CHANGELOG: https://github.com/ocaml/merlin/blob/master/CHANGELOG#L39

maelvalais:
You are right, the changelog mentionned cppo! I discovered that merlin 2.5.0 introduced the notion of "extensions" (in a separate repo called [merlin-extend](https://github.com/let-def/merlin-extend)).

The hack of the lexer could become an actual support for cppo (instead of just skipping it). I [searched](https://github.com/facebook/reason/search?utf8=✓&q=Extend_protocol&type=) in the reason project and found some examples of merlin-extend usage. But I don't really know where I should start. I am kind of unaware of the internals of merlin.

let-def:
I think something like cppo would be better implemented in merlin itself (at least at the beginning).
The frontend starts in [Mreader](https://github.com/ocaml/merlin/blob/master/src/kernel/mreader.mli), that should be the right place to start a prototype.

let-def:
(I haven't had time to review your PR properly, but it looks like a good start. Integration with Mreader could allow configurations to turn it into a proper preprocessor).

maelvalais:
Hi,

I didn't really have time to work on it; in the meantime, people interested in this small hack can still use it by running

    opam pin add merlin https://github.com/maelvalais/merlin.git#fix/skip-cppo-directives

maelvalais:
Hi,
I won't maintain this PR any longer. I found other ways to do what `cppo` does (as a result of switching from `oasis` to `jbuilder`):
- for the `@VERSION@`, I replaced it with  
     
      (rule
      ((targets (version.ml))
        (action  (with-stdout-to ${@}
                  (run echo "let v = \"${version:my_pkg}\"")))))

   I could also have used jbuilder's subst command but in the end it didn't work as I hoped, but I should retry later (in short, I don't do dist tarballs; I wanted the version thing to work on clone + build but jbuilder subst only seems to work for producing dist tarballs or with opam pin)

- for `#if pkg ... #endif` parts of code (enabling optional features depending on the presence of other libraries), I decided to use @avsm's [technique](https://github.com/mirage/ocaml-conduit/commit/80147aa36a7a5907ededa5d3722f095352d6497a#diff-28d4071c6e41ffbf883231551253f8a1); it requires duplicating code but I think it is way better than my `#if #endif`!

trefis:
@maelvalais you might also be interested in https://github.com/janestreet/ppx_optcomp (an "equivalent" of optcomp, but as a ppx), which should interact well with merlin.

Issue #658, opened 2017-06-02
Segmentation fault if TMPDIR is too long


lpw25:
On linux there is a limit of 108 characters for a socket address. Since `ocamlmerlin.c` uses a socket address based on TMPDIR without checking its length, you can get a segfault if the TMPDIR is too long. I've also observed `Invalid argument` for slightly less long TMPDIR.

let-def:
Is this still happening with current master?

PR #659, opened 2017-06-02
fix openbsd build

xguerin:


let-def:
Hi. Thanks for the patch.

In the light of https://github.com/ocaml/merlin/issues/658, maybe it is better to find a more robust way to generate the socket path.
So I prefer to wait a bit and if we come up with a nicer solution, check that it doesn't break on OpenBSD before merging (Merlin 3 hasn't been released yet anyway).

xguerin:
Ok, thanks. Since I am working on OpenBSD I'll be glad to take care of the testing if needed.

let-def:
xguerin: can you try to rebase on current master? If it works, I will happily merge your PR. Thanks.

xguerin:
Done. However, recent versions of merlin 3.0 are behaving strangely on `OpenBSD`. I am getting the following error when running `ocamlmerlin`:

```
execvp(ocamlmerlin-server): No such file or directory
merlin path: /home/xrg/merlin/ocamlmerlin-server
socket path: /tmp/<not computed yet>
```

I have not taken an in-depth look at what may be the source of the problem (`ocamlmerlin-server` is in my `PATH`). If I find something I'll let you know.

let-def:
I was able to run your patched version on OpenBSD (6.1) without problem, so the problem must a difference in your setup.

xguerin:
How do you call `ocamlmerlin` ? If I call it with its full path, it indeed works. If I call it with just `ocamlmerlin`, it fails.

It seems that `realpath` in `ocamlmerlin.c:378` expands `ocamlmerlin` (`argv0`) into `$PWD/ocamlmerlin`, therefore skipping the `search_in_path` call. Removing line `378` fixes it for me.

It looks like the culprit may be `w0rp/ale`, which calls `ocamlmerlin` without its full path.

However, I don't really see the point of calling `realpath` above. I'm also quite flabbergasted that it would work out of the box on your OpenBSD installation. AFAICT, that call is not biased by any outside environment.

let-def:
I see. My .vimrc for merlin development might workaround the issue.

The purpose of `realpath` is to follow symbolic links, but this approach is fragile, I have to think about something more robust for finding ocamlmerlin-server from ocamlmerlin.

xguerin:
What about checking `realpath` only when `argv0` is actually a path, i.e. skipping it when `argv0 = "ocamlmerlin"` ? That should cover pretty much all corner cases.

PR #667, opened 2017-06-28
check whether ocamlmerlin is called using an explicit path

xguerin:


let-def:
Thanks, it looks good. I will merge and test it for some time.

PR #668, opened 2017-06-28
Vim: add info to help file about using Neocomplete and Deoplete

bluddy:
I often find myself looking for this information, and Deoplete has a different regex syntax, which means you can't just reuse neocomplcache's example. Also, Deoplete and Neocomplete are the current completion engines for Neovim and Vim, respectively.


trefis:
Thanks!

PR #677, opened 2017-07-21
Use colons rather than asterisks for Vim names

dra27:
Internal buffer names are wrapped with asterisks which causes problems on Vim for Windows since it causes the filename globber to be invoked. Using colons works (and also has the benefit of creating an illegal filename, rather than one which is simply unlikely).

Without this change, the call to `bot split *merlin-type-history*` in `merlin_type.vim` causes an error message and the user's buffer gets the type information instead.


trefis:
Thanks!

PR #682, opened 2017-07-25
Fix `\r` problems on recent Cygwins

dra27:
The Cygwin packages for awk, grep and sed were updated on 20 February 2017 so that they no longer automatically strip `\r` characters on binary mounts. Add `tr` calls in various places in configure to get rid of them.

This also fixes the findlib version test, which would never have worked.

References:
  * https://cygwin.com/ml/cygwin/2017-02/msg00152.html
  * https://cygwin.com/ml/cygwin/2017-02/msg00189.html
  * https://cygwin.com/ml/cygwin/2017-02/msg00188.html


dra27:
This patch allows 2.5.5 to compile on Windows with OCaml 4.04.2

PR #684, opened 2017-07-27
Windows supportcd ..

dra27:


let-def:
Thanks, that's an awesome contribution.

> Why is ml_merlin_server_setup implemented in C? (curiosity...)

I recall two reasons, but they might be wrong and are certainly no longer applicable:
- the setup has been refactored, a C function made sense at the time but became less relevant later
- since this function was dealing with system specific things, it was easier to let C handle (and its preprocessor) handle platform specific code and have a single function in OCaml.

> I think there's no buffer overrun check in recv_buffer?

Indeed!

dra27:
Right, three new things:

 1. Builds reliably on MSVC (including on old MSVC - i.e. Windows 7 SDK)
 2. The testsuite is now equivalently failing on Windows as on other platforms
 3. AppVeyor scripts added

dra27:
The AppVeyor stuff warrants a bit of further explanation:

 - It tests 4.02.3, 4.03.0, 4.04.2 and 4.05.0 on all 4 Windows ports (i.e. 16 builds)
 - You obviously can't ever build that many compilers each time. AppVeyor caps execution time to an hour, so it can't even be done once in one go, so the scripts are set so that they incrementally build and cache compilers on each successive run. As long as you don't update a dependency or request a new version, then it converges on a full test. You can see this in action on my own fork:
    + [1.0.97](https://ci.appveyor.com/project/dra27/merlin/build/1.0.97/tests) - the cache had OCaml 4.02.2 and 4.04.1 but the commit requested an upgrade to 4.02.3 and 4.04.2. It therefore built 4.02.3 but skipped 4.04.2 (note Messages tab as well)
    + [1.0.98](https://ci.appveyor.com/project/dra27/merlin/build/1.0.98/tests). AppVeyor was told to build the same commit again - this time it builds 4.04.2 and so tests them all (40 minute build)
    + [1.0.99](https://ci.appveyor.com/project/dra27/merlin/build/1.0.99/tests). Again, AppVeyor was told to restart but this time the caches are all used and the full test on all 16 compilers takes 17 minutes, rather than 40.

Obviously, AppVeyor needs to be enabled for ocaml/merlin for this to be of any use!

dra27:
@let-def - I'm not sure when I'll have time to work on the other 4 issues, but none of them are blockers. An option could be to review all this code as one now, and file 4 issues to track these other items? I also haven't done any testing of emacs

bluddy:
If merlin works on windows again, I recommend releasing a new version on opam asap.

dra27:
Possibly, although `opam pin add merlin --dev-repo` should work just fine too.

kkirstein:
Just tried it on my Windows box (Win10-64bit, rel 1703, cygwin environment, mingw toolchain). After `opam pin add merlin --dev-repo`, merlin compiles out of the box. Vim integration works, but atom & vscode not (atom thows exception, vscode silently does nothing).
IMO, releasing would be o.k., but the protocol change of merlin 3.x still causes some problems. I will have a look at the respective vscode project...

dra27:
Thanks @kkirstein! It'd certainly be good to sync up an actual release with support for other editors (I haven't tested emacs on Windows, either).

In principle, atom and vscode should be working in legacy mode, though it would obviously be good to switch them over to the new protocol - let me know if I can help!

bluddy:
> since this function was dealing with system specific things, it was easier to let C handle (and its preprocessor) handle platform specific code and have a single function in OCaml.

@let-def It would be nice to put a to-do item to move the server to OCaml, and sort out whatever needs doing (perhaps in stdlib or some other library) so that others can benefit from / rely on this work.

kkirstein:
@dra27 It looks like that legacy mode does not proper work with the current atom & vscode extensions. I wanted to approach the respective projects with this issue, but unfortunately I am quite busy these days...

PR #687, opened 2017-07-28
Added missing include of alloca.h.

bschommer:
For clang one needs to include the alloca.h header file to make alloca available.

### DIFF
--- a/src/platform/os_ipc_stub.c
+++ b/src/platform/os_ipc_stub.c
@@ -3,6 +3,7 @@
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
+#include <alloca.h>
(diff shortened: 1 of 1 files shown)


let-def:
Thank you.

let-def:
Is this preventing any build on OSX? (That is a minor update is necessary :P)

bschommer:
I'm not sure, I'm just using clang under linux as replacement for an old system gcc.

bschommer:
A colleague tried merlin 3.0.0 on a mac and it compiled without problems.

let-def:
Yes I just tried the same, it worked well too. Anyway it's fine to include alloca even if it works as it is.
Thanks.

PR #696, opened 2017-08-11
FIX Support compiling on FreeBSD

orbitz:
This gets merlin to compile on FreeBSD.  I don't think it's probably the most portable solution (other BSD's might need to be covered).

### DIFF
--- a/src/platform/os_ipc_stub.c
+++ b/src/platform/os_ipc_stub.c
@@ -3,7 +3,11 @@
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
+
+#ifndef __FreeBSD__
 #include <alloca.h>
+#endif
+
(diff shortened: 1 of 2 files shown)


orbitz:
@let-def Poke

src/platform/os_ipc_stub.c
```diff
@@ -3,7 +3,11 @@
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
+
+#ifndef __FreeBSD__
```
bschommer: I would suggest testing for linux instead.

orbitz: Whoops, too late!  Hopefully it won't be a problem or it'll be clear to the next person what to fix.

Issue #607, opened 2016-11-25
investigate Language Server Protocol for Merlin


avsm:
https://github.com/Microsoft/language-server-protocol/wiki/Protocol-Implementations

There seems to be a sophisticated [OCaml implementation](https://github.com/freebroccolo/ocaml-language-server) here as well, that already interops with Merlin.  Should this be mentioned and/or tested more with Merlin?

let-def:
It is worth mentioning. I will ask @freeboccolo what he thinks of merging some features inside Merlin.

freebroccolo:
@let-def I would be happy to see language server support become part of merlin somehow.

I think there are two aspects to consider here:

1. merlin speaking the language server protocol
2. merlin supporting the features specified by the language server capabilities

For (1), this might be the right direction to move merlin in the long term. I'm not sure if it's a good idea for now though, at least not as the primary interface. The problem is that while there are several plugins for language server protocol support being developed for various editors, most of them except for vscode are not yet in a usable state.

But for the long term, it may make sense. Using JSON-RPC for communication with merlin seems fine to me and there's no problem adding custom notifications or requests specific to merlin on top of what the language server protocol already specifies.

For (2), there are some features I've built on top of merlin in the ocaml language server in order to support the full server capabilities specified by the protocol. I think the main ones are:

1. code action provider
2. code lens provider
3. workspace symbol provider

The code action feature (1) just analyzes certain error messages from OCaml (via merlin) and suggests various fixes. Some of these fixes can be applied automatically. The typical case would be like a simple syntax error due to a missing delimiter or something like that. But they don't have to be errors either. They could just be suggestions to fix certain kinds of warnings too. Having merlin be able to support a feature like this directly would be nice.

The code lens feature (2) allows the editor to display information above some text range. This can be used for all sorts of things but what I have done with it in the language server is display types:

![code-lens](https://cloud.githubusercontent.com/assets/151197/20631485/ff4bf51c-b2f3-11e6-91b8-b77901c2f3a3.png)

The feature is implemented by making an outline request to merlin, using the response to find the ranges of let bindings, and then asking merlin for the types. This could be simplified if merlin could also provide the types of structure items in the outline response or if it supported some other special code lens command.

[...]

PR #701, opened 2017-08-16
[RFC] Minimal quickfixes support

Kakadu:
At the moment only 1 type of quickfix (a.k.a. code action) is supported: when the user misspelled something a compiler can correct it using the environment.

I'm not sure that I placed the code into right place, btw. I also changed the output JSON format which is not great.

See also: https://github.com/ocaml/merlin/issues/607

Demo:
quickfix

### DIFF
--- a/src/frontend/query_commands.ml
+++ b/src/frontend/query_commands.ml
@@ -504,6 +504,8 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
     let lexer_errors  = Mpipeline.reader_lexer_errors pipeline  in
     let parser_errors = Mpipeline.reader_parser_errors pipeline in
     let typer_errors  = Mpipeline.typer_errors pipeline  in
+    let quick_fixes  = Mpipeline.typer_quickfixes pipeline  in
+
     (* When there is a cmi error, we will have a lot of meaningless errors,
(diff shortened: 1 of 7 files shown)


trefis:
Apologies for not having looked at this earlier, reviewing resources were scarce.

On principle this looks great!
I haven't looked in details at the implementation yet, but there is a fairly obvious issue with this PR in its current state (which you apparently already guessed as you somewhat mentioned it in your description): you're changing the protocol in a non retro-compatible way, without updating the various editor modes included in this repository.

andreypopp:
I really like this. I think this PR should be revived and added to LSP (at least initially).

klntsky:
If I were to reimplement this feature, would it be merged?
Are there any suggestions on how to do it properly?

let-def:
@klntsky: Yes, no problem to merge this feature (if the implementation is satisfying). I don't know how to approach that though.

The protocol should of course be updated in a retro-compatible way (adding new fields to a json dictionary is retro-compatible). At least one of editor should have complete integration (Emacs, Vim or LSP), and it would be better to have hint about how the feature could be integrated to other editors. 
Aslo the feature should have no noticeable impact on performance. (Maybe it should be opt-in with an appropriate flag).

PR #709, opened 2017-08-27
Pass include paths to ppx

keigoi:
Some ppx ([ppx_implicits](https://bitbucket.org/camlspotter/ppx_implicits), for example) uses include paths to refer to other files. The paths are propagated from `ocamlc`/`ocamlopt`'s `-I` parameter to the ppxs. However, current merlin seems not to simulate it well.

This fix will set `Clflags.include_dirs` before calling ppxs (and restores them after call). 

In `Pparse.apply_rewriters`, [`Ast_mapper.add_ppx_context_str` is called](https://github.com/ocaml/merlin/blob/d27a113ad6a103f2679c81706eb2427e74ac8882/src/ocaml/support/pparse.ml#L94) to append a `[@@ocaml.ppx.context]` attribute which includes that paths.

### DIFF
--- a/src/kernel/mppx.ml
+++ b/src/kernel/mppx.ml
@@ -11,6 +11,24 @@ let change_directory dir =
     false
 
 
+let with_include_dir path f =
+  let saved = !Clflags.include_dirs in
+  let restore () = Clflags.include_dirs := saved in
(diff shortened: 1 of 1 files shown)


let-def:
Thanks, that looks good.

PR #711, opened 2017-08-28
Make the *merlin-types* buffer read-only

Chris00:


let-def:
That looks good, thank you.

Issue #686, opened 2017-07-28
look for .merlin in _build?


vogler:
Since jbuilder generates the .merlin file, it would be nice if it could be moved to _build.
Is there something that speaks against looking for it in _build if it's not found in the root?
Alternatively one could also parse the jbuild file but I don't know if that has all the needed information.

rgrinberg:
The approach that we have in mind will require no .merlin files at all.

PR #712, opened 2017-08-29
look for .merlin in _build

vogler:
Closes issue #686 
After looking for `dir/.merlin` this also tries `dir/_build/.merlin` before going up the path.

### DIFF
--- a/src/kernel/mconfig_dot.ml
+++ b/src/kernel/mconfig_dot.ml
@@ -113,14 +113,14 @@ module Cache = File_cache.Make (struct
   end)
 
 let find fname =
-  if Sys.file_exists fname && not (Sys.is_directory fname) then
+  let is_file fname = Sys.file_exists fname && not (Sys.is_directory fname) in
+  if is_file fname then
     Some fname
   else
     let rec loop dir =
-      let fname = Filename.concat dir ".merlin" in
-      if Sys.file_exists fname && not (Sys.is_directory fname)
-      then Some fname
-      else
+      (* first try dir/.merlin, then dir/_build/.merlin *)
+      try Some (List.find ~f:is_file Filename.([concat dir ".merlin"; concat (concat dir "_build") ".merlin"]))
+      with Not_found ->
         let parent = Filename.dirname dir in
         if parent <> dir
         then loop parent


let-def:
Thanks. I am not opposed to the idea, bug different approaches have been suggested, especially for interop with jbuilder.
@vogler which build system are you using?

vogler:
jbuilder 😄 See [jbuilder/#207](https://github.com/janestreet/jbuilder/issues/207).
Just seemed like the easiest fix for now.
Which other approaches? Using `jbuild` directly would of course be even better.

rdavison:
@let-def just curious, what other approaches are you referring to?

let-def:
@rdavison I was thinking of supporting jbuild directly or provide an interface for plugging external build system plugins support.

The current way Merlin gets build information is too limited for practical uses nowadays and .merlin are an artifact of the past (... they were good enough when they were introduced, but the ecosystem started moving forward). Supporting jbuild is good for jbuilder but there are other systems out there.

I am thinking of a more general "log file" listing all compiler invocations in which Merlin pick the corresponding ocaml(c/opt) line but this is not the end of the story either (build systems tend to copy files, so the compiler view alone is too limited for Merlin).

bschommer:
How about using something similar to the [clang compilation database](https://clang.llvm.org/docs/JSONCompilationDatabase.html). This works for pretty well for C/C++ tooling with clang.

mroch:
how about an `INC` config? so you could do `INC _build/.merlin` to get the behavior of this PR, and I could do something like `INC ../../../buck-out/gen/.merlin` to make this work with [Buck](https://buckbuild.com).

mroch:
that is, rather than special-casing the resolution algorithm for each build system, continue to rely on the existing resolution algorithm. once it finds a .merlin file, that file can include the generated one from wherever it wants.

ELLIOTTCABLE:
It sounds like over-engineered intermediate solutions are a bad idea? The proposed change in the OP is a great hold-over, while something like clang's database can be hashed out with the build-tool people as the plan moving forward?

(I'm a newcomer to OCaml and have no idea what I'm talking about, so feel free to ignore me if there's a good reason for it — just been burned by too many hold-off measures with too many caveats in the past! 🤣)

PR #719, opened 2017-09-16
Fix emacs package header formatting

purcell:
This fixes the malformed header in `merlin.el`, which has been preventing `package.el` (and therefore MELPA) from reading package metadata from the file.


let-def:
Thank you.

PR #723, opened 2017-09-30
use case sensitive comparison vim

gregory-nisbet:
VimL comparison operators without a trailing `#` (case-sensitive) or `?` (case-insensitive) inherit the user's current settings.

Specifying a particular case sensitivity does not affect the behavior when the arguments are integers.


let-def:
Your patch looks good but I am not versed enough in VimL to be sure everything is fine. I will merge it and revert if something goes wrong.

However:
1) For future code, I get that normal comparisons operators is considered bad style in VimL?
2) Is this working on older versions of vim (say 7.0)?

PR #724, opened 2017-10-01
vim: Pass end of error location to enable highlighting of errors

copy:
If this is omitted, only the first character of an error is highlighted, since the error plugin (ALE in this case) doesn't know the end location.


let-def:
Thanks!

Issue #726, opened 2017-10-01
Use asynchronous completion in Vim 8 / Neovim


edwintorok:
Same as https://github.com/ocaml/merlin/issues/725 but for Vim.

Vim 8 and Neovim now support asynchronous job execution, and there are several plugins for asynchronous completion that support both editors(e.g. [asyncomplete](https://github.com/prabirshrestha/asyncomplete.vim) or [nvim-completion-manager](https://github.com/roxma/nvim-completion-manager)).
They can be used with merlin as is, however calling `omnifunc` is always done synchronously (even in [deoplete](https://github.com/Shougo/deoplete.nvim), it would be nice if there was an async merlin completion source that these plugins could use (with the current blocking omnifunc as fallback for Vim 7.x).
See here for an example: https://github.com/prabirshrestha/asyncomplete-flow.vim/blob/master/autoload/asyncomplete/sources/flow.vim

With the default config recommended by the merlin docs Vim/Neovim with `neocomplete/deoplete` is in fact a lot slower than Emacs because it invokes the completion over and over for each character typed in `hosts` (and setting the various delay parameters of the plugins doesn't help because it only delays displaying the popup, not calling omnifunc).
If the pattern is modified to trigger only on the dot then it is somewhat faster but the UI still freezes for several seconds:
```vim
if has("nvim")                                                                                                                                                
    let g:deoplete#enable_at_startup=1                                                                                                                        
    if !exists('g:deoplete#omni_input_patterns')                                                                                                              
      let g:deoplete#omni#input_patterns = {}                                                                                                                 
    endif                                                                                                                                                     
    let g:deoplete#omni#input_patterns.ocaml = '[^. *\t]\.'                                                                                                   
else                                                                                                                                                          
   let g:neocomplete#enable_at_startup = 1                                                                                                                    
    if !exists('g:neocomplete#force_omni_input_patterns')                                                                                                     
        let g:neocomplete#force_omni_input_patterns = {}                                                                                                      
    endif                                                                                                                                                     
    let g:neocomplete#force_omni_input_patterns.ocaml = '[^. *\t]\.'                                                                                          
    let g:neocomplete#auto_complete_delay = 50                                                                                                                
endif     
```

copy:
I'm working on a plugin for this, I will send a PR probably this week.

copy:
I released this as a standalone vim plugin: https://github.com/copy/deoplete-ocaml

edwintorok:
[...]

PR #728, opened 2017-10-05
Add plugin for deoplete

copy:
Deoplete provides asynchronous completion via neovim's rpc api.

This implements #726.

### DIFF
--- a/merlin.install
+++ b/merlin.install
@@ -17,6 +17,7 @@ share: [
   "vim/merlin/ftplugin/ocaml.vim" { "vim/ftplugin/ocaml.vim" }
   "vim/merlin/ftplugin/omlet.vim" { "vim/ftplugin/omlet.vim" }
   "vim/merlin/plugin/merlin.vim" { "vim/plugin/merlin.vim" }
+  "vim/merlin/rplugin/python3/deoplete/sources/deoplete_ocaml.py" { "vim/rplugin/python3/deoplete/sources/deoplete_ocaml.py" }
   "vim/merlin/syntax_checkers/ocaml/merlin.vim" { "vim/syntax_checkers/ocaml/merlin.vim" }
   "vim/merlin/syntax_checkers/omlet/merlin.vim" { "vim/syntax_checkers/omlet/merlin.vim" }
   "vim/merlin/syntax/merlin.vim" { "vim/syntax/merlin.vim" }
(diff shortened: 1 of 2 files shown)


rlbaker:
I tried this out locally and was wondering if maybe the regex for plugin is a bit overeager compared to the deoplete regex recommended by [merlin itself](https://github.com/ocaml/merlin/blob/a2facd4bb772ee0261859382964c30e8401866e8/vim/merlin/doc/merlin.txt#L352).

Plugin: `[^\s\'"()]*$`
Merlin: `[^. *\t]\.\w*|\s\w*|#`

I'd love to see this get integrated though, as async vim completion for merlin would be great.

copy:
I released this as a standalone vim plugin: https://github.com/copy/deoplete-ocaml

It would still be nice to have this integrated into merlin, but I'm guessing merlin's authors don't want to deal with the extra maintenance burden.

> I tried this out locally and was wondering if maybe the regex for the plugin is a bit overeager compared to the deoplete regex recommended by merlin itself.
> 
> Plugin: `[^\s\'"()]*$`
> Merlin: `[^. *\t]\.\w*|\s\w*|#`

If you have a test case where this breaks completion, [let me know](https://github.com/copy/deoplete-ocaml/issues/new). I believe merlin completion works even if you provide a longer prefix than necessary, but I haven't tested this thoroughly.

ELLIOTTCABLE:
I'm a little confused about this: Deoplete is explicitly mentioned in `:help merlin`; but the only relevant commit seems to be the one adding that (see #668). Meanwhile, this pull-request was never merged …

So is, or isn't, Deoplete explicitly supported by Merlin right now? If so, what does your plugin add, @copy? Some information about this in the docs would be very welcome.

copy:
@ELLIOTTCABLE By default merlin installs an omnifunc for completion, which works in vim without plugins, but is also picked up by deoplete if no other sources are available ([source](https://github.com/Shougo/deoplete.nvim/blob/master/doc%2Fdeoplete.txt#L615)). However, completion via omnifunc is run on the main thread, so if merlin takes one second to get the completion entries the ui is frozen for that time. See also the related issue: #726.

In order to get proper asynchronous completion I implemented a deoplete plugin. Merlin's authors didn't show much interest (I'm assuming due to the extra maintenance burden of maintaining a plugin they wouldn't use), so I released it as a separate plugin and will maintain it [there](https://github.com/copy/deoplete-ocaml). It uses merlin's api, so it doesn't necessarily need to be integrated into merlin anyway.

Issue #710, opened 2017-08-28
UTF-8 documentation should work


Chris00:
When calling `merlin-document` (in Emacs) on a function whose documentation contains UTF-8 symbols, the message or buffer displayed contains "garbage" instead of the original symbols.

let-def:
Should be fixed in master. Can you test?

Chris00:
It is not complete.  It seems to output the right bytes as copying the message gives
```
(* Computes cos(x) for x ∈ ]-2⁶³, 2⁶³[. *)
```
but in Emacs I see (even though the buffer is in UTF-8):
![emacs-type](https://user-images.githubusercontent.com/1255665/29898442-1621813a-8de6-11e7-87d2-564ff0c1572f.png)

Chris00:
Don't know whether it is related but sometimes Merlin report types such as
```
b:(float -> float) ->¯loat ->
speed:float ->¯loat ->
(float, Bigarray.float64_elt, Bigarray.fortran_layout) Bigarray.Array1.t
```
(notice the `¯loat` instead of `float`).

let-def:
This is strange and I cannot reproduce. On my system I get proper utf-8 support. Maybe this has to do with some default.

What is your locale (and operating system)?
Also would you mind trying to investigate from your side as you are probably more knowledgeable on Emacs lisp than I am :) (I will help of course).

Chris00:
The first problem is fixed with the above commit.

For the second one, the funny thing is that the first time I request the type it prints `b:(float -> float) -> float -> speed:float -> float -> Odepack.vec` while — without moving the cursor — I request it a second time, it prints `b:(float -> float) ->¯loat ->
speed:float ->¯loat ->
(float, Bigarray.float64_elt, Bigarray.fortran_layout) Bigarray.Array1.t`.  Both are correct but the second one has the `¯loat` problem (this is not a unibyte/multibyte problem, it rather seems to come from Merlin).

Chris00:
Running with `(setq merlin-debug t)` yields (truncated):
```
# calling binary: "/home/trch/.opam/4.04.0/bin/ocamlmerlin" with arguments: ("server" "type-enclosing" "-protocol" "sexp" "-log-file" "-" "-verbosity" "2" "-filename" "/home/trch/research/Feltrin/superlinear-sign-changing/ocaml/ode.ml" "-position" "12:5" "-index" "0").
# stdout
[...]

PR #738, opened 2017-10-16
Return doc as a multibyte string

Chris00:
Fixes the documentation problem reported in
https://github.com/ocaml/merlin/issues/710


let-def:
This is now fixed I think?

Chris00:
Yes, thanks to https://github.com/ocaml/merlin/pull/758

let-def:
Good. Thanks again for your work!

PR #751, opened 2017-11-15
Standard q behavior for types buffer in emacs

MiloDavis:
Currently, Merlin does not follow the Emacs convention of having the `q` key quit a pop-up buffer. This is a bit of a hack, so I'm willing to rework it into having a proper key map for the types buffer that contains the binding.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -958,6 +958,13 @@ An ocaml atom is any string containing [a-z_0-9A-Z`.]."
            (setq count (1+ count)))
     (<= count 8)))
 
+(defvar merlin-types-buffer-map
+  (let ((map (make-sparse-keymap)))
+    (set-keymap-parent map special-mode-map)
+    (define-key map "g" nil)
+    map)
+  "Keymap for types buffer.")
+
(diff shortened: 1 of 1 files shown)


trefis:
I think I once upon a time heard rumors of a special kind of buffer (which you perhaps call a special function to create, or there's just some property to set when creating them or something) that do indeed disappear when you hit q. (something like http://www.gnu.org/software/emacs/manual/html_node/elisp/Temporary-Displays.html perhaps?)

Would you mind having a look at that? It seems like it would be "the right way" to do things.

Thanks for taking the time to work on this, it has indeed been somewhat "broken" for years :)

MiloDavis:
Sure. I'll look into it and update the PR.

MiloDavis:
I've looked into things a bit. Emacs defines `special-mode` for this sort of thing, but it's a major mode so we'd lose the syntax highlighting that merlin currently has. The solution used by the emacs compilation mode is to extend the keymap from special mode. I've done this and pushed a new version of the patch.

trefis:
OK!

trefis:
Thanks for the changes!

PR #757, opened 2017-12-10
Fix Neomake autoload

bobbypriambodo:
I couldn't get the Neomake integration working, and it would seem that the problem was:

1) The `.vim` file was not included in `merlin.install`
2) Neomake expects filetype makers to be inside `autoload/makers/ft` directory

I've tried only solving 1 but it didn't work, and only after 2 is fixed that it works perfectly.

I also added `EnabledMakers` function so that users wouldn't need to add any configuration to get it working. Let me know if this is considered too intrusive (previously we need to add `let g:neomake_ocaml_enabled_makers = ['merlin']` to `.vimrc`).

Also pinging @statianzo as the original author.

### DIFF
--- a/merlin.install
+++ b/merlin.install
@@ -6,6 +6,7 @@ bin: [
 share: [
   "vim/merlin/autoload/ctrlp/locate.vim" { "vim/autoload/ctrlp/locate.vim" }
   "vim/merlin/autoload/ctrlp/outline.vim" { "vim/autoload/ctrlp/outline.vim" }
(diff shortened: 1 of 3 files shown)


let-def:
I don't know neomake, but the commits looks good to me. I suggest waiting for @statianzo feedback before merging.

Thanks.

let-def:
Thanks to you two!

bobbypriambodo:
Thank you!

PR #758, opened 2017-12-11
Emacs: Interpret the "value" returned by Merlin as multibyte

Chris00:
Fixes https://github.com/ocaml/merlin/issues/710

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -511,7 +511,7 @@ return (LOC1 . LOC2)."
   "Execute a command and parse output: return an sexp on success or throw an error"
   (let ((result (merlin--call-merlin command args)))
     (condition-case err
-        (setq result (car (read-from-string (string-as-multibyte result))))
+        (setq result (car (read-from-string result)))
       (error
         (error "merlin: error %s trying to parse answer: %s"
                err result)))
--- a/src/utils/sexp.ml
+++ b/src/utils/sexp.ml
@@ -32,19 +32,10 @@ let escaped str =
     Buffer.add_string buf str
   ) else (
     for i = 0 to len - 1 do
-      match str.[i] with
-      | '"' ->
+      let c = str.[i] in
+      if c = '"' || c = '\\' then
         Buffer.add_char buf '\\';
-        Buffer.add_char buf '"'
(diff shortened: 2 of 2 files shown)


Chris00:
Right — I just hit that myself.  Pushed a new version.  Are there other cases/values to handle (don't know all the possibilities that Merlin can return)?

let-def:
Yes, that's worrying.
Either:
- we recurse to rewrite deep in the sexp
- I(we) remove escaping of characters other than " and \\
- I(we) follow multibyte escaping convention

Chris00:
If we go for the [unicode escaping](https://www.gnu.org/software/emacs/manual/html_node/elisp/Non_002dASCII-in-Strings.html) convention, we need to check that other editors support it as well.  This would assume the initial string is valid UTF-8 and would require correctly [identifying the UTF-8 chars](https://en.wikipedia.org/wiki/UTF-8#Description).  (It would be nice to know which past problems you encountered with escaping.)  Failing that, just escape `"` and `\`.  Which one would you like to try—I'll modify this PR accordingly?

Chris00:
Note that the current escaping scheme does not escape `\`.

let-def:
Pff, thanks for pointing that '\' is not yet escaped :). Shame on me.

Just go with escaping '\' and '"'.
I don't remember which problems I had with escaping (the one I remembered had to do with parsing propertized string, quite far from what we are dealing now, ...), but it can't be worse than current situation.
This is simpler (and it is ok to assume that everything is UTF-8 valid, and if it is not, it's not merlin job to deal with that, garbage in, garbage out?). And this code path is only used for emacs, no need to bother about whether other editors are going to suffer from this patch.

let-def:
(I am really not good at dealing with emacs stuff, maybe the modes should be separate project...)

Chris00:
> I am really not good at dealing with Emacs stuff

Don't worry, things like this may happen to anyone!

> maybe the modes should be separate project...

IMHO, this should not be the case as the two must evolve hand in hand.

Chris00:
> Just go with escaping '\\\\' and '"'.

Done.

let-def:
Thanks!

Issue #729, opened 2017-10-10
Allow the use of fzf.vim instead of CtrlP.vim


ELLIOTTCABLE:
I've switched from fzf.vim to CtrlP.vim for non-OcamL files; I'd like to be able to drop CtrlP entirely.

For other visitors, here's my vim-plug line to restrict CtrlP to OCaml files, since Merlin currently depends on it:

```viml
Plug 'ctrlpvim/ctrlp.vim', { 'for': ['ocaml'] }
```

Not a very high-priority request, though! `<3`

trefis:
I don't mind in principle, but am personally not a user of fzf (and neither is @let-def I believe), and don't really have time to investigate and write this new plugin.

Would you mind sending a PR?

ELLIOTTCABLE:
My OCaml is new/shaky, but I'll take a look `(=`

trefis:
Don't worry, for this part you should only need VimL/Python :)

andreypopp:
There's PoC https://github.com/andreypopp/fzf-merlin (don't know if it works still but I used it quite successfully)

PR #768, opened 2017-12-30
RFC: Denite.nvim locate-support (as an alternative to Ctrl-P)

ELLIOTTCABLE:
This is just my first stabs at a proof-of-concept; it's not even working, much less ready for merge!

Simply posting as an invitation for comment or feedback. Further work is forthcoming; I've run into [a problem with how `import` works inside Denite plugins](https://github.com/Shougo/denite.nvim/issues/403#issuecomment-354528911).

Largely cribbed from [denite-git](https://github.com/neoclide/denite-git) and [denite-rails](https://github.com/5t111111/denite-rails/blob/master/rplugin/python3/denite/source/rails.py), along with the relevant parts of [`autoload/ctrlp/locate.vim`](https://github.com/ocaml/merlin/blob/master/vim/merlin/autoload/ctrlp/locate.vim#L9-L34).

(Also see: #729.)

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -319,17 +319,17 @@ def command_occurrences(pos):
     except MerlinExc as e:
         try_print_error(e)
 
-def command_expand_prefix(base, pos, kinds=[]):
+def command_expand_prefix(base, pos, kinds, calculate_types):
     try:
         kinds = concat_map(lambda kind: ("-kind",kind), kinds)
         if pos is None:
-            args = ["expand-prefix",
-                    "-position", fmtpos(vim.current.window.cursor),
-                    "-prefix", base] + kinds
-        else:
-            args = ["expand-prefix",
-                    "-position", fmtpos(pos),
-                    "-prefix", base] + kinds
+            pos = vim.current.window.cursor
+
+        args = ["expand-prefix",
+                "-position", fmtpos(pos),
+                "-prefix", base,
+                "-types", "y" if calculate_types else "n"] + kinds
+
         l = command2(args)
         return l['entries']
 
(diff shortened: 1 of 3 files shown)


ELLIOTTCABLE:
Okay! After much flailing, I've basically got something cobbled together. The selection-action of actually jumping to the location is blocked on a bug in Denite (see: Shougo/denite.nvim#411); but here's what it looks like so far:

screen shot 2018-01-10 at 00 21 22

I also want to add improved highlighting groups (if I can figure out a way, maybe highlight them using the existing OCaml grammar?); and change the default Denite keybinds, settings, and actions.

----

One thing though — I'm making good progress on the Denite version of `:MerlinILocate`, but I'm a little lost as to the purpose of `:MerlinOutline`. Isn't that what `ocp-index` is for? I think? Genuine question, because I'm very new to OCaml. If it's for a different purpose, I'll dive into implementing a Denite-based outline next!

ELLIOTTCABLE:
(Still here, still working on a resolution of Shougo/denite.nvim#411.)

copy:
Thanks for this, I will definitely try it when the critical issues are resolved.

I ran into similar issues regarding importing from merlin while developing [deoplete-ocaml](https://github.com/copy/deoplete-ocaml) (deoplete is from the same author as denite and uses the same asynchronous plugin structure). I ended up not importing merlin directly, and instead [calling merlin's api](https://github.com/copy/deoplete-ocaml/blob/master/rplugin/python3/deoplete/sources/deoplete_ocaml.py#L89-L114) while respecting the same configuration variables as merlin. The major problem with importing `merlin.py` is that it contains a lot of global state, which would exist multiple times if merlin was imported from deoplete's Python instance (as it runs in a separate process from the neovim process where merlin lives).

Issue #770, opened 2018-01-04
vim_next_enclosing breaks when already at the root of the document


ELLIOTTCABLE:
In Vim, given the following document, with the cursor on the `1`,

    let i = 123 in

An initial `<LocalLeader>t` yields, as expected, `int` — but a following `<LocalLeader>n` to select the “next enclosing” type-signature spits up the following Python error:

```
Error detected while processing function merlin#GrowEnclosing[1]..provider#python#Call:
line   18:
Traceback (most recent call last):

  File "<string>", line 1, in <module>

  File "/Users/ec/Documents/Code/Source/merlin/vim/merlin/autoload/merlin.py", line 705, in vim_next_enclosing
    return vim_current_enclosing()

  File "/Users/ec/Documents/Code/Source/merlin/vim/merlin/autoload/merlin.py", line 695, in vim_current_enclosing
    tmp['type'] = enclosing_type_text(tmp)

  File "/Users/ec/Documents/Code/Source/merlin/vim/merlin/autoload/merlin.py", line 686, in enclosing_type_text
    return types[record['type']]['type']

IndexError: list index out of range
Error detected while processing function merlin#GrowEnclosing:
line    2:
E121: Undefined variable: l:type
E116: Invalid arguments for function <SNR>142_ShowTypeEnclosing
```

ELLIOTTCABLE:
Oop, accidentally a duplicate of my own #732. 🙄

PR #771, opened 2018-01-04
Fixing vim's handling of the enclosing-type cache

### LINKED ISSUE #732: MerlinGrowEnclosing / LL-n explodes if I attempt to grow the selection too much
After `<LocalLeader>n`'ing enough times, I always get something like the following:

screen shot 2017-10-10 at 13 09 51

(Forgive the screenshot, but I can't quite figure out how to copy error-messages! 🤣)

 - Merlin **3.0.2**
 - NeoVim **v0.2.1-331-g6650588c**

ELLIOTTCABLE:
The usage of the `enclosing_types` cache in the Vim plugin is pretty bugged — fixes #732, #770.

This is my first PR, so I intentionally went a bit overboard with explanatory comments; I also tried to match the code-style while moving things around.

I won't be offended at all if you have any feedback on how you want this cleaned up — feel free to throw any requirements at me if you want this massaged! `<3`

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -675,24 +675,43 @@ def enclosing_tail_info(record):
     return ''
 
 def enclosing_type_text(record):
+    global enclosing_types
+
+    # The server has an undocumented functionality where it still returns *all*
+    # enclosing nodes when the `type-enclosing` command is passed `-index` (this
+    # is contrary to the documentation of the protocol); with only the reqested
+    # element having an actual type-string attached. The remaining elements do
+    # not have their type calculated (which *is* in line with the protocol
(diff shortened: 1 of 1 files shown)


ELLIOTTCABLE:
Anything blocking this? `(=`

trefis:
> Anything blocking this?

Yes, the lack of people available for reviewing :p 

So, I just had a look, and it took me a while to see what the actual change was, that's because the diff is bigger than it needs to: why did you inline the function instead of living in on its own?
If you hadn't, the change would have been more obvious.

Apart from that nitpick I think I understand and agree with your change.

If you can undo the inlining and squash all your commits then I'll happily merge this.

trefis:
Thank you!

PR #775, opened 2018-01-16
fix escape sexp

### LINKED ISSUE #744: Windows/Emacs: escaping of \
merlin-switch* and similar functions don't work for me on Emacs (Windows only).
Path separators (`\`) are always messed up and emacs therefore fails to open the according files.
If I escape `\` inside https://github.com/ocaml/merlin/blob/1f9a49822d29f1a7fb2dbc59881cfe4622f0b3c5/src/utils/sexp.ml#L19, the problem disappears.
But perhaps the culprit is the counterpart on the emacs side or something else ...

fdopen:
The for loop was not updated in #758
Therefore backslashes are usually not escaped. This should finally fix #744

### DIFF
--- a/src/utils/sexp.ml
+++ b/src/utils/sexp.ml
@@ -21,9 +21,7 @@ let escaped str =
   let extra_chars = ref 0 in
   for i = 0 to len - 1 do
     match str.[i] with
-    | '"' -> incr extra_chars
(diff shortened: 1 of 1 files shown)


let-def:
Indeed, this looks good. Thanks. Sorry for the delay!

PR #776, opened 2018-01-22
Merlin xref

### LINKED ISSUE #735: Please make `merlin-locate/pop-stack` aware of xref
It would be great if `merlin-locate` could integrate with standard emacs [xref](https://www.gnu.org/software/emacs/manual/html_node/emacs/Xref.html) instead of shipping its own solution.

rgrinberg:
Attempt to fix #735

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ EMACS_OBJECTS = emacs/merlin.elc \
 							  emacs/merlin-imenu.elc \
 							  emacs/merlin-ac.elc \
 							  emacs/merlin-cap.elc \
-							  emacs/merlin-company.elc
+							  emacs/merlin-company.elc \
+							  emacs/merlin-xref.elc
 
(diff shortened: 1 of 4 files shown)


rgrinberg:
@ejgallego one problem that remains to be solved in this PR is how to handle the symbol at point properly. xref expects the mode to extract the prefix that needs to be completed, while merlin just wants to know the position. Currently I just ignore the `symbol` from xref. But that's probably not right.

let-def:
Thanks, the patch looks good.

Can you add the new elisp file to `EMACS_OBJECTS` in `Makefile`?

rgrinberg:
Did you mean `TARGET_EMACS`? Anyway, I think I updated the makefile and .install file correctly now.

trefis:
Ping @let-def : apparently you've already reviewed this and your request was addressed, do you want to merge this?

trefis:
Actually, this looks fine to me.

PR #782, opened 2018-02-06
[autoload/vim] Fix umlaut issue in filepaths on manjaro

Tehlyria:
Umlauts in filepath led to errors after switching from Ubuntu to Manjaro while loading the plugin.

    UnicodeEncodeError: 'ascii' codec can't encode character u'\xdc' in position 34: ordinal not in range(128)

Encoding the ```fnames``` fixes said issue.


let-def:
Thanks.

PR #784, opened 2018-02-09
Fix v3 performance regression

### LINKED ISSUE #779: Large v3 performance regression (and proposed solutions)
At Facebook, we upgraded to Merlin 3, but encountered a huge performance regression in doing so. To typecheck https://github.com/facebook/hhvm/blob/master/hphp/hack/src/typing/typing.ml went from ~2s to ~6s. I timed how long it took to typecheck this file for each previous version of Merlin and tracked down two commits, each of which increase the typechecking time by about 2s each.

The commits are b4ce455 (~2s -> ~4s) and 68a0095 (~4s -> ~6s).

[You can see the raw timing data here](https://gist.github.com/arxanas/17eb45513f66ad10afbf609864a0b2f9). Each line is the commit hash followed by the benchmarked total elapsed time, generated with `/usr/bin/time ./ocamlmerlin single errors -filename /path/to/typing.ml </path/to/typing.ml`. I double-checked around the problematic commits in question, and ran the timing several times, but there's no question that these two commits are the causes. It is possible that my use of `ocamlmerlin single` is not representative of typical IDE use, though.

Why do these commits cause performance regressions? Here's my analysis of the first problematic commit. I didn't look into the second commit very much.

Here's a profile from `perf` before b4ce455:

```
24.09%  ocamlmerlin-ser  [kernel.vmlinux]    [k] __d_lookup_rcu
 6.65%  ocamlmerlin-ser  [kernel.vmlinux]    [k] avc_has_perm_noaudit
 6.50%  ocamlmerlin-ser  [kernel.vmlinux]    [k] selinux_inode_permission
 5.94%  ocamlmerlin-ser  [kernel.vmlinux]    [k] link_path_walk
 3.90%  ocamlmerlin-ser  [kernel.vmlinux]    [k] lookup_fast
 3.39%  ocamlmerlin-ser  [kernel.vmlinux]    [k] walk_component
 2.31%  ocamlmerlin-ser  [kernel.vmlinux]    [k] generic_permission
 2.02%  ocamlmerlin-ser  [kernel.vmlinux]    [k] security_inode_permission
 1.97%  ocamlmerlin-ser  [kernel.vmlinux]    [k] inode_permission
 1.95%  ocamlmerlin-ser  [kernel.vmlinux]    [k] __inode_permission
 1.95%  ocamlmerlin-ser  [kernel.vmlinux]    [k] strncpy_from_user
 1.81%  ocamlmerlin-ser  libc-2.17.so        [.] _int_malloc
 1.64%  ocamlmerlin-ser  [kernel.vmlinux]    [k] creds_are_invalid
 1.64%  ocamlmerlin-ser  [kernel.vmlinux]    [k] in_group_p
 1.49%  ocamlmerlin-ser  libc-2.17.so        [.] __memmove_ssse3_back
 1.43%  ocamlmerlin-ser  [kernel.vmlinux]    [k] do_syscall_64
 1.30%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlFilename__concat_1330
 1.11%  ocamlmerlin-ser  libc-2.17.so        [.] __xstat64
 1.06%  ocamlmerlin-ser  ocamlmerlin-server  [.] mark_slice
 1.05%  ocamlmerlin-ser  libc-2.17.so        [.] __memcpy_ssse3_back
 0.99%  ocamlmerlin-ser  libc-2.17.so        [.] __strlen_sse2_pminub
 0.96%  ocamlmerlin-ser  [kernel.vmlinux]    [k] avc_has_perm_flags
 0.90%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_alloc_string
 0.82%  ocamlmerlin-ser  libc-2.17.so        [.] _int_free
 0.82%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlPervasives__$5e_1117
 0.78%  ocamlmerlin-ser  ocamlmerlin-server  [.] sweep_slice
 0.72%  ocamlmerlin-ser  [kernel.vmlinux]    [k] security_inode_getattr
 0.59%  ocamlmerlin-ser  [kernel.vmlinux]    [k] selinux_is_enabled
 0.50%  ocamlmerlin-ser  [kernel.vmlinux]    [k] kmem_cache_alloc
```

As you can see, much of the time is spent on the filesystem. Commit b4ce455 intends to add a layer of caching for the filesystem. But does it improve performance? Here's the output from `time` for the first bad commit:

```
2.46user 1.41system 0:03.91elapsed 99%CPU (0avgtext+0avgdata 149596maxresident)k
1608inputs+0outputs (0major+36902minor)pagefaults 0swaps
```

And here's the output from `time` for the last good commit:

```
0.42user 1.70system 0:02.13elapsed 99%CPU (0avgtext+0avgdata 51024maxresident)k
0inputs+0outputs (0major+12068minor)pagefaults 0swaps
```

So we did reduce system time by a little bit (-0.3s), but dramatically increased CPU time (+2s). Presumably your results will vary depending on what type of system you have; in our case, we use SSDs for our development work, so filesystem access may be faster than whatever this commit was tested on.

Why did this happen? Let's look at the `perf` profile:

Here's the `perf` profile:

```
10.97%  ocamlmerlin-ser  [kernel.vmlinux]    [k] __d_lookup_rcu
 7.45%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlString__needs_escape_1313
 7.15%  ocamlmerlin-ser  ocamlmerlin-server  [.] mark_slice
 5.61%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlHashtbl__find_1406
 4.05%  ocamlmerlin-ser  ocamlmerlin-server  [.] compare_val
 3.89%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_hash_mix_string
 3.26%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_page_table_lookup
 3.22%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_string_length
 2.89%  ocamlmerlin-ser  [kernel.vmlinux]    [k] selinux_inode_permission
 2.84%  ocamlmerlin-ser  libc-2.17.so        [.] __memmove_ssse3_back
 2.81%  ocamlmerlin-ser  ocamlmerlin-server  [.] sweep_slice
 2.36%  ocamlmerlin-ser  [kernel.vmlinux]    [k] avc_has_perm_noaudit
 2.09%  ocamlmerlin-ser  [kernel.vmlinux]    [k] link_path_walk
 1.83%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_hash
 1.80%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlFilename__concat_1330
 1.74%  ocamlmerlin-ser  libc-2.17.so        [.] __memcmp_sse4_1
 1.48%  ocamlmerlin-ser  [kernel.vmlinux]    [k] lookup_fast
 1.21%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_fl_allocate
 1.19%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlHashtbl__insert_bucket_1368
 1.11%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_oldify_one
 1.01%  ocamlmerlin-ser  [kernel.vmlinux]    [k] walk_component
 0.98%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_alloc_string
 0.97%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlFilename__is_dir_sep_1247
 0.93%  ocamlmerlin-ser  [kernel.vmlinux]    [k] inode_permission
 0.70%  ocamlmerlin-ser  [kernel.vmlinux]    [k] __inode_permission
 0.69%  ocamlmerlin-ser  [kernel.vmlinux]    [k] generic_permission
 0.63%  ocamlmerlin-ser  [kernel.vmlinux]    [k] strncpy_from_user
 0.59%  ocamlmerlin-ser  [kernel.vmlinux]    [k] do_syscall_64
 0.58%  ocamlmerlin-ser  [kernel.vmlinux]    [k] clear_page_c_e
 0.58%  ocamlmerlin-ser  libc-2.17.so        [.] _int_malloc
 0.57%  ocamlmerlin-ser  [kernel.vmlinux]    [k] security_inode_permission
 0.56%  ocamlmerlin-ser  [kernel.vmlinux]    [k] in_group_p
 0.54%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlCamlinternalFormat__strput_acc_64848
 0.54%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_modify
 0.52%  ocamlmerlin-ser  ocamlmerlin-server  [.] caml_apply2
 0.50%  ocamlmerlin-ser  ocamlmerlin-server  [.] camlFilename__base_1238
```

We're spending a tad less time on filesystem-related functions, but a lot more time in the garbage collector, hashing strings, and manipulating strings (such as with `Filename.concat`). After looking at this for a while, my conclusion is that these strings are the filenames that we're using as cache keys, and operations like `Filename.basename`, `Filename.dirname`, and `Filename.concat` are being called too often and causing serious memory pressure and using a lot of CPU time. This may not manifest in a smaller codebase, but we do have a few hundred or thousand OCaml files in the repo, and their paths are on the order of 100 characters long.

I'm not exactly sure what the second commit does, but it looks like it's related to filesystem operations as well.

What can we do about it? Some solutions:

* **Revert the two commits**. This works in my testing and drops the benchmark time back down to 2s.
* **Opt out of these behaviors using per-project configuration**. I couldn't figure out how to implement this. In particular, simply changing the file caching strategy from `Stat_dir_cache` to `Stat_fs` everywhere did not seem to affect the performance, and I didn't understand why.
* **Switch to using a faster filename manipulation implementation**, to reduce memory pressure and speed up common operations like basename, dirname, and concat, as well as hashing. One can envision a structure where each component of the path is stored in a (persistent) linked list, and to manipulate these we need only push/pop things onto the linked list or append two linked lists. Since we can reuse most of the linked list nodes, and don't have to allocate more strings for these operations, it should reduce CPU time and memory pressure significantly. We can also store the hash alongside this structure to make hashing operations fast.

arxanas:
Fixes #779. This just reverts b4ce455 and 68a0095. As far as I can tell it doesn't impact correctness, and significantly improves performance on low-latency filesystems.


let-def:
`without_cmis` is not correct. Is this patch significantly impacting performance in your case?
The stat cache is definitely wrong, so I will try to diagnose what is wrong about it.

arxanas:
This isn't a huge issue for us anymore, as per my comment in https://github.com/ocaml/merlin/issues/779. Thanks for looking into it, though!

let-def:
Current master includes a reworked implementation of the caches. I am curious if it brings any performace improvement to your case.

Thanks.

arxanas:
Before (3 runs):

```
0.21user 0.08system 0:00.30elapsed 98%CPU (0avgtext+0avgdata 55060maxresident)k
0inputs+0outputs (0major+13113minor)pagefaults 0swaps
0.21user 0.08system 0:00.29elapsed 99%CPU (0avgtext+0avgdata 55132maxresident)k
0inputs+0outputs (0major+13118minor)pagefaults 0swaps
0.22user 0.06system 0:00.29elapsed 97%CPU (0avgtext+0avgdata 55096maxresident)k
0inputs+0outputs (0major+13119minor)pagefaults 0swaps
```

After (3 runs):

```
0.13user 0.06system 0:00.19elapsed 97%CPU (0avgtext+0avgdata 51632maxresident)k
0inputs+0outputs (0major+12242minor)pagefaults 0swaps
0.14user 0.05system 0:00.20elapsed 100%CPU (0avgtext+0avgdata 51564maxresident)k
0inputs+0outputs (0major+12244minor)pagefaults 0swaps
0.12user 0.06system 0:00.19elapsed 97%CPU (0avgtext+0avgdata 51660maxresident)k
0inputs+0outputs (0major+12241minor)pagefaults 0swaps
```

So it looks like a ~33% performance improvement, which is always welcome :)

let-def:
Thanks for the quick feedback :)

PR #791, opened 2018-03-09
Cleaner handling of emacs match data

rgrinberg:
* prefer string-match-p over string-match when we don't need match data
* use save-match-data to restore match data just in case. This is done a bit defensively in some cases, but it causes no harm to leave it in.


trefis:
LGTM thanks!

PR #792, opened 2018-03-09
Speed up some tight loop in Emacs

rgrinberg:
It's sad to undo the beautiful functional programming here, but elisp is not such a performant language for FP. Therefore, I replaced the tight loops in completion functions with the `cl-loop` macro. It's not that ugly, and leads to a decent boost in completion speed boost with company mode on my system (10% in cases where it's noticeable). This is especially important as it affects the typing latency if you have auto-completion on.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -874,15 +874,14 @@ prefix of `bar' is `'."
     (cons prefix suffix)))
 
 (defun merlin--completion-prepare-labels (labels suffix)
-  ; Remove non-matching entry, adjusting optional labels if needed
-  (setq labels (delete-if-not (lambda (x)
-                                (let ((name (cdr (assoc 'name x))))
-                                  (or (string-prefix-p suffix name)
-                                      (when (equal (aref name 0) ??)
-                                        (aset name 0 ?~)
-                                        (string-prefix-p suffix name)))))
-                              labels))
-  (mapcar (lambda (x) (append x '((kind . "Label") (info . nil)))) labels))
+  ;; Remove non-matching entry, adjusting optional labels if needed
+  (cl-loop for x in labels
+           for name = (cdr (assoc 'name x))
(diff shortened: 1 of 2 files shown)


emacs/merlin.el
```diff
@@ -874,15 +874,14 @@ prefix of `bar' is `'."
...
-                                (let ((name (cdr (assoc 'name x))))
-                                  (or (string-prefix-p suffix name)
-                                      (when (equal (aref name 0) ??)
-                                        (aset name 0 ?~)
-                                        (string-prefix-p suffix name)))))
-                              labels))
-  (mapcar (lambda (x) (append x '((kind . "Label") (info . nil)))) labels))
+  ;; Remove non-matching entry, adjusting optional labels if needed
+  (cl-loop for x in labels
+           for name = (cdr (assoc 'name x))
```
trefis: Annoying request from an elisp-novice: I would find it more readable if this was `with name = (cdr ...)` rather than `for name = (cdr ...)`.
If I understood https://www.gnu.org/software/emacs/manual/html_node/cl/Other-Clauses.html#Other-Clauses correctly then both form are equivalents.

Do you mind changing? (Is the style you chose more idiomatic?)

rgrinberg: I don't know what's idiomatic elisp either but those forms aren't the same.

E.g. try:

```
(cl-loop for x in '(1 2 3) with y = (+ x 1) do (message "%d %d" x y))
```

Which doesn't work because you don't have access to `x` in the binding for `y`.

While this works as expected:

```
(cl-loop for x in '(1 2 3) for y = (+ x 1) do (message "%d %d" x y))
```

My reference for performance was this: http://nullprogram.com/blog/2017/01/30/

Though I did actually benchmark the changes here.

trefis: Ah, understood.
Also, elisp is a very sad language.

PR #805, opened 2018-04-13
Remove hard ocamlfind dependency

nojb:
We are using `merlin` under Windows without `ocamlfind`.

Currently, `merlin` does not work if `findlib` is not installed at run-time. This patch adds a stub module `Findlib_stub` and a configure-time option `--disable-findlib` to disable the runtime dependency on `findlib`.

Note that `ocamlfind` is still necessary for **building** `merlin`. In this direction, I have another patch that switches the build system to `jbuilder` (which supports OCaml all the way back to 4.02.3). This removes the compile-time dependency on `ocamlfind` and makes it trivial to build on Windows (our motivation). Do let me know if you would be interested in this patch, I can submit it for consideration as a PR as well.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ ocamlmerlin-test.exe
 config.iss
 ._d
 ._ncdi
+src/utils/findlib_stub.ml
 
 # Ignore garbage files from editors
 *.un~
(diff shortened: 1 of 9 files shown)


let-def:
Fyi, Merlin should work without findlib at runtime if you set OCAMLFIND_CONF=<some
empty file>`, or maybe even `OCAMLFIND_CONF=NUL` (null windows device)

nojb:
Ah, I did not think of that! Will give it a try and close this PR if it works. Thanks!

nojb:
I confirm that one can disable `ocamlfind` dependency along these lines. Thanks!

PR #812, opened 2018-05-01
emacs: support reason-mode in *merlin-types* buffer

Khady:
Correctly set face of content of `*merlin-types*` when type comes from reason file.


trefis:
Looks good, thanks!

PR #815, opened 2018-05-14
vim: fnameescape path argument to vim commands

ygrek:
allows to "jump to definition" when project path has spaces

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -246,6 +246,8 @@ def differs_from_current_file(path):
     buf_path = vim.eval("expand('%:p')")
     return buf_path != path
 
+def vim_fnameescape(s):
+    return vim.eval("fnameescape('%s')" % s.replace("'","''"))
 
 def command_locate(path, pos):
     try:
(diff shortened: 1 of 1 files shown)


ygrek:
ping

vim/merlin/autoload/merlin.py
```diff
@@ -246,6 +246,11 @@ def differs_from_current_file(path):
     buf_path = vim.eval("expand('%:p')")
     return buf_path != path
 
+def fnameescape(s):
+    # this will not work for filenames with quotes
+    # but to fix that need fnameescape defined on python side
+    # at least it will handle spaces
+    return vim.eval("fnameescape('%s')" % s)
```
trefis: I think you can probably do some form of escaping on `s` from python, we already do this in some places (search for `.replace`).

trefis:
I left one (possibly annoying) comment.
I'd be happier if it was addressed before merging, but otherwise I think the change you propose is fine indeed.

ygrek:
Thanks! Now it works for paths with quotes too

trefis:
Thanks!

Issue #820, opened 2018-05-25
In Neovim "MerlinTypeOf" shifts entire buffer


macthecadillac:
OS: Debian 9.4/Testing (on two separate machines)
Editor: NVIM v0.2.2
Merlin version: 3.0.5

With the version of Merlin I have installed, when I invoke `MerlinTypeOf` in Neovim, the entire buffer moves until the cursor touches the statusline. This is not the expected behavior from previous versions (I don't know which version specifically since I only update my opam packages once in a blue moon).

Minimal `init.vim` setup that reproduces the problem:
```vim
let g:opamshare = $HOME . '/.opam/4.06.1/share'
execute 'set rtp+=' . g:opamshare . '/merlin/vim'
execute 'set rtp+=' . g:opamshare . '/ocp-index/vim'
execute 'set rtp^=' . g:opamshare . '/ocp-indent/vim'
```

Thanks

copy:
I can't reproduce this (on Arch Linux with neovim  v0.2.2 and Merlin 3.0.5). Can you provide an OCaml source file and the location where you run `MerlinTypeOf` when this happens?

macthecadillac:
This file:

https://github.com/macthecadillac/Mancala/blob/master/common.ml

Thanks

copy:
@macthecadillac Can you try https://github.com/ocaml/merlin/pull/825 and see if it fixes the problem for you? 

(`git clone https://github.com/copy/merlin.git && cd merlin && git checkout neovim-fixes && opam pin add .`)

macthecadillac:
@copy It didn't fix it

copy:
@macthecadillac Can you try it with `opam pin add merlin 'git+https://github.com/copy/merlin.git#neovim-fixes'`?

yeshengm:
@copy This issue is still not resolved.

mreppen:
I can also replicate. Here are the details:
```
ocamlmerlin -version
The Merlin toolkit version 73e9771b, for Ocaml 4.09.0
```
NVIM v0.4.3

Minimal neovim config `merlintypeof.nvim.init`
```
let g:opamshare = substitute(system('opam config var share'),'\n$','','''')
execute 'set rtp+=' . g:opamshare . '/merlin/vim'
```

I cloned this repository at f6954e953b4168e6a798a0255d6a2dfbd868a3c6. The following replicates the behavior.
```
[...]

PR #825, opened 2018-05-30
Fix change of viewport after MerlinTypeOf in neovim

copy:
In newer versions of neovim, assigning to`vim.current.window.cursor`causes a change in the viewport (#820). In these versions, changing the cursor isn't necessary in the first place, as it was a workaround for older versions of neovim.

This PR also contains an unrelated fix for unnecessary filling of the `:messages` buffer due to an unsilenced `buffer` command.

### DIFF
--- a/vim/merlin/autoload/merlin_type.vim
+++ b/vim/merlin/autoload/merlin_type.vim
@@ -33,7 +33,7 @@ function! merlin_type#ShowTypeHistory()
   if l:win <# 0
     let t:merlin_restore_windows = winrestcmd()
     silent execute "bot " . g:merlin_type_history_height . "split"
-    execute "buffer" g:merlin_type_history
+    silent execute "buffer" g:merlin_type_history
   elseif winnr() !=# l:win
     exe l:win . "wincmd w"
   endif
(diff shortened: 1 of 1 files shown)


copy:
I'm not sure if this still fixes the original issue, because I can't reproduce it anymore. There's is another issue being fixed by this change, which is that the screen position is changed after the window from  `MerlinTypeOf` is closed. Compare the last frame: [master](https://asciinema.org/a/223044) and [this branch](https://asciinema.org/a/223043) (`vim` is aliased to `nvim`).

This fix reverts an old hack from 7efcb2265ed6525bf330666713f70cc83b32ce66 for neovim that doesn't seem to be necessary anymore.

Optimally, the type window shouldn't move the buffer at all if there's enough space, but my vimscript skills aren't good enough to fix this.

cc @trefis

trefis:
> This fix reverts an old hack from 7efcb22 for neovim that doesn't seem to be necessary anymore

Can you expand on that a bit: do you know how long ago this issue was fixed?
Basically, I'm just wondering how likely it is that someone is going to use a version of nvim that still has that bug, and that we'd *break* merlin for him, just to make it visually nicer to other people.

copy:
I tested this a bit more and found that the workaround is actually still necessary in recent versions of neovim. I have pushed an alternative fix that restores the old view after creating the type buffer.

trefis:
The current version looks fine, thanks!

PR #829, opened 2018-06-12
Various locate improvements

trefis:
Notable changes:
- support for partial `.cmt[i]` files
- better context detection
- better tracking of namespaces (i.e. less confused about various kind of items having the same name)
- initial support for functor, e.g. in the following jumping from `Foo.t` on line 15 will bring the cursor to line 6.
  ```ocaml
  (*  1 *) module type S = sig
  (*  2 *)   type t
  (*  3 *) end
  (*  4 *) 
  (*  5 *) module M = struct
  (*  6 *)   type t = T
  (*  7 *) end
  (*  8 *) 
  (*  9 *) module Make(Arg : S) : S = struct
  (* 10 *)   include Arg
  (* 11 *) end
  (* 12 *) 
  (* 13 *) module Foo = Make(M)
  (* 14 *) 
  (* 15 *) type t = Foo.t
  ```
- fix handling of exceptions and extension constructors
- properly handle local bindings

This fixes a couple of open issues.

There are no regression on the tests in the `tests` directory. No guarantee about tests not present in that directory. :)

### DIFF
--- a/Makefile.ocamlmakefile
+++ b/Makefile.ocamlmakefile
@@ -8,7 +8,7 @@ ifndef OVERRIDE_ENV
 	OCAMLDEP := ocamldep
 endif
 
-OCAMLFLAGS += -w -3
+OCAMLFLAGS += -w -3-40
 REALLY_QUIET ?= 1
 
 ifndef WITHOUT_BIN_ANNOT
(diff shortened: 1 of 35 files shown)


trefis:
Forgot to mention: one can now jump to `.mll` files.
Adding support for `.mly` should be trivial but I didn't have a test case at hand, so it will have to wait.

Khady:
> initial support for functor, e.g. in the following jumping from Foo.t on line 15 will bring the cursor to line 6.

Is there a way to run locate without this new feature?

PR #833, opened 2018-06-20
Switch to jbuilder ?

nojb:
This PR is a (WIP) switch of the build system to use `jbuilder`.

**Short story:**

If you clone this branch and do `jbuilder build` you should find the `ocamlmerlin_server.exe` binary in `_build/default/src/frontend`.

**Long story:**

The following is the dependency graph of the main libraries defined (continues after the image):

deps

- the `_X` suffix ranges over `402, 403, 404, 405, 406, 407`. These libraries are defined within `src/ocaml/typer_X`.
- there are two "virtual" libraries `typer` and `printf_compat` which chooses the right `*_X` according to the current OCaml version.
- `src/ocaml/support` contains code which is shared among all OCaml versions. Files here are copied into the corresponding subdirectories of `src/ocaml/typer_X` as needed.  It would be nice to structure `support` as a set of libraries, but this requires some care to avoid dependency cycles.
- `ocamlmerlin.exe` binary is not being built yet (it is a pure C binary)
- a couple of modules where moved around to simplify the `jbuilder` description: `support/pparse -> kernel/pparse`, `frontend/query_protocol -> kernel/query_protocol`.
- `jbuilder` does not support OCaml < 4.02.3, so I dropped the `typecore.ml` preprocessing in `configure`.
- ~~in order to avoid pulling `Cmt_cache` as a dependency of `Mocaml` (in `kernel`) I used forward definitions `Mocaml.cmt_cache_clear`, `Mocaml.cmt_cache_flush` which are filled during initialization of `Cmt_cache`. **This is too horrible to bear, something needs to be done about it.** My guess is that a little refactoring of the code may give us a way around.~~ [Fixed thanks to code refactoring]
- the `preprocess` step is not yet converted to `jbuilder`, I haven't looked into it, so am not sure how hard it would be.

So, in summary, there is a fair amount of work left to do. My impression is that by moving around some code the `jbuilder` description could be simplified considerably. And `jbuilder` enuthsiasts may have ideas to finish the remaining steps (`ocamlmerlin.exe` binary, `preprocess` step, etc).

Ideas, suggestions, etc welcome!


trefis:
>- in order to avoid pulling `Cmt_cache` as a dependency of `Mocaml` (in `kernel`) I used forward definitions `Mocaml.cmt_cache_clear`, `Mocaml.cmt_cache_flush` which are filled during initialization of `Cmt_cache`. **This is too horrible to bear, something needs to be done about it.** My guess is that a little refactoring of the code may give us a way around.

I just pushed 8fed929f036f2b37a135ce7adadcfd95295655a5 which should makes this unnecessary.
It's also fairly ugly, but `Cmi_cache` already works that way, and I think it's fine.

nojb:
Thanks, this solves that point.

trefis:
Note: I talked a bit with diml about the switch to jbuilder, and there is one problem with the organisation of our repo, which is that the directory structure is this:
```
src/
  ocaml/
    support/
      *.ml{,i}
    typer_407/
      *.ml{,i}
      parsing/
        *.ml{,i}
      typing/
        *.ml{,i}
    typer_406/
      *.ml{,i}
      parsing/
        *.ml{,i}
      typing/
        *.ml{,i}
    ...
```
And each of the `typer_XXX`'s jbuilds (and their subdirs') do a `(copy_files# ../support/<files>)`.
The problem with this is that jbuilder won't be able to generate a `.merlin` in `src/ocaml/support/`.
Whereas if the structure was:
```
src/
  ocaml/
    *.ml{,i}
    parsing/
      *.ml{,i}
      407/
      406/
      ...
    typing/
      *.ml{,i}
      407/
      406/
      ...
```
Without any jbuild files in the version-specific subdirs but the right `(copy_files# ...)` in their parent, then jbuilder will generate correct `.merlin` in the parent directories.

I'll do that reorganization when I go forward with the switch.

nojb:
Indeed! I was trying something similar over the weekend, it also results in a nice simplification of the build description (I added the commits to this PR). Once this is done, there are only a couple of points left:
- compilation of `ocamlmerlin` binary
- currently, OCaml syntax is being used in order to select the version-specific directories; alternatively, the `jbuild` files could also be generated by a preliminary `configure` step.
- `preprocess` step (support for `menhir --infer` is still WIP, so this part may need to be kept in a Makefile for now)

nojb:
> I'll take it from here. :)

Sure, I hope there wasn't too much duplicated effort.

PR #835, opened 2018-06-28
Switch to jbuilder

trefis:
This builds upon #833 (without directly reusing the commits though) and is (almost) complete:
- `jbuilder build` will build `ocamlmerlin[.exe]` and `ocamlmerlin-server[.exe]`
- `jbuilder build @src/ocaml/preprocess/XXX/preprocess --auto-promote` will regenerate `Lexer_raw`, `Parser_raw`, `Parser_explain`, `Parser_printer` and `Parser_recover` for the `XXX` version of OCaml.
- `jbuilder runtest` runs the testsuite.

Running `jbuilder build -j16` with the following `jbuild_workspace`:
```
(context ((switch 4.02.3)))
(context ((switch 4.03.0)))
(context ((switch 4.04.2)))
(context ((switch 4.05.0)))
(context ((switch 4.06.1) (merlin)))
(context ((switch 4.07.0+rc1)))
```
takes between 10 and 15s on average on my machine to build merlin for all these variants of OCaml.

This relies on `jbuilder.1.0+beta20`.

---

I'm not sure the windows support is quite right, in particular when building `ocamlmerlin.exe` with msvc the command line used to contain `/Feocamlmerlin.exe` and I think `jbuilder` might be inserting a space after the `/Fe` here, which might be a problem.

There might also be random other issues.
@nojb: would you be able to test this?

---

We used to have a rule to compile `.el` to `.elc` files when emacs is available. This is not expressible currently with `jbuilder`'s dsl, I could have another jbuild generated in ocaml, but I'm too lazy for that.
I'd be happy to accept a follow up PR reintroducing that.

---

- There were various annoying points during the conversation that I brought up with diml, I'm going to post an issue on dune's tracker to list all of these.
- I will send a follow up PR to fix all the warnings reported when running with `--dev` soon
- Preprocessing is still a bit clunky but should get better soon, hopefully before we need to support a new version of OCaml

Assuming this works on windows I'm in favor of merging.


trefis:
Also, one thing that became apparent with this switch, is that error messages still have an "Error: " prefix on 4.02 and some other versions; which means the tests are currently broken on these versions.

I'll fix that another time.

nojb:
> @nojb: would you be able to test this?

Will do.

nojb:
I did a quick test, indeed there was some trouble with the `/Fe` flag; this diff fixes it for me:
```diff
diff --git a/src/frontend/gen_ccflags.ml b/src/frontend/gen_ccflags.ml
index cbdc13bd..50952584 100644
--- a/src/frontend/gen_ccflags.ml
+++ b/src/frontend/gen_ccflags.ml
@@ -4,7 +4,7 @@ let post_flags_f = Sys.argv.(3)

 let pre_flags, post_flags =
   if Str.string_match (Str.regexp "msvc") ccomp_type 0 then
-    "/nologo /Fe", "advapi32.lib"
+    "/Fe", "advapi32.lib"
   else
     "-o", ""

diff --git a/src/frontend/jbuild b/src/frontend/jbuild
index e6859ed1..00147a78 100644
--- a/src/frontend/jbuild
+++ b/src/frontend/jbuild
@@ -34,7 +34,7 @@
 (rule
  ((targets (ocamlmerlin.exe))
   (deps    (ocamlmerlin.c pre-flags post-flags))
-  (action  (run ${CC} ${read-lines:pre-flags}"${@}" ${<} ${read-lines:post-flags}))))
+  (action  (run ${CC} "${read-lines:pre-flags}${@}" ${<} ${read-lines:post-flags}))))

 (install
  ((section bin)
```
(the `/nologo` flag is not necessary anyways since `dune` captures the output of the commands)

`ocamlmerlin` and `ocamlmerlin-server` seemed to compile fine afterwards.

It would be nice not to have a hard dependency on `findlib`, but that is for another time...

trefis:
Great, thanks for that Nicolás!

I'll include your diff and merge this PR later today then.

I wholeheartedly support dropping the hard dependency on `findlib`, but indeed, that will need to wait a bit longer (not too long though, that should be done in the next few weeks).

trefis:
I included @nojb's diff, and extended the "preprocess" alias to also regenerate `menhirLib.ml{,i}`.

PR #839, opened 2018-07-12
Implement async merlin/call

obscurans:
Implements an asynchronous version of merlin/call

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -578,7 +578,7 @@ return (LOC1 . LOC2)."
 (defun merlin/call (command &rest args)
   "Execute a command and parse output: return an sexp on success or throw an error"
   (let ((result (merlin--call-merlin command args)))
-    (merlin/callback-parse-result (merlin--parse-call-result result)))
+    (merlin/callback-parse-result (merlin--parse-call-result result))))
 
 (defun merlin/call-async (callback command &rest args)
   "Execute a merlin command asynchronously, returning the process object.
@@ -586,7 +586,7 @@ return (LOC1 . LOC2)."
 When the command finishes without being deleted, executes the callback with parsed output as single argument
 - if it is a success, argument is a cons cell with the sexp-result as car
 - if failure, argument is a single string as an error message
-- see merlin--callback-parse-result for convenience"
+- see merlin/callback-parse-result for convenience"
   (cl-letf (((symbol-function #'merlin--call-process)
              (apply-partially #'merlin--call-process-async callback)))
     (merlin--call-merlin command args)))
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -430,6 +430,29 @@ return (LOC1 . LOC2)."
 ;; PROCESS MANAGEMENT ;;
 ;;;;;;;;;;;;;;;;;;;;;;;;
 
+(defun merlin/callback-parse-result (result)
+  "For use by the callback: retrieve the actual result of the merlin call, or reraise the error"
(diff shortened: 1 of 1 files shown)


let-def:
Thanks for the patch. Do you have use case in mind?

Merlin mode used to be asynchronous by default, though we moved all subtle concurrency to the sub-process to prevent bugs in emacs.
I guess it is safe to make it async in emacs now, but we will have to be careful in which context the async version is used.

rgrinberg:
My main use case for async merlin calls would be to implement an async company completion backend. Completion is the only call where I feel some lag from merlin and it would be nice if this could be used to make it go away.

emacs/merlin.el
```diff
@@ -515,22 +578,18 @@ return (LOC1 . LOC2)."
...
-            (t (error "unknown answer: %S:%S" class value))))))
+    (merlin/callback-parse-result (merlin--parse-call-result result)))
+
+(defun merlin/call-async (callback command &rest args)
+  "Execute a merlin command asynchronously, returning the process object.
+
+When the command finishes without being deleted, executes the callback with parsed output as single argument
+- if it is a success, argument is a cons cell with the sexp-result as car
+- if failure, argument is a single string as an error message
+- see merlin--callback-parse-result for convenience"
```
Khady: Should be `merlin/callback-parse-result`

emacs/merlin.el
```diff
@@ -515,22 +578,18 @@ return (LOC1 . LOC2)."
...
-          (value (cdr-safe (assoc 'value result))))
-      (dolist (notification notifications)
-        (message "(merlin) %s" notification))
-      (cond ((string-equal class "return") value)
-            ((string-equal class "failure")
-             (error "merlin-mode failure: %s" value))
-            ((string-equal class "error")
-             (error "merlin: %s" value))
-            (t (error "unknown answer: %S:%S" class value))))))
+    (merlin/callback-parse-result (merlin--parse-call-result result)))
```
Khady: There is one closing parenthesis missing here.

Khady:
I was trying to look at this branch to use it in merlin-eldoc yesterday. IIUC with this API there is no way to make multiple async calls in a row and then wait for them to be finished using `accept-process-output`, right?

But I think the API would be fine to use with async company completion from what I checked.

obscurans:
Thanks for the bugfixes.

I think you can make multiple calls on separate processes and they will run in parallel, if you have a merlin server up.

Khady:
(and I am now wondering if using emacs-async wouldn't be more disgusting but easier)

rgrinberg:
@Khady I'm thinking that this PR should be reconsidered. The LSP client for Emacs is already as asynchronous as possible. Seems like we should stop with non-essential improvements to the old backend.

Khady:
I'm not sure to be the good person to ping as I'm not the one deciding what is interesting and should be included or not. And I don't know what is the policy regarding the old protocol vs lsp.

I never managed to benefit from this PR. My elisp isn't great though. So I don't have a strong attachment to it. It could make merlin-eldoc better. But most of its features are superseded by a good lsp client.

rgrinberg:
@Khady As far as I'm aware, only me and have you have experience with LSP on Emacs :) From what I've seen, both of the popular lsp clients for Emacs give you much better performance than the builtin client in merlin.

@obscurans would you mind trying out either `lsp-mode` or `eglot` once you install merlin-lsp. The performance should be quite good and completion is async.

rgrinberg:
Closing this because the author hasn't responded and LSP sounds like the right approach.

PR #842, opened 2018-07-19
Use echoerr for error message in vim plugin

copy:
Otherwise the message is shadowed by the error caused by `call merlin#Register`.

### DIFF
--- a/vim/merlin/autoload/merlin.vim
+++ b/vim/merlin/autoload/merlin.vim
@@ -11,7 +11,7 @@ elseif has('python')
   let g:merlin_python_version = 2
   command! -nargs=1 MerlinPy python <args>
 else
-  echo "Error: Required vim compiled with +python or +python3"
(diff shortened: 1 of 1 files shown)


vim/merlin/autoload/merlin.vim
```diff
@@ -242,7 +242,8 @@ function! merlin#PolaritySearch(debug,query)
   let s:search_result = []
   MerlinPy merlin.vim_polarity_search(vim.eval("a:query"), "s:search_result")
   if a:debug != 1 && s:search_result != []
-    call feedkeys("i=merlin#PolarityComplete()","n")
+    call feedkeys("i=merlin#PolarityComplete()
+","n")
```
trefis: I don't think this change should be here.

copy: Whoops, fixed.

(to be fair, a stray carriage return in the middle of a line is a bit odd).

trefis:
Thank you!

PR #847, opened 2018-07-31
Add a dune-project file

diml:
Otherwise `opam pin add merlin --dev` fails with:

```
...
### output ###
# Error: There is no dune-project file in the current directory, please add one with a (name <name>) field in it.
# Hint: dune subst must be executed from the root of the project.
...
```

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,0 +1,2 @@
+(lang dune 1.0)
+(name merlin)


let-def:
Thanks, I forgot to add this one :).

diml:
No problem. Could you rebase the pp branch? That would make it easier for me to pin it and test it with dune

diml:
thanks

PR #848, opened 2018-08-07
Extension calling error on MacOS

ewenmaclean:
In the current master of merlin, if you run a command such as

`ocamlmerlin single errors -reader reason < test.re`

for a reason file, you get an EOF error. Example output is:

```{"class":"exception","value":"End_of_file\nRaised at file \"pervasives.ml\", line 413, characters 9-26\nCalled from file \"pervasives.ml\", line 424, characters 2-25\nCalled from file \"src/extend/extend_main.ml\", line 108, characters 17-67\nCalled from file \"src/extend/extend_driver.ml\", line 34, characters 8-64\nRe-raised at file \"src/extend/extend_driver.ml\", line 39, characters 4-13\nCalled from file \"src/kernel/mreader_extend.ml\", line 52, characters 15-52\n...```

The reason for this is the change to `create_process` from `open_process` here:

https://github.com/ocaml/merlin/blob/master/src/extend/extend_driver.ml#L26

unfortunately the executable at (in this case) `ocamlmerlin-reason` cannot be found as the environment is empty, except for the necessary `__MERLIN_MASTER_PID` environment variable. What happens on single mode is that the path is made empty by this line:

https://github.com/ocaml/merlin/blob/master/src/frontend/new/new_merlin.ml#L138

In order to remedy this I have reset the `PATH` variable after this command, but I suspect that there is a more principled fix in the external functions provided by the `Os_ipc` module.

PR #856, opened 2018-09-04
Reader suffix association

ewenmaclean:
This PR allows new suffixes to be added to merlin so that `locate` will work on files with non-standard suffixes. It also allows associations of file extensions to specific readers. The existing behaviour is kept (automatic reader association for `.re` files), and any explicit reader argument overrides this association. 

The reason why this is a useful feature is so that one can choose a reader for a particular file type. For example in a folder with `imandra` files (`.iml` extension), you need to add the suffix pair `.iml,.imli` to the suffixes and associate the `imandra` reader with these files. Equally you might want to see the `imandra` types for `.ml` files and this is also now possible. For example one could call `ocamlmerlin` thus:
```
ocamlmerlin server errors -addsuffix .iml,.imli -assocsuffix .iml:imandra -assocsuffix ml:imandra  -filename test.ire < test.ire
```
equally with reason for example:
```
ocamlmerlin server errors -addsuffix .ire,.irei -assocsuffix .irel:imandra-reason -assocsuffix  -filename test.ire < test.ire
```
would use the `imandra-reason` reader for `.ire` files, but use the standard reason reader for `.re` files.


src/frontend/ocamlmerlin_server.ml
```diff
@@ -79,7 +79,9 @@ let main () =
   (* Setup env for extensions *)
   Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()));
   match List.tl (Array.to_list Sys.argv) with
-  | "single" :: args -> exit (New_merlin.run "" args)
+  | "single" :: args -> exit (New_merlin.run 
+    (String.concat "" (Array.to_list (Array.map (fun s -> s ^ "\000") (Unix.environment ())))) 
+    args)
```
trefis: I don't think this diff is meant to be there. Seems to be copied from #848 .

src/kernel/mconfig.ml
```diff
@@ -118,6 +118,8 @@ type merlin = {
 
   failures    : string list;
 
+  assocsuffixes : (string * string) list
```
trefis: I'd be in favor of a more explicit name like `extension_to_reader` (or `suffix_to_reader` if we want to keep the `suffix` term).

src/kernel/mconfig.ml
```diff
@@ -155,6 +157,12 @@ let dump_merlin x =
     "packages_path"    , `List (List.map ~f:Json.string x.packages_path);
 
     "failures"         , `List (List.map ~f:Json.string x.failures);
+    "assoc_suffixes"   , `List (
+      List.map ~f:(fun (suffix,reader) -> `Assoc [
+          "impl", `String suffix;
+          "intf", `String reader;
```
trefis: c/p failure I guess, it should be `"reader"` not `"intf"`

src/kernel/mreader.ml
```diff
@@ -66,16 +66,23 @@ let instantiate_reader tr spec config source = match spec with
...
     let reader = lazy (Mreader_extend.start tr name args config source) in
     (reader, (fun () ->
-       if Lazy.is_val reader then
-         match Lazy.force reader with
-         | None -> ()
-         | Some reader -> Mreader_extend.stop tr reader))
+         if Lazy.is_val reader then
+           match Lazy.force reader with
+           | None -> ()
+           | Some reader -> Mreader_extend.stop tr reader))
```
trefis: That diff shouldn't be here.

src/kernel/mreader.ml
```diff
@@ -66,16 +66,23 @@ let instantiate_reader tr spec config source = match spec with
...
 
 let get_reader config =
+  let rec find_reader assocsuffixes =
+    match assocsuffixes with
+    | [] when Filename.check_suffix Mconfig.(config.query.filename) ".re"
+           || Filename.check_suffix Mconfig.(config.query.filename) ".rei"  -> ["reason"]
```
trefis: I'd vote for pre-populating the `assocsuffix` field of `Mconfig.initial` with `["reason"]`, instead of keeping that special case here.

src/kernel/mreader.ml
```diff
@@ -66,16 +66,23 @@ let instantiate_reader tr spec config source = match spec with
...
+           | None -> ()
+           | Some reader -> Mreader_extend.stop tr reader))
 
 let get_reader config =
+  let rec find_reader assocsuffixes =
+    match assocsuffixes with
+    | [] when Filename.check_suffix Mconfig.(config.query.filename) ".re"
+           || Filename.check_suffix Mconfig.(config.query.filename) ".rei"  -> ["reason"]  
+    | [] -> []
+    | (hSuffix,reader)::t ->
```
trefis: What does the `h` of `hSuffix` stand for? head? If yes, I'd just drop it.

Also (stylistic remark) in ocaml we'd name this `h_suffix`.

src/frontend/ocamlmerlin_server.ml
ewenmaclean: Thanks - I will remove it, but is there a fix for the related bug in the latest master?

src/kernel/mconfig.ml
ewenmaclean: ok thanks - will update

src/kernel/mconfig.ml
ewenmaclean: yep apologies - will fix

src/kernel/mreader.ml
ewenmaclean: ocp-indent failure I think - will try to revert it

src/kernel/mreader.ml
ewenmaclean: ok good idea - will implement this

src/kernel/mreader.ml
ewenmaclean: ok thanks for the advice - will change

ewenmaclean:
@trefis  - in my experimentation I have come up with an alternative suggestion for this - I am not sure if you and the other core contributors might find it more principled.

Currently, this PR allows us to associate a specific reader with an extension - this means, for example, that in a VSCode workspace we can interpret the types for ml files using the standard reader or the imandra reader. Our current `ocamlmerlin` command with flags are something like

`ocamlmerlin server errors -addsuffix .iml,.imli -assocsuffix iml:imandra -open Mod - package pack ....`

which allows us to inspect types for `.iml` files using the imandra reader and `.ml` using the standard reader, but we could do this:

`ocamlmerlin server errors -addsuffix .iml,.imli -assocsuffix ml:imandra -assocsuffix iml:imandra  -open Mod - package pack ....`

if we wanted to see the imandra types on ml files.

The problem is that the `-open` and `-package` statements are universal. I was wondering if you would consider a PR which allowed arbitrary flags per extension - perhaps something like

`ocamlmerlin server errors -addsuffix .iml,.imli -assocsuffix iml:(-reader imandra -open Mod - package pack) ....`

for the first example and maybe

`ocamlmerlin server errors -addsuffix .iml,.imli -assocsuffix iml:(-reader imandra -open Mod - package pack)  -assocsuffix ml:(-reader imandra -open Mod - package pack) ....`

for the second

ewenmaclean:
@trefis  - thanks for the comment. Just a note - in order for `locate` to work properly the suffix has to be added so this (or something equivalent) is a necessary change. Of course the precise syntax and naming of the arguments is free for changing!

trefis:
Sorry for the delay!

ewenmaclean:
@trefis  - thanks a lot! Do you think we might be able to merge the fix given in #848 too, or come up with a better solution?

PR #858, opened 2018-09-08
WIP: Restore CI

dra27:
The CI is in a somewhat dilapidated state (possibly a combination of Merlin 3.1 and last-minute Jbuilder/Dune ports?!).

This PR is a start at getting it back up and running - I'd like to bring the AppVeyor scripts up-to-date and have AppVeyor activated for this repo, but it's a difficult to maintain Windows ports when the official Unix ones are not in a pristine state too!

The commits not marked "WIP" are all finalised. I've switched Travis to the experimental Xenial images, the switch to `realpath` in the testsuite is a problem on Trusty, as it uses the old `realpath` package, rather than the coreutils one.

The testsuite running on various versions of OCaml is having trouble, which is what the WIP commits try to address. I've got an opam workspace set-up with those versions of OCaml - if I post the diffs, can I have an opinion on whether each test should be fixed for that version of OCaml or whether that test should definitely be disabled. At the moment, the 4.02.3 testsuite is so broken I've just disabled it for Travis.

In theory, this PR should "pass" Travis, since it's disabled all tests which are failing - but I don't think it should be merged without first renaming the WIP commits and also ensuring that there are issues tracking any pertinent failing tests.

### DIFF
--- a/tests/merlin-wrapper-template
+++ b/tests/merlin-wrapper-template
@@ -2,7 +2,7 @@
 
 src_dir="$(dirname ${BASH_SOURCE[0]})/.."
 cd "$src_dir"
-src_dir=$(realpath -e $(pwd))
+src_dir=$(realpath -e $(pwd) 2>/dev/null || readlink -f $(pwd))
 cd - > /dev/null
 $(dirname ${BASH_SOURCE[0]})/%%OCAMLMERLIN_PATH%% $@ \
     | jq 'del(.timing)' \
--- a/tests/merlin-wrapper-template
+++ b/tests/merlin-wrapper-template
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 src_dir="$(dirname ${BASH_SOURCE[0]})/.."
 cd "$src_dir"
--- a/tests/merlin-wrapper-template
+++ b/tests/merlin-wrapper-template
@@ -4,7 +4,7 @@ src_dir="$(dirname ${BASH_SOURCE[0]})/.."
 cd "$src_dir"
 src_dir=$(realpath -e $(pwd) 2>/dev/null || readlink -f $(pwd))
 cd - > /dev/null
-$(dirname ${BASH_SOURCE[0]})/%%OCAMLMERLIN_PATH%% $@ \
+$(dirname ${BASH_SOURCE[0]})/%%OCAMLMERLIN_PATH%% "$@" \
     | jq 'del(.timing)' \
     | sed -e "s:$src_dir/::" \
     | sed -e 's:"/.*/\(\.opam\|opam2\)/[^/]*/\([^"]*\)":"~opam/\2":' \
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ script: bash -ex .travis-opam.sh
 env:
   matrix:
   - OCAML_VERSION=4.02
+    TESTS=false
   - OCAML_VERSION=4.03
   - OCAML_VERSION=4.04
   - OCAML_VERSION=4.05
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: c
 sudo: true
+dist: xenial
 install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-opam.sh
 script: bash -ex .travis-opam.sh
 env:
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ clean:
 	dune clean
 
 test:
-	dune runtest
+	dune runtest --force
 
 preprocess:
 	dune build @preprocess
--- a/merlin.opam
+++ b/merlin.opam
@@ -17,7 +17,7 @@ build-test: [
 ]
 
 depends: [
-  "dune"  {build}
+  "dune"  {build & >= "1.1.0"}
   "ocamlfind" {>= "1.5.2"}
   "yojson"
   "craml" {test}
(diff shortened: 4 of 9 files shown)


tests/merlin-wrapper-template
```diff
@@ -1,8 +1,8 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 src_dir="$(dirname ${BASH_SOURCE[0]})/.."
 cd "$src_dir"
-src_dir=$(realpath -e $(pwd))
+src_dir=$(realpath -e $(pwd) 2>/dev/null || readlink -f $(pwd))
```
dra27: This change doesn't appear to be working correctly on my Trusty VM, but I haven't yet looked into why (equally, it's only a few more months before we don't have to worry about Ubuntu 14.04...)

trefis:
Hey, sorry for the silence treatment...

I agree the tests are a mess currently.
I agree with your approach of "let's disable the tests on all the versions but one" until it is fixed.

I'll try to have a look at your commits in the next few days so we can get this merged.

Thanks!

dra27:
No worries - we're all busy! I'd countenance against merging with tests disabled - they'll never get fixed at that point! All I need to know for the WIP commits is whether it's Merlin failing or the analysis (for example, in some cases the JSON clearly varies slightly, it's just not clear whether that's an issue if the output should be normalised, or compared with the reference file "more intelligently").

trefis:
- the warnings tests are fixed on all versions of caml by c846744f381e8b7a4767be5ebd660c93dce8d5a7 so [WIP: warnings test only works on 4.06](https://github.com/ocaml/merlin/pull/858/commits/084c54c98ccae9647cdfb834a1c2ef3753db728f) can be removed
- [WIP: completion test only works on Travis for 4.07](https://github.com/ocaml/merlin/pull/858/commits/923f1e264f9306e6481960f0bed2033addc0422b) I wasn't able to reproduce the issue locally
- disabling the 4.02 tests seem like the right move for the moment: various locations have changed after that version and it'd require some work to make 4.02 agree with the other versions.
- [WIP: enable most of the tests for 4.03.0](https://github.com/ocaml/merlin/pull/858/commits/712411285bf59738c83e7ff49bc12943471be7d4) can be dropped, as it is subsumed by the conjunction of 8c616703694436dd7699abd21d046b4cf6dd4510 and 11fecbde192a5f85ed48f894ded40cabfedd4b93

Everything else looks good, if you rebase I'll click the merge button.

Note: there might be issues with the new `tests/pp/simple-pp.t` test, at least I have some weird errors on my setup: `cpp: error trying to exec 'cc1': execvp: No such file or directory` but I'll have a look at that another day.

dra27:
Let's see how this one fares - I think 5f4aa1e should probably be removed too, but let's see if the 4.07 issue can be repeated by Travis first.

dra27:
Final commit not needed once https://github.com/ocaml/ocaml-ci-scripts/pull/274 merged

trefis:
There seem to be a couple of issues remaining. I believe @let-def is looking at the "cpp -E" problem.

The failure of tests/type-expr/test.t is interesting and apparently caused by 8b40d64. And I think it's actually a fix rather than a breakage. So you're welcome to accept the correction for it in that commit.

dra27:
I was able to repro the cpp issue in my box, and the commits I added fixed it it here. It’s to do with the environment reset in `New_merlin.run`. I hadn’t had a chance to look into why it hasn’t fixed Travis.

There’s also a problem with the ppx tests because the craml script assumes sh is bash - not sure what the best fix for that is.

I’d got the impression that the other failure was is in fact a fix - thanks for confirming.

I am, amusingly, rehearsing in a dungeon today where there is little signal, but I’ll pick this back up this evening!

dra27:
Fixing the `sh` problem is now related to https://github.com/realworldocaml/mdx/pull/85

If I'm _really_ lucky, the tests may start passing apart from those ones...

trefis:
OK, only the `pp` test is broken right now.
So I think I'll just merge the PR as is, disable that test, and come back to you once a new release of mdx (is that the new craml?) happens.

Thanks @dra27 !

trefis:
I was too optimistic. The fact that the tests passed locally didn't actually imply they'd pass on travis. :sob:

dra27:
Yeah, sorry - I'd got distracted with the MDX bit and hadn't commented on the rest ... that test reproducibly fails on my Ubuntu box as well. The items are the same, but the list is in a different order... I hadn't got as far as debugging whether the output should be sorted (using jq) or whether there was an underlying bug which needs fixing.

MDX is indeed the new craml ... unfortunately at the moment it's 4.06+, but that'll be being sorted "soon", hopefully!

trefis:
> The items are the same, but the list is in a different order

Indeed. That's surprising. @let-def: any idea what that would be happening? (Note: it's the result of expand-prefix)

let-def:
I suspect this is due to the module enumeration code: it depends on `Sys.readdir`, which returns directory entries in an arbitrary order. This might fix it:

```diff
diff --git a/src/ocaml/utils/misc.ml b/src/ocaml/utils/misc.ml
index feed3c34..cec6c991 100644
--- a/src/ocaml/utils/misc.ml
+++ b/src/ocaml/utils/misc.ml
@@ -579,6 +579,8 @@ let modules_in_path ~ext path =
   List.fold_left ~init:[] path
     ~f:begin fun results dir ->
       try
+        let dirs = Sys.readdir dir in
+        Array.sort String.compare dirs;
         Array.fold_left
           begin fun results file ->
             if Filename.check_suffix file ext
@@ -588,7 +590,7 @@ let modules_in_path ~ext path =
                else
                  (Hashtbl.add seen name (); String.capitalize name :: results))
             else results
-          end results (Sys.readdir dir)
+          end results dirs
       with Sys_error _ -> results
     end
```

trefis:
> I suspect this is due to the module enumeration code: it depends on `Sys.readdir`, which returns directory entries in an arbitrary order.

I agreed with that diagnostic and pushed the diff to master. But that doesn't to have affected the test at all.

PR #865, opened 2018-10-05
Fix preprocess step

nojb:
As the title says; not sure if this may become simpler/not needed with the improved `menhir` stanza.


trefis:
That looks better but indeed, I don't think this will be needed when we use the new stanza.
And:
1. I have a patch to use the new stanza
2. I don't think we'll need to run the preprocess step before switching to the new stanza

But thanks!

nojb:
Sure, no problem.

PR #868, opened 2018-10-08
Fix build against latest sturgeon

rgrinberg:
and add appropriate opam constraints

Signed-off-by: Rudi Grinberg <rudi.grinberg@gmail.com>

### DIFF
--- a/merlin.opam
+++ b/merlin.opam
@@ -23,6 +23,14 @@ depends: [
   "craml" {test}
 ]
 
+depopts: [
+  "sturgeon"
+]
+
+conflicts: [
+  "sturgon" {< "0.4.2"}
+]
+
 available: [ ocaml-version >= "4.02.1" & ocaml-version < "4.08" ]
 
 post-messages: [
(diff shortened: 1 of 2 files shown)


let-def:
The fix seems right. @rgrinberg: the version of sturgeon released on opam as a wrong install script that copy no files... I made a patch to fix the install procedure. I wonder if I should rerelease 0.4.1 or make a 0.4.2 that is fixed. Do you have any opinion on that?

rgrinberg:
Releasing 0.4.2 would be preferable in my opinion.

rgrinberg:
I think the conflict is necessary as otherwise opam will attempt installation of newer version of merlin along side older versions of sturgeon.

I just changed the conflict to 0.4.2

trefis:
Ping @let-def

let-def:
That's good.

Fourchaux:
In 'merlin.opam' - Line 31 
 
"sturgon" {< "0.4.2"}   should be  "sturgeon" {< "0.4.2"}, no?

PR #870, opened 2018-10-09
Ppx cwd

let-def:
This branch executes ppx binary in the same directory as the .merlin, such that relative paths that are passed to a ppx can be interpreted in the right directory.

Dubious: I left a patch that allows expansion of `~/` in the paths processed by Merlin.
I added it for experimentation purposes but I think it should be removed, unless @trefis thinks otherwise :).

### DIFF
--- a/src/ocaml/driver/pparse.ml
+++ b/src/ocaml/driver/pparse.ml
@@ -14,8 +14,6 @@ type error =
   | CannotRun of string
   | WrongMagic of string
 
-exception Error of error
-
 (* Note: some of the functions here should go to Ast_mapper instead,
    which would encapsulate the "binary AST" protocol. *)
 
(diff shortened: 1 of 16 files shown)


src/ocaml/driver/pparse.ml
```diff
@@ -28,44 +28,70 @@ let write_ast magic ast =
...
+    Logger.logf "Pparse" "report_error"
+      "External preprocessor does not produce a valid file. Command line: %s" cmd
 
 let ppx_commandline cmd fn_in fn_out =
-  Printf.sprintf "%s %s %s 1>%s 2>%s"
-    cmd (Filename.quote fn_in) (Filename.quote fn_out) null null
-
-let pp_commandline cmd fn_in fn_out =
-  Printf.sprintf "%s %s 1>%s"
+  Printf.sprintf "%s %s %s 1>&2"
```
trefis: Unless I missed something, I don't like that diff: we were previously redirecting `stdout` and `stderr` to `/dev/null`, while now you're just redirecting `stdout` to `stderr`.
Which means that if a ppx ever prints something, I'll get weird text appearing on my vim window, that isn't actually in the buffer. And in emacs it will probably spawn a `*spurious*` buffer or something of the sort, won't it?

Piping to `null` as we were doing is definitely the easiest way to avoid these problems. But if you feel particularly motivated you could also capture both outputs and have merlin display them to the user.

let-def: Vim (through python) handle the output properly by redirecting stderr to another buffer:
https://github.com/ocaml/merlin/blob/master/vim/merlin/autoload/merlin.py#L136-L139

Emacs will redirect stderr to a file on disk if debug-mode is enabled and drop it otherwise:
https://github.com/ocaml/merlin/blob/master/emacs/merlin.el#L433-L455
as well as the blurb explaining the brilliant `call-process-region` function: https://www.gnu.org/software/emacs/manual/html_node/elisp/Synchronous-Processes.html

trefis: Oh, I didn't know about that. It's all "new"!
Thanks.

trefis:
PS: feel free to remove the expansion of `~/` if you want.

PR #875, opened 2018-10-18
Pp: interpret `-pp` OCaml flag to run source-to-source preprocessors.

let-def:
This PR implements the -pp flag from the OCaml compiler.
Arbitrary rewriting can be done on the source before it gets parsed by merlin.

It has been tested with https://github.com/realworldocaml/mdx to get a "literate programming" experience.


src/ocaml/driver/pparse.ml
```diff
@@ -138,28 +138,19 @@ let apply_rewriters ~ppx ?restore ~tool_name = function
...
-    close_in ic;
-    Misc.remove_file fn;
-    ast
-  with exn ->
-    close_in ic;
-    Misc.remove_file fn;
-    raise exn
-
 let pp_commandline cmd fn_in fn_out =
   Printf.sprintf "%s %s 1>%s"
```
trefis: Same kind of comment as in the other PR: we should either drop everything printed on `stderr`, or properly capture it.

trefis:
Apart from the comments above, I think this looks good.
Although perhaps you could explain why "Lexer_raw.update_loc: ignore filename change" is needed?

let-def:
All errors should refer to the buffer in the editor, but the preprocessor is likely to introduce the name of the temporary file in location directives. Ignoring the changes is a quick fix.

PR #880, opened 2018-11-09
Check for errors when opening new ocaml files in Emacs

Wilfred:
Otherwise, the user has to explicitly save the file before they can
see any merlin errors.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1720,6 +1720,7 @@ Empty string defaults to jumping to all these."
   (let ((configuration (merlin--configuration)))
     (when configuration (setq merlin-buffer-configuration configuration)))
   (add-to-list 'after-change-functions 'merlin--on-edit)
+  (merlin-error-check)
   ;; TODO: Sanity check for selected merlin version
   (unless merlin--idle-timer
     (setq merlin--idle-timer


trefis:
I don't think this has its place in merlin's default configuration:
- it's dubious that everyone would want this: it can slow down the opening of new files, it can be very noisy when you're opening a file in a project that isn't compiled yet, ...
- unless I'm mistaken, it's very easy to update your .emacs to have that behavior

Thank you for the suggestion though!

rgrinberg:
I also agree that this seems like the wrong way to accomplish this. Is there a flycheck option to check a buffer after opening it? That would solve the problem in general.

Wilfred:
Flycheck and flymake always check the buffer when you open the file, so that set my expectations. I also found it surprising that if I explicitly call `M-x merlin-mode` in a buffer, I don't see any merlin errors until I save.

It's possible to work around this in my own Emacs configuration, but my concern is that it's not discoverable. Have I missed something?

I'm less concerned about performance: I've tested this on an ocaml codebase with >100KLOC and >100 errors and it wasn't an issue. Flycheck is actually more aggressive than merlin: most checkers will update even if you don't save.

I suppose it's a question of whether you see merlin.el as primarily a code completion tool, or an error display package :)

jberdine:
FWIW, maybe it's just me, but I not infrequently open files and explicitly do not want merlin to report errors, since the file has not been compiled, is at the other end of an only minimally-configured tramp connection, etc.

rgrinberg:
> Flycheck and flymake always check the buffer when you open the file, so that set my expectations. I also found it surprising that if I explicitly call M-x merlin-mode in a buffer, I don't see any merlin errors until I save.

So why doesn't it work in this case for you? Flycheck should drive this via https://github.com/flycheck/flycheck-ocaml. My main concern is such a setting shouldn't really be chosen at the mode level. It would be annoying to set this for ocaml, python, haskell, etc.

Wilfred:
I've added a note to the wiki page. Thanks for clarifying :)

jberdine:
> FWIW I've been thinking of adding tramp support for Merlin, which 
> would suit my workflow and maybe yours too?

tramp support for merlin would be most excellent!

Wilfred:
I did explore using flycheck with merlin, but it doesn't handle error ranges. It can't handle highlighting the part of the line that has the type error (see https://github.com/flycheck/flycheck-ocaml/issues/12 and https://github.com/flycheck/flycheck/pull/1400). At this point, I think merlin.el works better without flycheck.

PR #882, opened 2018-11-14
[RFC] Changed Vscode link to vscode-reasonml.

bschommer:
Since the original vscode update is deprecated and recommends
using the other plugin instead.

I tested the reasonml plugin and it did work for basic OCaml files even if the reason stuff was not configured and only merlin was installed. Maybe @hackwaly can give a bit more context.


bschommer:
It seems that the plugin I was linking to is also not maintained any longer.

PR #883, opened 2018-11-14
Fix warnings by ignored return value chdir.

bschommer:
The chdir function is marked with the warn_unused_result attribute
in some clibraries which triggers a warning if the return value of
the function is not checked.

These warnings appear if merlin is compiled for example in more recent Ubuntu version.


trefis:
Thanks!
But #858 which I have left waiting for a while but will hopefully merge very soon already include such changes.

PR #888, opened 2018-12-11
Emacs: Ensure we handle buffer narrowing correctly

### LINKED ISSUE #886: Emacs: Merlin is confused by narrowing
If the buffer is narrowed, merlin only checks the visible buffer. This is noticeable when you have some code:

```
let foo x = ...

and bar x = ...
```

Narrowing to the definition of `bar` leads to merlin complaining about syntax errors.

Wilfred:
Refactor `merlin--point-of-pos` to be a pure function that can return
positions outside of the narrowed region. `merlin--goto-char` now uses
`merlin--point-of-pos` and widens if necessary.

Update callers to use `merlin--point-of-pos` if they don't want widening.

Also simplify `merlin--search`, which never used its second argument.

Fixes #886.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -397,36 +397,46 @@ containing fields file, line and col."
 ;; Position management
 
 (defun merlin--goto-point (data)
-  "Go to the point indicated by `DATA' which must be an assoc list with fields
-line and col"
-  (goto-char (point-min))
-  (forward-line (1- (merlin-lookup 'line data 0)))
-  ; caml gives us the byte offset of the column, which doesn't necessarily match
-  ; the character offset, so we can't just use "forward-char"
-  (let* ((bol-offset (position-bytes (point)))
-         (col-offset (max 0 (merlin-lookup 'col data 0)))
-         (target-off (+ bol-offset col-offset)))
-    (goto-char (byte-to-position target-off))))
+  "Go to the point indicated by DATA which must be an assoc list with fields
+line and col. If narrowing is in effect, widen if DATA is outside the visible region."
+  (let ((line-num (merlin-lookup 'line data 0))
+        (col-byte-offset (merlin-lookup 'col data 0))
+        (target-pos (merlin--point-of-pos data)))
+    ;; If our target position is outside the narrowed region, we'll
+    ;; have to widen.
+    (when (or (< target-pos (point-min))
+              (> target-pos (point-max)))
+      (widen))
+    (goto-char target-pos)))
 
-(defun merlin--point-of-pos (pos)
-  "Return the buffer position corresponding to the merlin
-position POS."
-  (save-excursion
-    (merlin--goto-point pos)
-    (point)))
+(defun merlin--point-of-pos (data)
+  "Transform DATA (a remote merlin position) into a point.
+DATA must be an assoc list with fields line and col."
+  (let ((line-num (merlin-lookup 'line data 0))
+        (col-byte-offset (merlin-lookup 'col data 0)))
+    (save-excursion
+      (save-restriction
+        (widen)
+        (goto-char (point-min))
+        (forward-line (1- line-num))
+        ;; Find the target position, converting the byte position to a
+        ;; character offset.
+        (let* ((bol-offset (position-bytes (point)))
+               (col-offset (max 0 col-byte-offset))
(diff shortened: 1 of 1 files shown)


rgrinberg:
I tested this PR out and can confirm that merlin now works with narrowing. However, I see a weird issue with invalid buffers. Consider this buffer:

```
let foo = ()

let bar = ()

List.filter
```

Narrow so that only the last 3 lines are visible. Then ask merlin to give you the type of `List.filter`. 	Merlin will output the write type, but highlight an incorrect section.

rgrinberg:
@Khady I believe your eldoc package is broken by this PR as well.

Khady:
I don't mind adapting merlin-eldoc so that the ignored argument is not necessary.

Thanks for this PR, this is a useful change.

rgrinberg:
I now get the following error when trying this out:

```
Debugger entered--Lisp error: (wrong-number-of-arguments (lambda (_ignored) "Destruct the current point position to line / col." (save-restriction (widen) (format "%d:%d" (line-number-at-pos) (current-column)))) 0)
  (merlin/unmake-point)
  (let* ((merlin/verbosity-context t) (position (merlin/unmake-point)) (verbosity (cdr-safe merlin--verbosity-cache)) (types (merlin/call "type-enclosing" "-position" position "-index" 0)) (types (condition-case nil (progn (mapcar (function (lambda (obj) (let* ((tail (cdr (assoc 'tail obj))) (tail (cond ((equal tail "position") " (* tail position *)") ((equal tail "call") " (* tail call *)") (t ""))) (type (cdr (assoc 'type obj)))) (cons (if (stringp type) (concat type tail) (list type position tail verbosity)) (merlin--make-bounds obj))))) types)) (error nil))) (types (delq nil types))) (if types (progn (setq merlin-enclosing-types types) (setq merlin-enclosing-offset -1) merlin-enclosing-types)))
  (merlin--type-enclosing-query)
  (if (merlin--type-enclosing-query) (progn (merlin-type-enclosing-go-up) (merlin--type-enclosing-after)))
  (if (region-active-p) (merlin--type-region) (if (merlin--type-enclosing-query) (progn (merlin-type-enclosing-go-up) (merlin--type-enclosing-after))))
  (merlin-type-enclosing)
  (funcall-interactively merlin-type-enclosing)
  (call-interactively merlin-type-enclosing nil nil)
  (command-execute merlin-type-enclosing)
```

Wilfred:
@rgrinberg thanks, fixed.

emacs/merlin.el
```diff
@@ -412,12 +433,12 @@ position POS."
     (merlin--goto-point data)
     (point)))
 
-(defun merlin/unmake-point (point)
-  "Destruct POINT to line / col."
-  (save-excursion
-    (goto-char point)
-    (beginning-of-line)
-    (format "%d:%d" (line-number-at-pos) (- point (point)))))
+;; Accept extra parameters to support merlin-eldoc which relies on this API.
```
trefis: Not sure what `merlin-eldoc` is, but aren't you breaking it if you just ignore the argument?

trefis: Ah sorry, I had missed the fact that @Khady's aware of this.
Then I'd say we should just remove the argument, instead of ignoring it.

rgrinberg:
This is good to do once the extra argument for merlin-eldoc is removed

Wilfred:
@rgrinberg removed :)

rgrinberg:
@Wilfred I'm still seeing some weirdness that I can't consistently reproduce. The issue is roughly as follows:

* I narrow to a region (say some function)
* I ask merlin to give me the type of some function.
* Merlin gives me the type of the function
* The buffer is widened automatically for me.

To reproduce, try opening this project https://github.com/ocaml/dune/blob/master/src/dune_file.ml#L66 narrowing to the default function and asking merlin to give you the type of `default`.

Wilfred:
@rgrinberg Thanks again :)

I've refactored the code so `merlin--point-of-pos` is a pure function, and code that doesn't want widening side-effects use that.

This has also simplified the patch somewhat, so `merlin/unmake-point` still takes a point as an argument. This should save @Khady needing to make any changes.

rgrinberg:
Works great. Sorry this took so long.

PR #889, opened 2018-12-12
Emacs: Document that merlin-type-enclosing can increase verbosity

Wilfred:
This is a really useful feature that isn't widely known.


trefis:
Thanks!

PR #895, opened 2018-12-29
Fix Occurrences query to climb out of patterns with type constraints

andreypopp:
My first time contribution to merlin, don't know codebase at all so would appreciate some help!

This fixes `Occurrences` query to handle cases where queried identifier is annotated with type constraint:
```
let withTypeAnnot (some : int) =
  some + some
```

I've also added general test cases for `Occurrences` query as a part if this PR.

### DIFF
--- a/tests/occurrences/dune
+++ b/tests/occurrences/dune
@@ -1,0 +1,9 @@
+(alias
+ (name runtest)
+ (deps (:t basic.t) basic.ml)
+ (action
+   (progn
+     (setenv MERLIN %{exe:../merlin-wrapper}
+     (setenv OCAMLC %{ocamlc}
+       (run %{bin:craml} %{t})))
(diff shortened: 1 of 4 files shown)


andreypopp:
Any feedback on this?

tests/occurrences/dune
```diff
@@ -0,0 +1,9 @@
+(alias
+ (name runtest)
+ (deps (:t basic.t) basic.ml)
+ (action
+   (progn
+     (setenv MERLIN %{exe:../merlin-wrapper}
+     (setenv OCAMLC %{ocamlc}
```
trefis: That line is not needed (only merlin is used in your test, not the compiler).

andreypopp: Fixed.

PR #901, opened 2019-01-14
Return type info in module outlines

andreypopp:
This adds type information for value bindings in `outline` query.

This allows to implement CodeLens like functionality. Example with neovim below:

codelens-neovim

### DIFF
--- a/src/analysis/outline.ml
+++ b/src/analysis/outline.ml
@@ -39,8 +39,8 @@ let id_of_patt = function
   | { pat_desc = Tpat_var (id, _) ; _ } -> Some id
   | _ -> None
 
-let mk ?(children=[]) ~location outline_kind id =
-  { Query_protocol.  outline_kind; location; children;
+let mk ?(children=[]) ~location outline_kind outline_type id =
+  { Query_protocol.  outline_kind; outline_type; location; children;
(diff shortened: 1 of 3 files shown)


src/analysis/outline.ml
```diff
@@ -49,21 +49,33 @@ let get_class_field_desc_infos = function
...
   match node.t_node with
-  | Value_binding vb      -> id_of_patt vb.vb_pat >>| mk `Value ~location
-  | Value_description vd  -> Some (mk `Value ~location vd.val_id)
+  | Value_binding vb      ->
+    begin match id_of_patt vb.vb_pat with
+    | None -> None
+    | Some ident ->
+      let outline_type =
+        let ppf, to_string = Format.to_string () in
+        Type_utils.print_type_with_decl ~verbosity node.t_env ppf vb.vb_pat.pat_type;
```
trefis: If I'm reading the diff correctly, this is the only use of `verbosity`. If that's the case, could you inline it its definition here and remove the binding? I think it'd make things more readable.

andreypopp: Fixed this1

andreypopp:
Thanks, bringing this in.

PR #905, opened 2019-01-22
Add LSP frontend

andreypopp:
This adds `ocamlmerlin-lsp` which is a part of a new `merlin-lsp.opam` package.

`ocamlmerlin-lsp` adds LSP frontend for merlin with the following initial set of features:

- `textDocument/definition` — analogue of `locate`
- `textDocument/typeDefinition` — no analogue in current merlin frontend, like `locate` but locates the type of the specified expression.
- `textDocument/completion` — analogue of `complete-prefix`
- `textDocument/publishDiagnostics` — analogue of `errors` query, pushed by server on changes to sources.
- `textDocument/documentSymbol` — analogue of `outline` query.

Structure:

- `merlin-lsp.opam` — new `merlin-lsp` package definition
- `src/lsp` — LSP protocol library
- `src/frontend/lsp` — `ocamlmerlin-lsp` frontend
- `tests-lsp` — tests for `ocamlmerlin-lsp` (far from comprehensive)

I'm using this with neovim/ALE now and have vscode plugin which I'm not submitting in this PR but will do after some cleanup. For now it's hosted at https://github.com/andreypopp/vscode-merlin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ jbuild-workspace
 
 ocamlmerlin
 ocamlmerlin-server
+ocamlmerlin-lsp
 
 # Ignore garbage files from editors
 *.un~
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,17 @@ script: bash -ex .travis-opam.sh
 env:
   matrix:
   - OCAML_VERSION=4.02
+    PACKAGE=merlin
     TESTS=false
   - OCAML_VERSION=4.03
+    PACKAGE=merlin
   - OCAML_VERSION=4.04
+    PACKAGE=merlin
   - OCAML_VERSION=4.05
+    PACKAGE=merlin
   - OCAML_VERSION=4.06
+    PACKAGE=merlin
   - OCAML_VERSION=4.07
+    PACKAGE=merlin
 os:
   - linux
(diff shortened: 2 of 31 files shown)


andreypopp:
Ok, it's green now on 4.02: I moved `Std.Result` to `Lsp.Utils.Result` for now so it won't interfere with `merlin.lsp` support of 4.02.

By the way `merlin-lsp.opam` itself is marked with `ocaml >= 4.04` but it would be trivial (I think) to fix to run under 4.03 and then 4.02.

hcarty:
I think the `dune` version constraints need to be bumped since the language version is bumped in `dune-project`.

andreypopp:
@hcarty good catch! Will update the constraint.

andreypopp:
For the record the implementation is based heavily off [bryphe/melrin-language-server](https://github.com/bryphe/merlin-language-server) by @bryphe with protocol additions from [facebook/flow](https://github.com/facebook/flow).

I already mentioned that in commit messages but should have mentioned that here too.

avsm:
Great to see this feature! However, we need to be careful of not having any license violations. In particular, the `LICENSE` file in Merlin is not a reproduction of the `LICENSE` present in @bryphe and the Flow repository. The names of the contributors should be added to the root `LICENSE` file to ensure that the correct names are present, or reproduced in the individual files which were derived from the upstream repositories.

avsm:
And of course @let-def is too fast for me; he has addressed this in https://github.com/ocaml/merlin/pull/906 :-)

rbjorklin:
This looks interesting! Would this warrant a new release of merlin any time soon?

PR #908, opened 2019-01-29
Update missing copyrights

let-def:
Copyright line is now "   Copyright (C) 2013 - 2019  Merlin contributors".

IANAL, I have no idea what I am doing.

### DIFF
--- a/src/analysis/expansion.ml
+++ b/src/analysis/expansion.ml
@@ -1,3 +1,29 @@
+(* {{{ COPYING *(
+
+  This file is part of Merlin, an helper for ocaml editors
+
+  Copyright (C) 2013 - 2019  Merlin contributors
+
+  Permission is hereby granted, free of charge, to any person obtaining a
+  copy of this software and associated documentation files (the "Software"),
+  to deal in the Software without restriction, including without limitation the
+  rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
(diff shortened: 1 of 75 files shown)


let-def:
@avsm: Flow copyright is copied there (with the path to MIT LICENSE updated) https://github.com/ocaml/merlin/pull/908/files#diff-1ead05ed940a205c93fcfead82ab8caeR36

src/analysis/expansion.ml
```diff
@@ -1,3 +1,29 @@
+(* {{{ COPYING *(
+
+  This file is part of Merlin, an helper for ocaml editors
+
+  Copyright (C) 2013 - 2019  Merlin contributors
```
trefis: "Merlin contributors"

Not sure what value that holds...

avsm:
(IANAL also) My understanding of the Merlin repository is that the original MIT license was added to the repository 6 years ago.  Since then, the common understanding is that any contributions sent here via PRs would be covered by that license as well.  Therefore, you can define the `Merlin contributors` somewhere in the repository as the assemblage of those GitHub ids.  So the PR here seems fine to me, with some clarification perhaps being helpful about who those contributors are so there is no ambiguity.

XVilka:
Should be this one reconsidered too? Would be a nice addition for the next release (with 4.09 support) if accepted

PR #910, opened 2019-01-31
emacs: make sure PATH is updated when merlin-command is 'opam

actionshrimp:
Previously the path to `ocamlmerlin` was built through prepending the result of `opam config var bin`. This works for the merlin binary itself, but if additional readers are used (such as ocamlmerlin-reason), merlin reverted to using the environmental path to look up the location of these binaries.

This means that if you have an reader on your global path that's incompatible with the current opam switch's merlin/ocaml version, merlin gets very unhappy.

This calls `ocamlmerlin` via `opam config exec` instead, which sets up the environment correctly and means it finds other merlin binaries in the switch as you'd expect.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1674,7 +1674,8 @@ Empty string defaults to jumping to all these."
          (with-temp-buffer
            (if (eq (call-process-shell-command
                     "opam config var bin" nil (current-buffer) nil) 0)
-               (progn
+               (let ((bin-path
+                      (replace-regexp-in-string "\n$" "" (buffer-string))))
                  ;; the opam bin dir needs to be on the path, so if merlin
                  ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
                  ;; correct version is used rather than the version that
@@ -1682,11 +1683,8 @@ Empty string defaults to jumping to all these."
 
                  ;; this was originally done via `opam exec' but that doesnt
                  ;; work for opam 1, and added a performance hit
-                 (setq merlin-opam-bin-path
-                       (list (concat "PATH=" (string-trim (buffer-string)))))
-                 (concat
-                  (replace-regexp-in-string "\n$" "" (buffer-string))
-                  "/ocamlmerlin"))
+                 (setq merlin-opam-bin-path (list (concat "PATH=" bin-path)))
+                 (concat bin-path "/ocamlmerlin"))
 
              ;; best effort if opam is not available, lookup for the binary in
              ;; the existing env
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1685,7 +1685,9 @@ Empty string defaults to jumping to all these."
                             "/ocamlmerlin"))
 
                        ;; best effort if opam is not available, lookup for the binary in the existing env
-                       "ocamlmerlin")))))
+                       (progn
+                         (message "merlin-command: opam config failed (%S)" (buffer-string))
+                         "ocamlmerlin"))))))
 
       ;; cache command in merlin-buffer configuration to avoid having to shell
       ;; out to `opam` each time.
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1665,29 +1665,35 @@ Empty string defaults to jumping to all these."
 
   (let ((command (merlin-lookup 'command merlin-buffer-configuration)))
     (unless command
-      (setq command
-            (cond ((functionp merlin-command) (funcall merlin-command))
-                  ((stringp merlin-command) merlin-command)
-                  ((equal merlin-command 'opam)
-                   (with-temp-buffer
-                     (if (eq (call-process-shell-command "opam config var bin" nil (current-buffer) nil) 0)
-                         (progn
-                           ;; the opam bin dir needs to be on the path, so if merlin
-                           ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
-                           ;; correct version is used rather than the version that
-                           ;; happens to be on the path
+      (setq
+       command
+       (cond
+        ((functionp merlin-command) (funcall merlin-command))
+        ((stringp merlin-command) merlin-command)
+        ((equal merlin-command 'opam)
+         (with-temp-buffer
+           (if (eq (call-process-shell-command
+                    "opam config var bin" nil (current-buffer) nil) 0)
+               (progn
+                 ;; the opam bin dir needs to be on the path, so if merlin
+                 ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
+                 ;; correct version is used rather than the version that
+                 ;; happens to be on the path
 
-                           ;; this was originally done via `opam exec' but that doesnt
-                           ;; work for opam 1, and added a performance hit
-                           (setq merlin-opam-bin-path (list (concat "PATH=" (string-trim (buffer-string)))))
-                           (concat
-                            (replace-regexp-in-string "\n$" "" (buffer-string))
-                            "/ocamlmerlin"))
+                 ;; this was originally done via `opam exec' but that doesnt
+                 ;; work for opam 1, and added a performance hit
+                 (setq merlin-opam-bin-path
+                       (list (concat "PATH=" (string-trim (buffer-string)))))
+                 (concat
+                  (replace-regexp-in-string "\n$" "" (buffer-string))
+                  "/ocamlmerlin"))
 
-                       ;; best effort if opam is not available, lookup for the binary in the existing env
-                       (progn
-                         (message "merlin-command: opam config failed (%S)" (buffer-string))
-                         "ocamlmerlin"))))))
+             ;; best effort if opam is not available, lookup for the binary in
+             ;; the existing env
+             (progn
+               (message "merlin-command: opam config failed (%S)"
+                        (buffer-string))
+               "ocamlmerlin"))))))
 
       ;; cache command in merlin-buffer configuration to avoid having to shell
       ;; out to `opam` each time.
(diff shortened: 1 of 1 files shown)


Khady:
When you say that additional readers are used, you mean that ocamlmerlin will make a call to ocamlmerlin-reason? And you would face the issue you describe if you start your merlin from a shell that doesn't never ran `eval opam env`, right?

Sorry for the orthogonal questions, I am trying to understand what this is solving exactly and why must setup is working.

Two comments about the patch:
1. `'command` in `merlin-buffer-configuration` is not updated anymore. Which doesn't seem to break anything. I don't know if line 1656 `(push (cons 'command command) merlin-buffer-configuration)` was here for a reason. 
2. This doesn't fallback to looking for the ocamlmerlin binary in the path if `opam` is not executable.

actionshrimp:
Thanks for taking a look - I think you're right on the first part and it should be possible to replicate from a shell without `eval $(opam env)` - the part in particular is here https://github.com/ocaml/merlin/blob/master/src/extend/extend_driver.ml#L26. If you do have the other reader binary on your path then it may work 'accidentally' - i only ran into this when i had an ocaml 4.02 built `ocamlmerlin` in my local switch calling out to `ocamlmerlin-reason` on my non-switch path built with ocaml 4.06, which was causing errors.

Thanks for the comments, I'll take a look at 2) specifically when I get a chance. RE 1) as 'command' is now still just `'opam` as the main `merlin-command` var is, I'm not sure what implications that has on the buffer-configuration which had a string in it before. I'll try and understand that in a bit more detail too.

actionshrimp:
Added a few tweaks:
- Left the original `opam config` call in to check opam works, so the fallback to `ocamlmerlin` is still there
- Set 'command to 'opam in merlin-buffer-configuration (rather than the absolute path to ocamlmerlin which it was before). This matches the configured `merlin-command` variable value more closely
- Use `opam exec` rather than `opam config exec` - turns out the `config` isn't required.

Khady:
opam exec is here since opam2. I don’t know if it worth keeping compatibility with opam1. Everyone should have migrated by now, but that’s not sure.

actionshrimp:
One thing I've just noticed - `opam config exec` seems to have a fair bit of overhead to it oddly:

```$ time opam config exec -- ocamlmerlin single
Usage: ocamlmerlin command [options] -- [compiler flags]
Help commands are:
  -version        Print version and exit
  -vnum           Print version number and exit
  -warn-help      Show description of warning numbers
  -flags-help     Show description of accepted compiler flags
  -commands-help  Describe all accepted commands

opam config exec -- ocamlmerlin single  0.12s user 0.09s system 99% cpu 0.206 total

$ time _opam/bin/ocamlmerlin single
Usage: ocamlmerlin command [options] -- [compiler flags]
Help commands are:
  -version        Print version and exit
  -vnum           Print version number and exit
  -warn-help      Show description of warning numbers
  -flags-help     Show description of accepted compiler flags
  -commands-help  Describe all accepted commands

_opam/bin/ocamlmerlin single  0.00s user 0.01s system 93% cpu 0.009 total
```

I've got something working that builds the path up as before but also adds the `opam config bin` dir to the PATH env var so it works on both opam 1 and 2, and also seems to be quite a lot faster - I'd noticed my emacs was getting a bit sluggish. Will tidy it up and push it up.

actionshrimp:
All seems to work as before without the perf hit. I've also cached the env/path building in the merlin buffer configuration to avoid having to shell out an extra time on each call.

emacs/merlin.el
```diff
@@ -1662,21 +1662,40 @@ Empty string defaults to jumping to all these."
...
+
+                  ;; this was originally done via `opam exec' but that doesnt
+                  ;; work for opam 1, and added a performance hit
+                  (setq merlin-opam-bin-path (list (concat "PATH=" (string-trim (buffer-string))))))
+
+              (setq command "ocamlmerlin")))
+
+        (setq command
+              (cond ((functionp merlin-command) (funcall merlin-command))
+                    ((stringp merlin-command) merlin-command))))
```
Khady: Nitpicking, but I think the code would be easier to follow if the `equal merlin-command 'opam` check was part of this cond.

```emacs-lisp
(defun merlin-command ()
  "Return or update path of ocamlmerlin binary selected by configuration"
  (unless merlin-buffer-configuration
    (setq merlin-buffer-configuration (merlin--configuration)))

  (let ((command (merlin-lookup 'command merlin-buffer-configuration)))
    (unless command
      (setq
       command
       (cond
        ((functionp merlin-command) (funcall merlin-command))
        ((stringp merlin-command) merlin-command)
        ((equal merlin-command 'opam)
         (with-temp-buffer
           (if (eq (call-process-shell-command "opam config var bin" nil (current-buffer) nil) 0)
               (progn
                 ;; the opam bin dir needs to be on the path, so if merlin
                 ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
                 ;; correct version is used rather than the version that
                 ;; happens to be on the path

                 ;; this was originally done via `opam exec' but that doesnt
                 ;; work for opam 1, and added a performance hit
                 (setq merlin-opam-bin-path (list (concat "PATH=" (string-trim (buffer-string)))))
                 (concat
                  (replace-regexp-in-string "\n$" "" (buffer-string))
                  "/ocamlmerlin"))

             ;; best effort if opam is not available, lookup for the binary in the existing env
             "ocamlmerlin")))))

      ;; cache command in merlin-buffer configuration to avoid having to shell
      ;; out to `opam` each time.
      (push (cons 'command command) merlin-buffer-configuration)
      (when (boundp 'merlin-opam-bin-path)
        (push (cons 'env merlin-opam-bin-path) merlin-buffer-configuration)))

    command))
```

trefis:
Can someone comment on the relation between this and #848?

emacs/merlin.el
actionshrimp: Ah good spot, thanks - pushed.

actionshrimp:
@trefis - Yes - from reading that PR I believe it's fixing an issue where the PATH was being cleared completely when calling a merlin subprocess - I'm not sure which code path it was, possibly `ocamlmerlin single` only. This would cause merlin to not find the subprocess binary at all, even if on the global path.

This PR is around an issue where the PATH from the environment is still preserved, but has not been updated correctly before calling `ocamlmerlin` itself from emacs. This would cause a version of `ocamlmerlin` from an opam switch to call the global path subprocess binary, rather than the appropriate binary from the switch.

emacs/merlin.el
```diff
@@ -1662,21 +1662,37 @@ Empty string defaults to jumping to all these."
...
-    (when (equal command 'opam)
-      (with-temp-buffer
-        (if (eq (call-process-shell-command
-                 "opam config var bin" nil (current-buffer) nil) 0)
-            (progn
-              (setq command (concat
-                             (replace-regexp-in-string "\n$" "" (buffer-string))
-                             "/ocamlmerlin"))
-              (push (cons 'command command) merlin-buffer-configuration))
-          (message "merlin-command: opam config failed (%S)" (buffer-string))
```
trefis: I think we should keep that message.

emacs/merlin.el
```diff
@@ -1662,21 +1662,37 @@ Empty string defaults to jumping to all these."
...
-          (setq command "ocamlmerlin"))))
+      (setq command
+            (cond ((functionp merlin-command) (funcall merlin-command))
+                  ((stringp merlin-command) merlin-command)
+                  ((equal merlin-command 'opam)
+                   (with-temp-buffer
+                     (if (eq (call-process-shell-command "opam config var bin" nil (current-buffer) nil) 0)
```
trefis: Github doesn't show it, but I'd be willing to be that goes over 80c.

emacs/merlin.el
```diff
@@ -1662,21 +1662,37 @@ Empty string defaults to jumping to all these."
...
+                         (progn
+                           ;; the opam bin dir needs to be on the path, so if merlin
+                           ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
+                           ;; correct version is used rather than the version that
+                           ;; happens to be on the path
+
+                           ;; this was originally done via `opam exec' but that doesnt
+                           ;; work for opam 1, and added a performance hit
+                           (setq merlin-opam-bin-path (list (concat "PATH=" (string-trim (buffer-string)))))
```
trefis: That's probably over 80c too, isn't it?

trefis: Also, why is the `(replace-regexp-in-string "\n$" ...)` needed for setting `command`, but not for `merlin-opam-bin-path`?

emacs/merlin.el
actionshrimp: Re-added.

emacs/merlin.el
actionshrimp: Good point. I've used that in both places now as of the last commit - originally I had a `string-trim` but perhaps there's a reason the `replace-regexp-in-string "\n$"` was used originally so I'll stick with that, thanks.

actionshrimp:
Thanks for reviewing - have included your suggestions and formatted to 80c.

emacs/merlin.el
```diff
@@ -1662,21 +1662,43 @@ Empty string defaults to jumping to all these."
...
+               (let ((bin-path
+                      (replace-regexp-in-string "\n$" "" (buffer-string))))
+                 ;; the opam bin dir needs to be on the path, so if merlin
+                 ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
+                 ;; correct version is used rather than the version that
+                 ;; happens to be on the path
+
+                 ;; this was originally done via `opam exec' but that doesnt
+                 ;; work for opam 1, and added a performance hit
+                 (setq merlin-opam-bin-path (list (concat "PATH=" bin-path)))
```
Wilfred: This should be defined as a global variable with `defvar`. This code produces a byte-compiler warning.

trefis: Would you mind sending a PR for it?

Wilfred: Done, see #962.

PR #914, opened 2019-02-17
lsp: add jsonrpc version field to responses as per spec

actionshrimp:
Tried hooking up the emacs lsp-mode client to merlin's LSP server and it expects the `jsonrpc` field in responses, as per the 'abstract message' part of the spec: https://microsoft.github.io/language-server-protocol/specification .

With this change the emacs client connects fine.


andreypopp:
Thank you!

PR #916, opened 2019-02-18
Re-align identifier numbers with compiler

lpw25:
Move `Btype.unbound_class` to `Env` so that it appears after `Predef` in
the dependency order and doesn't change the identifier stamps of the
predefined identifiers.


trefis:
I've built this locally for all the ocaml versions we support.
I then tested it with my patched version of mdx, on all the versions except 4.02: it fixes the issue.

PR #918, opened 2019-02-19
Update merlin.opam

Bronsa:


trefis:
Shouldn't that `version` field just be removed from the opam file?

trefis:
Ping @AltGr .

AltGr:
It's allowed either way: not having it can be more convenient when you have your own version change management mechanics, but having it means it gets picked up by `opam pin` (which will just default to version `dev` if not found)

trefis:
Thanks for your anwser!
If I understand correctly, this will matter for libraries (which might have revdeps), but probably doesn't for merlin (which shouldn't have revdeps).

So I'll close this and remove the version field from our opam file.

Thank you as well @Bronsa!

PR #922, opened 2019-02-24
LSP: Add "Find references"

rusty-key:
Support `text/findReferences`, analogue of `occurances`
https://microsoft.github.io/language-server-protocol/specification#textDocument_references

/cc @andreypopp


andreypopp:
Thanks @rusty-key, that looks great!

I've rebased on top of master which has LS tests hooked up to the CI. Will merge after CI confirms this.

PR #923, opened 2019-02-24
Run LSP tests on CI

andreypopp:
- Fixes tests for LSP frontend and makes deterministic (sort output of completion, not sure why it's different across machines but this is what original protocol tests are doing too)
- Add new job which tests LSP frontend with 4.07


andreypopp:
@let-def thanks!

Yeah, I've tried to make it so regular builds are unaffected. Should be ok for now, we can revisit later if needed.

One thing which worries me is that Travis CI is so slow lately (I think this is because of they being acquired recently.

I vote for a transition to azure which also can help testing on macOS and Windows in addition to having Linux builds.

PR #924, opened 2019-02-25
[RFC] do not stage parser

trefis:
Or any of the related files.
Instead directly depend on menhir.

In the process:
- fixed a type error in the 4.02 parser
- manually silenced a bunch of "unused value" warnings


trefis:
@let-def reported that menhir's API is unfortunately not stable, so we cannot afford not to stage.

andreypopp:
Curious what was the motivation for such cnahge.

trefis:
Well, not only do I find the various dune files who currently live under `src/ocaml/preprocess` ugly.
There's also the fact that the rules in there have been silently broken for a while, since dune has moved where it stores its `.cmi` files.

This PR not only simplified the whole thing, it also used the `menhir` rule which is built in dune, and which is kept up to date when `.cmi` files are moved around.

trefis:
Note: I'll try to cook up an alternative solution using the `include_subdirs` stanza.

andreypopp:
Makes sense!

Just that introduction of a new dependency (menhir) made worry a bit — merlin is installed in almost any project I've seen — it's a downside if it would impose a new constraint on menhir.

PR #925, opened 2019-02-26
LSP: provide sortText in completion items

Khady:
This PR adds fills the `sortText` field of completion items in `textDocument/completion` response. This is necessary to make sure that the client will respect the order decided by merlin.

Tested on emacs only.

https://microsoft.github.io/language-server-protocol/specification#textDocument_completion
https://github.com/tigersoldier/company-lsp/issues/56#issuecomment-424593632

> Per [specification](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion), the server can specify `sortText` for each candidate. Candidates should be sorted by `sortText` if the server populates them, or sorted by the labels if `sortText` is not populated.


andreypopp:
Thanks!

PR #926, opened 2019-02-26
LSP: change hover response from string to code

Khady:
I am not sure about this one. But I have the impression that the value returned by hover will always be some kind of ocaml/reason code. So the answer should be a MarkedCode instead of a MarkedString.

### DIFF
--- a/src/frontend/lsp/ocamlmerlin_lsp.ml
+++ b/src/frontend/lsp/ocamlmerlin_lsp.ml
@@ -215,36 +215,39 @@ let on_request :
     return (store, params)
 
   | Lsp.Rpc.Request.TextDocumentHover {textDocument = {uri;}; position;} ->
-    Document_store.get store uri >>= fun doc ->
-    let command =
-      Query_protocol.Type_enclosing (
-        None,
-        logical_of_position position,
-        None
-      )
+    let query_type doc pos =
+      let command = Query_protocol.Type_enclosing (None, pos, None) in
+      match Query_commands.dispatch (Document.pipeline doc) command with
+      | []
+      | (_, `Index _, _) :: _ -> None
+      | (_, `String value, _) :: _ -> Some value
     in
-    begin match Query_commands.dispatch (Document.pipeline doc) command with
-    | [] -> return (store, None)
-    | (_loc, `Index _, _is_tail) :: _rest -> return (store, None)
-    | (_loc, `String value, _is_tail) :: _rest ->
-      let contents =
-        let supports_markdown =
-          List.mem
-            Lsp.Protocol.MarkupKind.Markdown
-            client_capabilities.textDocument.hover.contentFormat
-        in
-        if supports_markdown
-        then {
-          Lsp.Protocol.MarkupContent.
-          value = "```ocaml\n" ^ value ^ "\n```";
-          kind = Lsp.Protocol.MarkupKind.Markdown;
-        }
-        else {
-          Lsp.Protocol.MarkupContent.
-          value;
-          kind = Lsp.Protocol.MarkupKind.Plaintext;
-        }
+
+    let format_contents ~as_markdown typ =
+      if as_markdown
+      then {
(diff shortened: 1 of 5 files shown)


andreypopp:
Yeah, have to test this one in vscode. Some thoughts:

- Maybe we don't need to wrap `contents` into backticks if we return `MarkedCode`
- We probably can infer `ocaml` or `reason` from document's URI
- I think this should enable syntax highlighting? That would be nice!

Khady:
```
        if
          List.mem
            Lsp.Protocol.MarkupKind.Markdown
            client_capabilities.textDocument.hover.contentFormat
        then [Lsp.Protocol.MarkedString.MarkedCode { language = "OCaml"; value = contents } ]
        else [Lsp.Protocol.MarkedString.MarkedString contents]
```

Maybe the answer should be like this actually. I don't have a clear view of what is the good solution. And I need to fix a bug in lsp-mode before to have a good client to test with.

andreypopp:
Few more thoughts:

- We should probably return `MarkupContent` which is either `markdown` or `plaintext`. `MarkedString` is marked as deprecated.
- You are correct (from discussion on Discord) that we need not to include fenced codeblocks and then vscode will highlight the content properly.

andreypopp:
I pushed a fix to use `MarkupContent` and tested it in vim and vscode.

src/frontend/lsp/ocamlmerlin_lsp.ml
```diff
@@ -224,18 +224,28 @@ let on_request :
...
           List.mem
             Lsp.Protocol.MarkupKind.Markdown
             client_capabilities.textDocument.hover.contentFormat
-        then "```\n" ^ contents ^ "\n```"
-        else contents
+        in
+        if supports_markdown
+        then {
+          Lsp.Protocol.MarkupContent.
+          value = "```ocaml\n" ^ value ^ "\n```";
```
Khady: this is not critical, but String.concat or sprintf would be more performant

andreypopp: Thanks, fixed.

andreypopp:
@Khady I think we can merge this if you can check emacs is happy with this.

Khady:
I'm not 100% sure emacs does the good thing, but I tend to think that this PR is correct.

https://github.com/emacs-lsp/lsp-mode/issues/673

andreypopp:
@Khady well, don't want to merge this and break workflow for emacs users... let's wait then

Khady:
So the situation with emacs is:

- when using MarkedString (as in master branch), the markdown is rendered using the good mode, displayed with nice colors and it basically work
- when using MarkedContent, it has some bugs, the content is not always displayed and when it is, the coloration is not good.

I didn't manage to find so far what creates this problem :(

andreypopp:
@Khady the coloration isn't good in vs code too to be honest — it tries to color it as OCaml but those snippets are not always valid OCaml.

The previous VsCode extension used a different grammar for this but to trigger this now we need to use different than `ocaml` lang specifier in markdown codeblock. I'm thinking to remove the lang completely for now (the coloration might be handled on per editor basis for now, personally I don't think it's super useful but still nice to have). Do you think if we remove language specifier and just return a codeblock — that would help emacs?

Khady:
I fixed emacs in https://github.com/emacs-lsp/lsp-ui/pull/249

Now IIUC the emacs situation is (without webkit support):
- for MarkupContent, the value returned by the server is "rendered" by emacs and then displayed to the user. There is no syntax highlighting in the markdown itself. But at least it is not messed up.
- for MarkedString:
  - if there is no language, the content is rendered as markdown and code blocks are colorized
  - if there is a language, the content is rendered using the major mode corresponding to the language. Or displayed as plaintext if there is no major mode for the language.

andreypopp:
@Khady am I correct that we can merge this now then?

andreypopp:
Thanks!

PR #927, opened 2019-02-26
LSP: Implement textDocument/codeLens

andreypopp:
This reuses merlin's outline feature and returns types per position.

TODO:
- [x] Add tests


andreypopp:
Tested with VsCode and vim. VsCode UX with CodeLens is suboptimal (not specific to `ocamlmerlin-lsp`) so we probably want it disabled by default. But it's nice in neovim:

codelens

PR #928, opened 2019-02-27
Lsp misc cleanup

Khady:
- remove some tabs used instead of spaces
- fix a conversion between type and json value
- add missing field in a type
- small additions to completion


andreypopp:
Looks great! Thanks!

PR #931, opened 2019-02-28
LSP: switch detail and doc in completion answer

Khady:
The content of `entry.info` is long and correspond to the documentation. The type (`entry.desc`) must be in detail to be displayed.

emacs-merlin-completion


andreypopp:
Thanks! This is indeed a correct fix — tested with VsCode too.

PR #935, opened 2019-03-07
LSP: add support for textDocument/rename

Khady:
The implementation is as complete as it could be, but it works for me. Tested in emacs and vscode.

- I did not implement all the parts of the yojson conversion that were not needed for rename support. So `WorspaceEdit` can be in a response but not a request. 
- I did not implemented all the `CreateFile | RenameFile | DeleteFile` types for `resource_operation` to be usable. They are not yet necessary for now as merlin doesn't work on multiple files. The type exist to reflect what is in the spec.
- in `WorkspaceEdit.changes` there is a hack to get a uri as a string. This is necessary because the spec for `changes` is `{ [uri: string]: TextEdit[]; }`.
- the capabilities of the client are not checked before to create the response, for now it always assumes that `workspace.workspaceEdit.documentChanges` is true.
- there are no tests

### DIFF
--- a/src/frontend/lsp/ocamlmerlin_lsp.ml
+++ b/src/frontend/lsp/ocamlmerlin_lsp.ml
@@ -557,27 +557,9 @@ let on_request :
       {Lsp.Protocol.TextEdit. newText = newName; range;}
     ) locs
     in
-    let documentChanges =
-      let textDocument = {
-        Lsp.Protocol.VersionedTextDocumentIdentifier.
-        uri;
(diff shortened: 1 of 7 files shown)


Khady:
I added tests and check of client capabilities. I don't plan to add what is missing to implement the `of_yojson` function unless it is absolutely mandatory.

andreypopp:
Thanks!

I've re-organised code a bit (mainly around not implemented cases).

PR #936, opened 2019-03-07
Use menhir stanza

trefis:
Second attempt.

This is waiting on: https://github.com/ocaml/opam-repository/pull/13605

We're still committing the generated parser, as well as `Parser_{explain,printer,recover}`.
But we're using dune's `menhir` stanza, and the parsers are all regenerated when working in dev mode.
When building in release mode (which happens when installing from opam), it's just using the staged versions of everything.

Note: I'd guess all the `parser_raw.mli` changes are due to the interface being dumped with `-short-paths` on.


trefis:
The CI failure is only for lsp, but that exact failure also exists on #935 .
I don't know what's going on, ping @andreypopp ? :)

Khady:
The failure in #935 looks different to me. It was an unused variable there. Here a library is missing

```
dune build
File "src/utils/dune", line 9, characters 11-39:
9 |    (copy %{lib:menhirLib:menhirLib.mli} menhirLib.mli))))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Public library "menhirLib" not found
File "src/ocaml/preprocess/explain/dune", line 3, characters 17-26:
3 |  (libraries unix menhirSdk))
                     ^^^^^^^^^
Error: Library "menhirSdk" not found.
Hint: try: dune external-lib-deps --missing @@default
Error: Program menhir not found in the tree or in PATH (context: default)
Hint: try: opam install menhir
File "src/ocaml/preprocess/printer/dune", line 3, characters 17-26:
3 |  (libraries unix menhirSdk))
                     ^^^^^^^^^
Error: Library "menhirSdk" not found.
Hint: try: dune external-lib-deps --missing @@default
File "src/ocaml/preprocess/recover/dune", line 3, characters 17-26:
3 |  (libraries unix menhirSdk))
                     ^^^^^^^^^
Error: Library "menhirSdk" not found.
```

trefis:
Ah! Thanks @Khady, I guessed I should have read more of the log...

trefis:
OK so the remaining problem with travis is in .travis.yml, the line:
```sh
opam config exec -- make release ocamlmerlin-lsp
```
is not run in the right environment / switch.
So: merlin-lsp is built a second time, but that time the dune version which is used is not the same that was used for the initial build (because it's not executing in the same env/switch). And in particular, it fails to interpret properly some of the rules.

I think the solution would be to simply remove that line. And make sure the lsp tests reference the binary in _build. As we do for the normal tests.
(It would also have the nice side benefit of not requiring merlin to be built twice)

andreypopp:
To be honest I'm puzzled why this change triggered this error. If you are super blocked on this — let's disable LSP tests for now — I can take a look into this more thoroughly early next week.

trefis:
It's not yet super urgent. At some point it will become blocking for the support of 4.08, but that's still a few days out. So we'll see!

I tried a few things today to get it to work, which failed miserably. Unfortunately it takes a long time to get feedback from travis...

Anyway, no rush!

andreypopp:
Hm... `make release` fail for me locally too with the same error (dup rules for parser_*.ml`), I was able to fix that with:
```
diff --git a/src/ocaml/preprocess/dune b/src/ocaml/preprocess/dune
index 12ef0e08..cbde2330 100644
--- a/src/ocaml/preprocess/dune
+++ b/src/ocaml/preprocess/dune
@@ -48,6 +48,7 @@ Printf.ksprintf J.send {|
 
 (menhir
  (modules parser_raw)
+ (enabled_if (<> %%{profile} "release"))
  (mode    (promote-into %s))
  (flags :standard --inspection --table --cmly))
 
@@ -55,6 +56,7 @@ Printf.ksprintf J.send {|
   (targets parser_recover.ml)
   (deps    parser_raw.cmly)
   (mode    (promote-into %s))
+  (enabled_if (<> %%{profile} "release"))
   (action
     (with-stdout-to %%{targets}
        (run %%{exe:./recover/gen_recover.exe} %%{deps}))))
@@ -63,6 +65,7 @@ Printf.ksprintf J.send {|
   (targets parser_explain.ml)
   (deps    parser_raw.cmly)
   (mode    (promote-into %s))
+  (enabled_if (<> %%{profile} "release"))
   (action
     (with-stdout-to %%{targets}
        (run %%{exe:./explain/gen_explain.exe} %%{deps}))))
@@ -71,6 +74,7 @@ Printf.ksprintf J.send {|
   (targets parser_printer.ml)
   (deps    parser_raw.cmly)
   (mode    (promote-into %s))
+  (enabled_if (<> %%{profile} "release"))
   (action
     (with-stdout-to %%{targets}
        (run %%{exe:./printer/gen_printer.exe} %%{deps}))))
```

trefis:
You shouldn’t have to do that with dune >= 1.8.

andreypopp:
I have dune 1.8.2. It still fails with the same error:
```
Multiple rules generated for _build/default/src/ocaml/preprocess/parser_recover.ml:
- _build/.dune/default/src/ocaml/preprocess/dune:45
- _build/.dune/default/src/ocaml/preprocess/dune:13
```

andreypopp:
This fixes the LSP tests — https://github.com/ocaml/merlin/commit/8958b298dae9494714388b9b2d02de44a23f4d96

More specifically `-p` flag fixed it. I've also removed installing yet another node (as travis image has one already) and added `set -e` so we fail earlier.

trefis:
Thanks @andreypopp!

I'm puzzled by the fact that "dune build --profile=release" wasn't good enough, but "dune build -p merlin-lsp" is, it looks like a bug to me. But perhaps @diml will have an explanation.

trefis:
OK, @diml just explained to me offline that `-p` implies `--profile=release` but also `--ignore-promoted-rules`, which is what actually matters here!

andreypopp:
> OK, @diml just explained to me offline that -p implies --profile=release but also --ignore-promoted-rules, which is what actually matters here!

That's good to know!

I guess this is also why `make` fails for me with missing `menhirLib` if I don't have menhir installed. I guess we need to add `menhir` as dev dependency then?

trefis:
> I guess this is also why make fails for me with missing menhirLib if I don't have menhir installed

Yes!
I don't know if opam has a notion of dev dependency... Let's ping @rjbou this time!

Khady:
yes, you can add `{dev}` annotation on a dep. To be used the same way as `{build}`.

rjbou:
You can find the full list of packages variables in the [opam doc](https://opam.ocaml.org/doc/Manual.html#Package-variables).

trefis:
Thanks!

IIUC adding a dependency on menhir conditional on `{dev}` means that if someone does `opam pin merlin --dev` they'll also depend on menhir. That doesn't sound ideal.
I think keeping menhir out of the opam file, and updating the readme to inform people of the menhir story is the most reasonable.

Khady:
In this kind of situation, I usually add a dev-tools.opam file to my repository. This way it is easy to install all the dependencies while developing the project. And at the same time it doesn’t annoy people who would install the project using opam pin —dev.

trefis:
Thanks for the trick @Khady! I'm not going to use it here since merlin is the only dev dep, so I don't think it's worth the trouble. I did update the readme though.

I'll merge now, thanks everyone!

PR #938, opened 2019-03-11
LSP: return doc on hover

Khady:
If documentation is available, add it to the hover response.

I wonder if the doc should be surrounded with `(** *)` so that it gets nice highlighting.

### DIFF
--- a/src/frontend/lsp/ocamlmerlin_lsp.ml
+++ b/src/frontend/lsp/ocamlmerlin_lsp.ml
@@ -225,16 +225,24 @@ let on_request :
       | (location, `String value, _) :: _ -> Some (location, value)
     in
 
-    let format_contents ~as_markdown typ =
+    let query_doc doc pos =
+      let command = Query_protocol.Document (None, pos) in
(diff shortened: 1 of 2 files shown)


andreypopp:
I've tried doing something similar but at the end decided not to proceed (but the reasons are not sound): in vim there's little space now to show the type info (cmdline only) and so I like when it does show only a minimal info on hover. I think now though we can provide a different UI as floating windows are merged into neovim.

So my question — can we make it configurable or also does it make sense to do so?

Khady:
A configuration is probably necessary indeed.

In merlin-eldoc that does a bit the same thing, I offered different settings. Maximum number of lines, display type and/or doc, wrapping the type on one line if it fits, etc. Do we want something as powerful or just a boolean to disable the doc?

Where should the doc be and how can it be transmitted to the server?

andreypopp:
Thanks!

(discussed offline: we decided to merge this now and refine later when we decide on configuration options, having docs in hover responses is just too useful)

PR #939, opened 2019-03-12
Emacs: Don't offer symbol completion inside comments or strings

Wilfred:
This is rarely useful and can hurt editing performance.

### DIFF
--- a/emacs/merlin-company.el
+++ b/emacs/merlin-company.el
@@ -80,6 +80,11 @@
                " (successive calls will expand aliases)"))
       (t default))))
 
+(defun merlin-company--in-string-or-comment-p ()
+  "Return non-nil if START is in a comment or string."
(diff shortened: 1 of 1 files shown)


emacs/merlin-company.el
```diff
@@ -80,6 +80,11 @@
                " (successive calls will expand aliases)"))
       (t default))))
 
+(defun merlin-company--in-string-or-comment-p ()
+  "Return non-nil if START is in a comment or string."
+  (let ((state (syntax-ppss)))
+    (or (nth 3 state) (nth 4 state))))
```
trefis: That is completely indecipherable to me.
Is there no nicer way to do this?

If not, could we at least get some comments / documentation?

dbuenzli:
What does that change exactly ? Symbol completion is *definitively* useful when you write documentation.

Issue #940, opened 2019-03-12
confusing error message: "The value ... is not an instance variable"


gasche:
``` ocaml
let test foo =
  match foo with
  | None -> exit 1
  | Some pkg -> pkg (*< ERROR *)
end (* < ERROR *)
```

The error is due to `pkg end` (`end` should be `in`); Merlin says:

```
The value pkg is not an instance variable
```

trefis:
Here is the AST that merlin is working on in that case:
```ocaml
let test foo =
  match foo with
  | None -> ((exit 1)[@merlin.loc ])
  | Some pkg -> ((pkg <- [%merlin.hole ])[@merlin.loc ])
```

Obtained with: `ocamlmerlin single dump -what parsetree < gasche.ml | jq .value | sed 's/"\([^"]*\)"/\1/' | sed 's/\\n/\n/g'`.

I don't know why it chose to introduce that `<-`, I'll let @let-def comment.

let-def:
:) unfortunate choice of the recovery algorithm. I will tell it to avoid `<-`.

PR #941, opened 2019-03-13
Lessminus recover

let-def:
This PR address bug #940. The bug had two parts:
- computation of recovery costs was incorrect for non-parameterized tokens
- `<-` was picked because it had a null recovery cost (making it a favorite candidate for the parser), which led to problematic recovery decisions.

The first fix might affect other recovery decisions. I will test this branch for some time to ensure this does not introduce any regression.


let-def:
The files in this commit were preprocessed with Clang rather than GCC.
We use `-P` to avoid generation of preprocessing directives in the output (`#line`, etc).

Their behavior differ: GCC removes blank lines while Clang preserves them.
I will regenerate the commit with GCC to keep the output close the existing state.

trefis:
I rebased the branch, and made the test slightly more stable (I have some more robust changes on my 4.08 branch to make this kind of tests more stable, but didn't want to pollute this PR with them).

I assume if there was an obvious regression @let-def would have caught it by now, so I will merge this soon, and we'll see if users complain.

PR #944, opened 2019-03-14
LSP: optionally return hierarchical document outline

andreypopp:
We check for hierarchicalDocumentSymbolSupport capability and returns a list of
document symbols instead of a flat list of symbol infos.

TODO:

- [ ] Check if we need to return outline in the correct order
- [ ] Filter out `let _ = ` bindings from outline (those are generated by ppx deriving and are noisy, wondering if those should be removed in original merlin protocol too cc @trefis)

### DIFF
--- a/src/frontend/lsp/ocamlmerlin_lsp.ml
+++ b/src/frontend/lsp/ocamlmerlin_lsp.ml
@@ -321,51 +321,80 @@ let on_request :
     return (store, lsp_locs)
 
   | Lsp.Rpc.Request.DocumentSymbol {textDocument = {uri;}} ->
+    let module DocumentSymbol = Lsp.Protocol.DocumentSymbol in
+    let module SymbolKind = Lsp.Protocol.SymbolKind in
+
     Document_store.get store uri >>= fun doc ->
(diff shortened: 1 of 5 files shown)


src/frontend/lsp/ocamlmerlin_lsp.ml
```diff
@@ -321,51 +321,80 @@ let on_request :
...
+        detail = item.Query_protocol.outline_type;
+        kind = kind item;
+        deprecated = false;
+        range = range;
+        selectionRange = range;
+        children;
+      }
+    in
+
+    let rec symbol_info ?containerName item =
```
Khady: I'd move this function in addition to `kind`, `range` and `symbol` before `Document_store.get store uri`. My logic is that `Document_store.get store uri` usually marks the start of the computation part. And those functions are helpers that can live before.

andreypopp: Ageed! Thanks.

andreypopp:
Going to merge this now (after CI) and address todos later (I think theese should be fixed in original `outline.ml`).

PR #947, opened 2019-03-23
LSP: fix conversion from LSP position to merlin position

Khady:
LSP positions are starting from line 0 column 0. Merlin positions are
starting from line 1 column 0.


andreypopp:
Thanks!

Issue #937, opened 2019-03-11
LSP: replace ppx_deriving_yojson?


Khady:
merlin-lsp relies on ppx_deriving_yojson for the (de-)serialization of messages in the protocol. One downside is that is not possible to use dune + ppx_deriving_yojson on ocaml 4.02.3, at least with the constraints that are in opam-repository (I have a branch that compiles on 4.02.3 https://github.com/Khady/merlin/commit/9325d1d9e4f2d3de0106a1a51d9479c0326bebb0).

Should merlin-lsp be able to support 4.02.3 and bucklescript? If yes, what is the best solution?

1. The messages could be parsed by hand, if the goal is to limit the number of dependencies. The json objects are not terribly complicated. But it is tedious, repetitive and error prone.
2. A tool like atdgen could be used. One good point with atdgen is that the output could be committed to the repository. In this situation, atdgen would only be a dev dependency.
3. Something I didn't think about?

Maybe it's just better to stick to the current situation and wait for bucklescript to add support for 4.06.

cc @andreypopp @bryphe @rvantonder @rusty-key who might have an opinion.

andreypopp:
I think even if BS adds support for 4.06 we still want to get rid of `ppx_deriving_yojson` simply because it's introduces so many deps into opam switch / esy project which has merlin (and merlin is used in 99% projects I've seen).

I'm voting for atdgen — #936 does something similar with menhir and thanks to dune's promote feature it doesn't sound too difficult to do (maybe I'm missing something though as I haven't used atdgen yet).

trefis:
The other option you could consider is keep using `ppx_deriving_yojson`, but only in dev mode, using `[@@deriving_inline]`.

Base uses that workflow, it is described [on its README](https://github.com/janestreet/base#base-and-ppx-code-generators).

Khady:
deriving_inline would be great. Support is actually available in dune. But it seems to to only handle plugins based on ppxlib. So not ppx_deriving_yojson. I'm almost tempted to write a yojson deriver based on ppxlib at this point. But it's a bit a waste of time.

hcarty:
> I'm almost tempted to write a yojson deriver based on ppxlib

There is https://github.com/andersfugmann/ppx_protocol_conv/tree/master/drivers/json

I'm not sure how it compares to ppx_deriving_yojson in practice.

Khady:
[...]

PR #948, opened 2019-03-28
LSP: migrate to ppx_protocol_conv

Khady:
Beginning of the work on #937.

So I started to translate the code from ppx_deriving_yojson to ppx_protocol_conv_json. There are some good things. For example it can translate a constructor with no argument to a string. So there is no need to write the conversion functions by hand in this case. On the other hand, it uses exception instead of the result type, which is a bit sad and annoying. So either the code must be adapted to catch the exceptions. Or we can ask upstream if using the result type would be possible.

The code doesn't compile yet but I haven't had time to work on it in the past few days. Pushing it here in case someone wants to help.

### DIFF
--- a/merlin-lsp.opam
+++ b/merlin-lsp.opam
@@ -20,7 +20,7 @@ depends: [
   "dune"  {build & >= "1.8.0"}
   "ocamlfind" {>= "1.5.2"}
   "yojson"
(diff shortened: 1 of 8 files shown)


Khady:
I have the impression that ppx_protocol_conv brings a dependency on compiler-libs. The exact reason is not clear to me. But not much success so far.

PR #956, opened 2019-05-07
Add support for 4.08

trefis:
:expressionless: :gun:

N.B. we are currently **not** using the new `Load_path` introduced in OCaml 4.08.
We might have to reconsider that choice soon when: `-I foo/ -open Foo -I bar/ -open Bar` and `-I foo/ -I bar/ -open Foo -open Bar` start behaving differently (if both Foo and bar/ contain a `Bar`).
But that's a problem for future us.


trefis:
One thing to note: I tested 4.08 locally be pinning a non-released (non merged even) version of mdx.
But I didn't set anything up in travis.

PR #960, opened 2019-05-13
lsp: Trim the list of completion results to 100

wokalski:
Some LSP clients might choke on too many results from the server during
parsing and processing. It's certainly the case in my Vim setup where
whole processing happens synchronously.

### DIFF
--- a/src/frontend/lsp/ocamlmerlin_lsp.ml
+++ b/src/frontend/lsp/ocamlmerlin_lsp.ml
@@ -632,8 +632,14 @@ let on_request :
         List.map (fun entry -> `Keep entry) entries
     in
     let all = List.concat [labels; items] in
-    let items = List.mapi item all in
-    let resp = {Lsp.Protocol.Completion. isIncomplete = false; items;} in
+    (* max_results represents the maximum number of completion items
+     * that we send to the client. If we send too many results, clients
+     * might choke on parsing and processing. *)
+    let max_results = 100 in
+    let is_incomplete = List.length all > max_results in
+    let trimmed = if is_incomplete then Std.List.take_n max_results all else all in
+    let items = List.mapi item trimmed in
+    let resp = {Lsp.Protocol.Completion. isIncomplete = is_incomplete; items;} in
     return (store, resp)
 
   | Lsp.Rpc.Request.TextDocumentRename { textDocument = { uri }; position; newName } ->


trefis:
Naïve question: can't your client be fixed instead?

wokalski:
Of course it could, I could modify my whole setup, get a stronger machine, or perhaps introduce asynchrononicity in vimscript. On a more serious note, I think supporting this is reasonable. Some completion queries resulted in 4000 line long json responses and it choked on those. If there's further typing an LSP client should make another query to LSP.

andreypopp:
I don't think it can be solved for some clients (namely ALE which is implemented in vimscript) and I think it's reasonable thing to trim the number of completion results to some sensible number (100 sounds ok to me).

@trefis what downsides you see with this approach?

wokalski:
When I implemented this I wondered if there are some clients which do interesting client side preprocessing of the completion result list. This approach is limiting in a sense that there's no way to set the limit in the LSP which could potentially limit the functionality of such clients.

That said I'm not aware of existence of any concrete use cases of this.

Khady:
100 results means it is not possible to fully explore a module such as `Base.List`. But I don't know if there are many big modules like that.

wokalski:
@Khady Are there use cases where users explore big/huge modules like this in the completion list without any typing in between? If yes, maybe this is not the most fortunate solution indeed.

Khady:
I usually explore modules like this. Because I am very lazy and it's faster than opening utop and typing `#show Base.List`. There are workarounds, so this is not critical to me.

trefis:
> On a more serious note, I think supporting this is reasonable.

> I think it's reasonable thing to trim the number of completion results to some sensible number

I'm not saying it isn't reasonable, but I'm really not sure it is.

For one thing 100 seems like a really slow number, I'm pretty sure I regularly make requests that return over a 100 candidates (@Khady's example is a good one).

Also, I have a real dislike for magic constants.

But more importantly: merlin has been around for years, and has been used by dozens of people daily for many of those years. And I don't really remember people complaining about completion until now (apart if also asking for documentation at the same time, but that's a different issue). Also fwiw, I also use vim.
So what is different here? If it's one of the plugins you're using which dies one the amount of data merlin returns, I'm sorry but yes, that plugin should be fixed.

> There are workarounds, so this is not critical to me.

It is to me.

andreypopp:
> apart if also asking for documentation at the same time, but that's a different issue

I'm afraid we are asking for documentation in LSP frontend now by default.

wokalski:
@trefis I see, interesting, I wasn't aware there are people who use completion in this manner.

@andreypopp Interesting, LSP's documentation explicitly mentions expensive documentation computation as an example so it seems like a commonplace problem in this space. Should we skip documentation by default and implement `completionItem/resolve`?

let-def:
I am digressing, but retrieving documentation is slow and I don't see when we will have the time to improve on that. It would better be done lazily.

About trimming the list of completions to 100... I agree with @trefis, it is better to fix the LSP client.

trefis:
> I'm afraid we are asking for documentation in LSP frontend now by default.

It's not necessarily bad, as long as you give the option to not do it.
Also, the code in this PR wouldn't improve things if that were the cause of the slowness.

Considering the discussion so far, I'm closing this PR.
If you want to discuss this further, I'd suggest opening an issue, and giving more specific details about the context, where the slowness comes from, possibly some numbers as well, etc.

andreypopp:
@wokalski yeah, delating documentation request till the `completionItem/resolve` request seems like the better approach to me.

wokalski:
@trefis I understand, thanks for the feedback. 

Skipping the documentation in the response hasn't improved the perceived performance (i.e. the bottleneck is not merlin's speed but parsing speed on the client). I'll try to see if anything can be done in my client.

PR #961, opened 2019-05-21
Emacs: Allow users to disable completion inside comments and strings

Wilfred:
This is the default behaviour for many company backends, so provide an
option to make company-merlin behave similarly.

----

This is a follow-up to #939. I apologise for not responding last time. This does not change the default, but I'd really like the option to disable merlin completion inside strings.


Wilfred:
Thank you :)

PR #974, opened 2019-06-18
Install vim/ftplugin/reason.vim

hcarty:
This file has been in the source tree for a little while now but it was not being installed.


trefis:
Thanks @hcarty !

PR #981, opened 2019-06-24
Emacs fixes: log arguments in merlin-debug-last-commands, compute line offsets correctly 

### LINKED ISSUE #975: merlin-locate is confused by tabs (not a valid identifier)
~~~ocaml
let foo = 1
	let baz = foo
~~~
(the leading space before `let baz` is a tab character)
In emacs, place point anywhere in the second `foo` and call `merlin-locate`, it will say `not a valid identifier`.

let-def:
The first patch will help debugging emacs related problems.
The second patch fixes #975: `(current-column)` is a "visual notion" (it depends on the width of tabs), which makes sense only inside emacs, while `(- (point) (line-beginning-position))` will compute the offset as a number of character.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -529,7 +529,7 @@ return (LOC1 . LOC2)."
                  args))
     ;; Log last commands
     (setq merlin-debug-last-commands
-          (cons (cons binary args) merlin-debug-last-commands))
+          (cons (cons (cons binary args) nil) merlin-debug-last-commands))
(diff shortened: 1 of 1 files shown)


let-def:
I just realized that it is probably wrong if we have multiple byte characters in the line, for instance in a string or a comment :'(

let-def:
Changing the example to:
```ocaml
let foo = 1
(*おねがいします*)	let baz = foo
```
broke it again.
The last patch fixes that case.

PR #986, opened 2019-07-01
Browse tree: improve support for letop

trefis:
It's still somewhat clunky.


let-def:
The new test only works with 4.08, right?
(Should we add support for let-op in older version of OCaml, like the future syntax shims of dune).

trefis:
> The new test only works with 4.08, right?

That's right, I told myself I should add a `enabled_if ...` thingy in the dune rule, but then forgot.
Will do.

> (Should we add support for let-op in older version of OCaml, like the future syntax shims of dune).

I'm very tempted not too. But I guess it doesn't cost us much.
How about we wait until someone complains?

let-def:
> How about we wait until someone complains?
Yes.

trefis:
I added the `(enabled_if ..)` condition.
I think we could merge this as is; it's not perfect, but I don't have time to improve it right now and it fixes the immediate issue.

Issue #972, opened 2019-06-16
Large stacktraces in Emacs when upgrading to Ocaml 4.08.0


benkonz:
I'm getting large stack traces in Emacs when I hover over errors and try to autocomplete some of the modules added in Ocaml 4.08 (`Option`, `Int`, etc.). 

<img width="1677" alt="image" src="https://user-images.githubusercontent.com/18561476/59565494-64722200-9022-11e9-9757-0061624b6fc8.png">

Is there a way to fix this?

benkonz:
I'm pretty sure this was getting caused by conflicting installs of `merlin`. Removing the old install of `merlin` fixed the stack traces.

aatxe:
I seem to be having this same issue, but I don't think I have conflicting installs of `merlin`. I even uninstalled the one on my system switch to be safe.

benkonz:
Interesting, I'll re-open this issue

aatxe:
Can you clarify how you ended up with conflicting installs and what you did to rectify that?

benkonz:
I used `opam switch create 4.08.0` to install the new compiler while also having version 4.07.1 as my default switch and `merlin` was installed for both compiler switches.

aatxe:
I removed my old 4.07.1 switch entirely just now, and it's still reproducing.

Here's the full stack trace (since it's not totally visible in your screenshot):
```
Error from syntax checker ocaml-merlin: unknown answer: "exception":"Syntaxerr.Error(_)
Raised at file \"parser_raw.mly\", line 176, characters 4-65
Called from file \"parser_raw.mly\", line 1913, characters 5-48
Called from file \"src/utils/menhirLib.ml\", line 1885, characters 18-44
Called from file \"src/utils/std.ml\", line 93, characters 17-25
Called from file \"src/kernel/mreader_recover.ml\", line 210, characters 10-57
Called from file \"src/kernel/mreader_recover.ml\", line 248, characters 28-38
Called from file \"src/kernel/mreader_parser.ml\", line 130, characters 21-35
Called from file \"src/kernel/mreader_parser.ml\", line 175, characters 20-48
Called from file \"src/kernel/mreader_parser.ml\", line 190, characters 6-42
Called from file \"src/ocaml/utils/misc.ml\", line 30, characters 8-15
Re-raised at file \"src/ocaml/utils/misc.ml\", line 42, characters 10-24
Called from file \"src/kernel/mreader_parser.ml\", line 204, characters 20-56
[...]

PR #987, opened 2019-07-01
4.08: Fix exceptions leaking from Parser

let-def:
This is a tentative fix for #972.


Makefile
```diff
@@ -6,6 +6,10 @@ build:
 workspace:
 	dune build --workspace=dune-workspace.template
 
+workspace-ocamlmerlin:
+	dune build --workspace=dune-workspace.template \
+		src/frontend/ocamlmerlin_server.exe
```
trefis: This feels out of place in this PR. Also, I'd suggest building "merlin.install" instead of "src/frontend/ocamlmerlin_server.exe").

let-def: Right.

src/ocaml/preprocess/dune
```diff
@@ -1,10 +1,26 @@
...
+  if need_cpp then
+    ("*.mlyp",
+     "(rule\n\
+       (targets parser_raw.mly)\n\
+       (deps    parser_raw.mlyp)\n\
+       (action\n\
+         (with-stdout-to %{targets}\n\
+           (run gcc -E -x c -P %{deps}))))")
+  else 
+    ("*.mly", "")
```
trefis: Cute.

let-def:
This patch:
- removes all `error` rules in the grammar (they use the built-in error handling system of yacc which conflicts with merlin recovery algorithm)
- fixes a few conflicts in the grammar (inlining of `match_cases`, precedence annotation for `LET_LWT`)
- replaces all (most) exceptions thrown by a call to `Msupport_parsing.raise_error` to register the error and then returns a sane fallback value.

PR #991, opened 2019-07-04
ignore Ident.rename failures

let-def:
This issue appears when completing an empty identifier, e.g. `let foo = ‸` or `let foo = List.‸`.
I haven't diagnosed the root cause for the failure, but just silencing this `fatal_error` seems fine for now.


trefis:
Actually, can we make the caller of that function catch the exception and return the initial ident?

If that function fails in other contexts, I'd like to know, as it could also affect upstream.

PR #992, opened 2019-07-04
Parenthesize

let-def:
Identifiers that are infix operators (`+`, `::`, ... `mod`) are no longer parenthesized during completion.
This PR improves that.

### DIFF
--- a/src/ocaml/typing/403/oprint.ml
+++ b/src/ocaml/typing/403/oprint.ml
@@ -30,17 +30,19 @@ let rec print_ident ppf =
   | Oide_apply (id1, id2) ->
       fprintf ppf "%a(%a)" print_ident id1 print_ident id2
 
-let parenthesized_ident name =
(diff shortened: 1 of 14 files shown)


src/ocaml/typing/403/oprint.ml
```diff
@@ -30,17 +30,19 @@ let rec print_ident ppf =
...
+    match name.[0] with
+    | 'a'..'z' | 'A'..'Z' | '\223'..'\246' | '\248'..'\255' | '_' -> false
+    | _ -> true
 
 let value_ident ppf name =
   if parenthesized_ident name then
-    fprintf ppf "( %s )" name
+    if name <> "" && (name.[0] = '*' || name.[String.length name - 1] = '*')
+    then fprintf ppf "( %s )" name
+    else fprintf ppf "(%s)" name
```
trefis: Why do that extra work here (and in 4.02) but not in any of the following versions?

let-def: Following versions handle that differently, but have the right behavior (otherwise the tests would fail).

PR #996, opened 2019-07-09
Allow `esy` as merlin-command to call merlin via esy exec-command

actionshrimp:
The same idea as the `'opam` merlin command, however `esy exec-command` is a lot quicker so we can call it directly rather than grabbing it's config and setting the PATH.

This means we have to allow `merlin-command` to return more than just the binary name as we need the additional base args - hopefully I've updated all the call sites that depend on that but would appreciate a second pair of eyes!


andreypopp:
Don't use emacs so don't really know if this is the correct way to implement this but the idea is sound — I used the same in vim before I switched to merlin-lsp.

PR #1000, opened 2019-07-17
Fix merlin-imenu

artempyanykh:
That's what JSON look like for outline:
```
{
      "start": { "line": 181, "col": 0 },
      "end": { "line": 213, "col": 48 },
      "name": "pp_all",
      "kind": "Value",
      "type":
        "...",
      "children": []
    }
```

`children` are now at idx 6. Current code fails with `Wrong type argument: listp` when tries to call `parse-outline-tree` on a `type` string.


trefis:
Thanks!

PR #1001, opened 2019-07-17
Fix merlin-xref

artempyanykh:
When `loc.file` is **not** the current file, `(merlin/make-point pos)` tries to find the requested pos in the current file, may fail and return null. This null propagated to `merlin-xref--line` causes an error `Wrong type argument: integer-or-marker-p, nil`

So, let's switch to the correct file before invoking `(merlin/make-point)`.


artempyanykh:
Not sure why the badge shows that the build is in progress — when I follow the Details link it shows that the build is passing.

PR #1002, opened 2019-07-19
Fix merlin-lsp 4.08 compilation

mlasson:
This PR do a few fixes to allow the merlin-lsp package to compile on 4.08 (it's now possible since deriving yojson has been ported to 4.08).

### DIFF
--- a/merlin-lsp.opam
+++ b/merlin-lsp.opam
@@ -12,7 +12,7 @@ build: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
-  "ocaml" {>= "4.04.0" & < "4.08"}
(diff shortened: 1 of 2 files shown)


XVilka:
Shouldn't be this one merged?

andreypopp:
There's currently work in progress by @zindel to get rid of ppx dependency and make it compile on from 4.02 till 4.08. I think those fixes should go there (if not already).

Khady:
The diff looks unrelated to ppx. So I think this change can be integrated independently.

zindel:
@mlasson, just cherry-picked your commit into https://github.com/rusty-key/merlin/tree/ppx-yojson-conv. Worked well with 4.02, 4.07, 4.08

andreypopp:
Thank you!

PR #1004, opened 2019-07-24
Emacs: reset configuration in all buffers when stopping the merlin server

Wilfred:
Otherwise, restarting the server after changing `merlin-command` has no effect, because the old value for `merlin-command` is cached in `merlin-buffer-configuration`.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -571,8 +571,11 @@ return (LOC1 . LOC2)."
   (unless merlin-mode (message "Buffer is not managed by merlin."))
   (when merlin-mode
     (merlin--call-merlin "stop-server")
-    (setq merlin-erroneous-buffer nil)
-    (setq merlin-buffer-configuration nil)))
+    ;; These are buffer-local variables, so reset them in all buffers.
+    (dolist (buf (buffer-list))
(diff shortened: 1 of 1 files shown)


trefis:
That looks right to me, but I'm not confident enough to press the merge button.
@rgrinberg: would you mind having a look?

rgrinberg:
Looks good to me.

trefis:
Thanks all!

emacs/merlin.el
```diff
@@ -571,8 +571,11 @@ return (LOC1 . LOC2)."
   (unless merlin-mode (message "Buffer is not managed by merlin."))
   (when merlin-mode
     (merlin--call-merlin "stop-server")
-    (setq merlin-erroneous-buffer nil)
-    (setq merlin-buffer-configuration nil)))
+    ;; These are buffer-local variables, so reset them in all buffers.
+    (dolist (buf (buffer-list))
+      (with-current-buffer buf
+        (kill-local-variable 'merlin-buffer-configurationn)
```
fangyi-zhou: I think you have an extra `n`

trefis: Good catch, I pushed a fix.

Issue #977, opened 2019-06-20
safe-string is not applied in OCaml versions where it's the default


copy:
Merlin only emits safe-string errors when given a `FLG -safe-string` directive. On newer OCaml versions safe-string is the default, and merlin should also apply it.

PR #1008, opened 2019-08-06
Enable safe-string if it's the default in the current version of OCaml

copy:
Fixes https://github.com/ocaml/merlin/issues/977


trefis:
Thanks!

Issue #978, opened 2019-06-20
Handle warn-error


copy:
I tend to run merlin with `merlin_ignore_warnings = 'true'`, but this hides errors that were originally warnings. It would be nice if merlin could turning warnings from `warn-error` into errors.

PR #1009, opened 2019-08-06
vim: Display warn-error warnings as errors

copy:
Fixes https://github.com/ocaml/merlin/issues/978

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -12,6 +12,7 @@ atom_bound = re.compile('[a-z_0-9A-Z\'`.]')
 re_wspaces = re.compile("[\n ]+")
 re_spaces = re.compile(" +")
 re_spaces_around_nl = re.compile(" *\n *")
+re_error_warning = re.compile(r"Error \(warning (\d+)\): ")
 
 protocol_version = 3
 
@@ -409,6 +410,8 @@ def vim_loclist(vimvar, ignore_warnings):
             msg = msg[8:]
         elif msg.startswith("Error: "):
             msg = msg[7:]
+        elif msg.startswith("Error (warning"):
+            msg = re.sub(re_error_warning, r"\1: ", msg)
         lnum = 1
         col = 1
         if 'start' in error:
--- a/src/frontend/ocamlmerlin/query_json.ml
+++ b/src/frontend/ocamlmerlin/query_json.ml
@@ -228,15 +228,6 @@ let json_of_type_loc (loc,desc,tail) =
   ]
 
 let json_of_error (error : Location.error) =
-  let typ =
-    match error.source with
-    | Location.Lexer   -> "lexer"
-    | Location.Parser  -> "parser"
-    | Location.Typer   -> "typer"
-    | Location.Warning -> "warning"
-    | Location.Unknown -> "unknown"
-    | Location.Env     -> "env"
-  in
   let of_sub loc sub =
     let msg =
       Location.print_sub_msg Format.str_formatter sub;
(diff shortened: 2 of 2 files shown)


copy:
Gentle nudge @trefis

vim/merlin/autoload/merlin.py
```diff
@@ -399,16 +400,18 @@ def vim_loclist(vimvar, ignore_warnings):
     bufnr = vim.current.buffer.number
     nr = 0
     for error in errors:
-        ty = 'E'
-        if error['type'] == 'warning':
-            if vim.eval(ignore_warnings) == 'true':
-                continue
-            ty = 'W'
```
trefis: I don't understand that part of the diff, why did you delay that check? Aren't you just adding useless extra work?

vim/merlin/autoload/merlin.py
```diff
@@ -399,16 +400,18 @@ def vim_loclist(vimvar, ignore_warnings):
...
+        ty = 'W' if error['type'] == 'warning' else 'E'
         msg = re.sub(re_wspaces, " ", error['message']).replace("'", "''")
         if msg.startswith("Warning "):
             msg = msg[8:]
         elif msg.startswith("Error: "):
             msg = msg[7:]
+        elif msg.startswith("Error (warning"):
+            msg = re.sub(re_error_warning, r"\1: ", msg)
+            ty = 'E'
```
trefis: Is that line really necessary?
If it is, shouldn't it instead be fixed in the backend (to make sure the `type` field is set properly)?

copy: The stripping is in line with the stripping of the error and warning prefixes in the lines above (to keep the loclist concise).

Do you think the type of these warnings should be changed to an error in the backend? If so, I'd gladly include that change in this PR.

vim/merlin/autoload/merlin.py
copy: To prevent skipping over warn-error'd warnings when `ignore_warnings` is set.

vim/merlin/autoload/merlin.py
trefis: I was only talking about the `ty = 'E'` line.

> Do you think the type of these warnings should be changed to an error in the backend? If so, I'd gladly include that change in this PR.

Yes.

vim/merlin/autoload/merlin.py
trefis: That part can then be reverted if the other change is implemented in the backend, as discussed.

src/frontend/query_json.ml
```diff
@@ -249,6 +241,19 @@ let json_of_error (error : Location.error) =
...
     String.trim (Format.flush_str_formatter ())
   in
+  let typ =
+    match error.source with
+    | Location.Lexer   -> "lexer"
+    | Location.Parser  -> "parser"
+    | Location.Typer   -> "typer"
+    | Location.Warning ->
+      if String.is_prefixed ~by:"Error" msg then
+        "typer" (* Handle warn-error (since 4.08) *)
```
copy: I tried doing this earlier (in mtyper/msupport), but couldn't make it work across multiple versions of OCaml.

trefis:
Seems reasonable to me, let's see what @let-def thinks of the second commit.

trefis:
@let-def having disappeared from the face of the earth, I'll be happy to merge this once you rebase it @copy.
(Sorry for the very slow process)

copy:
@trefis Rebased.

trefis:
Thanks!

PR #1013, opened 2019-08-21
Enable merlin-imenu to show type information

taquangtrung:
Hi,

I improved merlin-imenu to show also the type information of expressions. The demo can seen in the attached picture, by calling the function `counsel-imenu` from Spacemacs.

Thanks to @artempyanykh in this PR https://github.com/ocaml/merlin/pull/1000 for pointing out that the recent version of Merlin can return type information.

merlin-imenu


taquangtrung:
Thanks for the suggestion! I have updated the code.

rgrinberg:
I gave this a try and the types seem to be truncated in the imenu. Here's a [screenshot](https://www.evernote.com/l/AaETv3jpzlxEEqxjKTaNZ66PCb5qPOR3Hzk) to demonstrate what I mean. Any idea what might be causing this?

trefis:
> Here's a screenshot 

Side comment: I'm slightly surprised by the use of `/` instead of `.` in ocaml paths.

taquangtrung:
> I gave this a try and the types seem to be truncated in the imenu. Here's a [screenshot](https://www.evernote.com/l/AaETv3jpzlxEEqxjKTaNZ66PCb5qPOR3Hzk) to demonstrate what I mean. Any idea what might be causing this?

Oh, that's because there is a variable `imenu-max-item-length` of imenu that controls it. Its default value is 60. In my Emacs, I setting to 200: `(setq imenu-max-item-length 200)`

taquangtrung:
x

> > Here's a screenshot
> 
> Side comment: I'm slightly surprised by the use of `/` instead of `.` in ocaml paths.

Thanks! It actually makes sense. I will update it.

emacs/merlin-imenu.el
```diff
@@ -1,10 +1,12 @@
 ;;; merlin-imenu.el --- Merlin and imenu integration.   -*- coding: utf-8 -*-
 ;; Licensed under the MIT license.
 
-;; Author: tddsg
-;; Created: 10 July 2016
-;; Updated: 27 April 2017
-;; Version: 0.2
+;; Author: Ta Quang Trung
```
rgrinberg: Could you append your own name rather than removing the old name?

emacs/merlin-imenu.el
```diff
@@ -38,64 +35,60 @@
...
-          ((string= item-kind "Exn")
-           (setq exception-list (cons item-marker exception-list)))
-          ((string= item-kind "Constructor")
-           (setq constructor-list (cons item-marker constructor-list)))
-          ((string= item-kind "Label")
-           (setq label-list (cons item-marker label-list)))
-          (t (setq misc-list (cons item-marker misc-list))))
-    (if (not (null sub-trees))
-        (parse-outline-tree (concat prefix item-name " / ") sub-trees))))
+  ;; (message "Item: %s" item)
```
Wilfred: Debugging code here.

Wilfred:
I haven't tried running the code, but it looks sensible to me :)

emacs/merlin-imenu.el
taquangtrung: Deleted.

taquangtrung:
@Wilfred @rgrinberg  I updated the new code!

@rgrinberg: For the threshold on item's length of `imenu`, do you think if I should put the new threshold in this module, like `(setq imenu-max-item-length 200)`. What I am afraid is that it may affects other modules using `imenu`.

rgrinberg:
@taquangtrung I would say that we shouldn't bother. It's possible that we'll overwrite a user set value, and this problem is not really specific to merlin-imenu.

emacs/merlin-imenu.el
```diff
@@ -1,10 +1,12 @@
 ;;; merlin-imenu.el --- Merlin and imenu integration.   -*- coding: utf-8 -*-
 ;; Licensed under the MIT license.
 
-;; Author: tddsg
-;; Created: 10 July 2016
-;; Updated: 27 April 2017
-;; Version: 0.2
+;; Author: tddsg (Ta Quang Trung)
```
rgrinberg: Ah, I wasn't aware you were the initial author. Feel free to erase the username and just keep your real name if you like.

taquangtrung: Thanks! But actually I am OK with your suggestion, so I think that I can keep the current information as it is.

Wilfred:
@tuquangtrung If you let bind imenu-max-item-length you should be able to override it only inside merlin-imenu. defvar in elisp is dynamically scoped.

taquangtrung:
> @tuquangtrung If you let bind imenu-max-item-length you should be able to override it only inside merlin-imenu. defvar in elisp is dynamically scoped.

@Wilfred: This is a very good advice! I updated the new code. Can you review if it is OK?

Wilfred:
Oh, using a buffer-local variable is a good idea!

I was thinking that you'd just do `(let ((imenu-max-item-length 300)) ...)` but you'd need to let-bind when `imenu--truncate-items` is called. That doesn't work here because `imenu-create-index-function` is called before, so a let binding inside your imenu function wouldn't do anything.

rgrinberg:
Question: if a user has already set this setting, will this override it?

taquangtrung:
> Question: if a user has already set this setting, will this override it?

This modification (set imenu-max-item-length to 300) is performed only in OCaml buffer, i.e., when merlin-imenu is enabled. If the user's setting is less than 300, then I think that increasing it to 300 will make no harm. For safer, I can increase the threshold to a value like 500?

trefis:
> @taquangtrung I would say that we shouldn't bother. It's possible that we'll overwrite a user set value, and this problem is not really specific to merlin-imenu.

I very much agree with @rgrinberg on this.

rgrinberg:
> This modification (set imenu-max-item-length to 300) is performed only in OCaml buffer, i.e., when merlin-imenu is enabled. If the user's setting is less than 300, then I think that increasing it to 300 will make no harm. For safer, I can increase the threshold to a value like 500?

Let's not touch this value at all. I think that this has the potential for side effects that the user can't really foresee and it does not offer that much benefit when it works. If anything, we could could just change the display in imenu to show the types are being truncated with ellipsis.

taquangtrung:
> Let's not touch this value at all. I think that this has the potential for side effects that the user can't really foresee and it does not offer that much benefit when it works. If anything, we could could just change the display in imenu to show the types are being truncated with ellipsis.

Using ellipsis also works. Or I can just simply let `imenu` automatically truncates, and merlin-imenu will prints a message to inform users that their length-threshold is too short (when < 100)?

rgrinberg:
> Or I can just simply let imenu automatically truncates, and merlin-imenu will prints a message to inform users that their length-threshold is too short (when < 100)?

If you want you can add it, but just make sure the message doesn't spam the user. Informing this to the user once is enough. I think this is all gratuitous though so not doing anything is fine as well.

taquangtrung:
> If you want you can add it, but just make sure the message doesn't spam the user. Informing this to the user once is enough. I think this is all gratuitous though so not doing anything is fine as well.

For a second thought, I think that I can report to the `imenu` guys to increase this threshold (currently, it's only 60) so that we don't have to do anything with it.

rgrinberg:
Thanks, I merged after testing.

PR #1019, opened 2019-09-11
merlin-lsp compiles using 4.02 - 4.08

zindel:
This PR is adding the ability of merlin-lsp to be compiled using older versions of OCaml (4.02 & 4.03).

This is done with the help of [ppx_yojson_conv](janestreet/ppx_yojson_conv) & [ppx_yojson_conv_lib](janestreet/ppx_yojson_conv_lib) and the `@deriving_inline` ppx. The code is generated using the newer OCaml version (4.08) and committed to the repo, so that earlier versions of the compiler could handle it without ppx.

This PR is heavily based on the work started by @rusty-key & @andreypopp in [rusty-key/merlin](https://github.com/rusty-key/merlin/tree/ppx-yojson-conv/src)

### DIFF
--- a/src/frontend/lsp/ocamlmerlin_lsp.ml
+++ b/src/frontend/lsp/ocamlmerlin_lsp.ml
@@ -1,3 +1,4 @@
+open Result
 let {Logger. log} = Logger.for_section "ocamlmerlin-lsp"
 
 let initializeInfo: Lsp.Protocol.Initialize.result = {
@@ -255,7 +256,7 @@ let on_request :
       in
       let contents = format_contents ~as_markdown ~typ ~doc in
       let range = Some {
-        Lsp.Protocol. start_ = position_of_lexical_position loc.Warnings.loc_start;
+        Lsp.Protocol. start_ = position_of_lexical_position loc.Location.loc_start;
         end_ = position_of_lexical_position loc.loc_end;
       } in
       let resp = {
@@ -269,10 +270,10 @@ let on_request :
   | Lsp.Rpc.Request.TextDocumentReferences {textDocument = {uri;}; position; context = _} ->
     Document_store.get store uri >>= fun doc ->
     let command = Query_protocol.Occurrences (`Ident_at (logical_of_position position)) in
-    let locs : Warnings.loc list = Query_commands.dispatch (Document.pipeline doc) command in
+    let locs : Location.t list = Query_commands.dispatch (Document.pipeline doc) command in
     let lsp_locs = List.map (fun loc ->
       let range = {
-        Lsp.Protocol. start_ = position_of_lexical_position loc.Warnings.loc_start;
+        Lsp.Protocol. start_ = position_of_lexical_position loc.Location.loc_start;
         end_ = position_of_lexical_position loc.loc_end;
       } in
       (* using original uri because merlin is looking only in local file *)
@@ -316,10 +317,10 @@ let on_request :
   | Lsp.Rpc.Request.TextDocumentHighlight {textDocument = {uri;}; position; } ->
     Document_store.get store uri >>= fun doc ->
     let command = Query_protocol.Occurrences (`Ident_at (logical_of_position position)) in
-    let locs : Warnings.loc list = Query_commands.dispatch (Document.pipeline doc) command in
+    let locs : Location.t list = Query_commands.dispatch (Document.pipeline doc) command in
     let lsp_locs = List.map (fun loc ->
       let range = {
-        Lsp.Protocol. start_ = position_of_lexical_position loc.Warnings.loc_start;
+        Lsp.Protocol. start_ = position_of_lexical_position loc.Location.loc_start;
         end_ = position_of_lexical_position loc.loc_end;
       } in
       (* using the default kind as we are lacking info
@@ -646,12 +647,12 @@ let on_request :
   | Lsp.Rpc.Request.TextDocumentRename { textDocument = { uri }; position; newName } ->
     Document_store.get store uri >>= fun doc ->
     let command = Query_protocol.Occurrences (`Ident_at (logical_of_position position)) in
-    let locs : Warnings.loc list = Query_commands.dispatch (Document.pipeline doc) command in
+    let locs : Location.t list = Query_commands.dispatch (Document.pipeline doc) command in
     let version = Document.version doc in
     let edits = List.map (fun loc ->
       let range =
         {
-          Lsp.Protocol. start_ = position_of_lexical_position loc.Warnings.loc_start;
+          Lsp.Protocol. start_ = position_of_lexical_position loc.Location.loc_start;
           end_ = position_of_lexical_position loc.loc_end;
         }
       in
--- a/src/frontend/lsp/ocamlmerlin_lsp.ml
+++ b/src/frontend/lsp/ocamlmerlin_lsp.ml
@@ -38,97 +38,6 @@ let initializeInfo: Lsp.Protocol.Initialize.result = {
   };
 }
 
-module Document : sig
-  type t
-
-  val make :
-    ?version:int
-    -> uri:Lsp.Protocol.documentUri
-    -> text:string
-    -> unit
-    -> t
-
-  val uri : t -> Lsp.Protocol.documentUri
-  val source : t -> Msource.t
-  val pipeline : t -> Mpipeline.t
-  val version : t -> int
-
-  val update_text : ?version:int -> Lsp.Protocol.DidChange.textDocumentContentChangeEvent -> t -> t
-end = struct
-  type t = {
-    tdoc : Lsp.Text_document.t;
-    source : Msource.t;
-    pipeline : Mpipeline.t;
-    config : Mconfig.t;
-  }
-
-  let normalize_line_endings text =
-    let text = Std.String.replace_all ~pattern:"\r\n" ~with_:"\n" text in
-    text
-
-  let uri doc = Lsp.Text_document.documentUri doc.tdoc
-  let source doc = doc.source
-  let pipeline doc = doc.pipeline
-  let version doc = Lsp.Text_document.version doc.tdoc
-
-  let make_config uri =
-    let path = Lsp.Uri.to_path uri in
-    let mconfig = Mconfig.initial in
-    let path = Misc.canonicalize_filename path in
-    let filename = Filename.basename path in
-    let directory = Filename.dirname path in
-    let mconfig = {
-      mconfig with query = {
-        mconfig.query with
-        verbosity = 1;
-        filename;
-        directory;
-      }
-    } in
-    Mconfig.load_dotmerlins mconfig ~filenames:[
-      let base = "." ^ filename ^ ".merlin" in
-      Filename.concat directory base
-    ]
-
-  let make ?(version=0) ~uri ~text () =
-    let tdoc = Lsp.Text_document.make ~version uri text in
-    (* we can do that b/c all text positions in LSP are line/col *)
-    let text = normalize_line_endings text in
-    let config = make_config uri in
-    let source = Msource.make text in
-    let pipeline = Mpipeline.make config source in
-    {tdoc; source; config; pipeline;}
-
-  let update_text ?version change doc =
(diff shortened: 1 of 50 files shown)


src/frontend/lsp/ocamlmerlin_lsp.ml
```diff
@@ -1,3 +1,4 @@
+open Result
```
Khady: Is it necessary given that you already have `-open Result` in the dune file?

src/lsp/protocol.ml
```diff
@@ -89,21 +705,210 @@ module TextDocumentItem = struct
...
 
 (* DidOpenTextDocument notification, method="textDocument/didOpen" *)
 module DidOpen = struct
-  type params = didOpenTextDocumentParams [@@deriving yojson { strict = false }]
+  type params = didOpenTextDocumentParams
 
   and didOpenTextDocumentParams = {
     textDocument: TextDocumentItem.t;  (* the document that was opened *)
   }
+  [@@deriving_inline yojson]
```
Khady: Not sure if this is necessary. But it's different from what was done before

```suggestion
  [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
```

src/lsp/protocol.ml
```diff
@@ -115,13 +920,298 @@ module DidChange = struct
     rangeLength: int option [@default None]; (* the length that got replaced *)
     text: string; (* the new text of the range/document *)
   }
+  [@@deriving_inline yojson]
```
Khady:
```suggestion
  [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
```

src/lsp/protocol.ml
```diff
@@ -334,6 +1680,380 @@ module PublishDiagnostics = struct
 
   (* legacy FB extension *)
   and relatedLocation = diagnosticRelatedInformation
+  [@@deriving_inline yojson]
```
Khady:
```suggestion
  [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
```

src/lsp/protocol.ml
```diff
@@ -344,16 +2064,16 @@ module Completion = struct
...
 
   let completionTriggerKind_of_yojson = function
-    | `Int 1 -> Ok Invoked
-    | `Int 2 -> Ok TriggerCharacter
-    | `Int 3 -> Ok TriggerForIncompleteCompletions
-    | _ -> Error "invalid completion.triggerKind"
+    | `Int 1 -> Invoked
+    | `Int 2 -> TriggerCharacter
+    | `Int 3 -> TriggerForIncompleteCompletions
+    | _ -> failwith "invalid completion.triggerKind"
```
Khady: why not `yojson_error` here?

src/lsp/protocol.ml
```diff
@@ -516,20 +2238,673 @@ module Completion = struct
...
     TextDocumentPositionParams.t
-    [@@deriving yojson { strict = false }]
 
   and result = hoverResult option [@default None]
 
   and hoverResult = {
     contents: MarkupContent.t;
     range: range option [@default None];
   }
+  [@@deriving_inline yojson]
```
Khady:
```suggestion
  [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
```

src/lsp/protocol.ml
```diff
@@ -699,7 +3923,7 @@ module Initialize = struct
     rootUri: documentUri option [@default None];  (* the root URI of the workspace *)
     client_capabilities: client_capabilities [@key "capabilities"] [@default client_capabilities_empty];
     trace: trace [@default Off];  (* the initial trace setting, default="off" *)
-  } [@@deriving yojson { strict = false }]
+  } [@@yojson.allow_extra_fields]
```
Khady: is it different than putting it near the `@@deriving` annotation?

src/lsp/protocol.ml
```diff
@@ -774,20 +3998,1380 @@ module Initialize = struct
...
 
 (* Goto Definition request, method="textDocument/definition" *)
-module Definition = struct
-  type params = TextDocumentPositionParams.t [@@deriving yojson { strict = false }]
+module Definition = struct [@@@ocaml.warning "-39"]
+  type params = TextDocumentPositionParams.t
 
   and result = DefinitionLocation.t list  (* wire: either a single one or an array *)
-end
+  [@@deriving_inline yojson]
```
Khady:
```suggestion
  [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
```

src/lsp/protocol.ml
```diff
@@ -796,20 +5380,216 @@ module References = struct
...
+let _ = yojson_of_params
+and _ = yojson_of_referenceContext
+and _ = yojson_of_result
+[@@@end]
 end
 
 (* DocumentHighlight request, method="textDocument/documentHighlight" *)
-module TextDocumentHighlight = struct
-  type params = TextDocumentPositionParams.t [@@deriving yojson { strict = false }]
+module TextDocumentHighlight = struct [@@@ocaml.warning "-39"]
```
Khady: I'm surprised that this annotation is necessary only on very few modules that have deriving

Khady:
Looks like the test should be updated. Because optional fields are encoded differently.

sync:
Tried this branch on my bucklescript 5.2.0 project and first of all it's a step forward ocaml 4.02 support, thank you.

But i am having issues with timeout when trying to get autocompletion:

Screen Shot 2019-10-12 at 2 42 49 pm

trefis:
@Khady @zindel: Any news about this PR?

@sync: I doubt this PR has anything to do with the timeout. Let's try and get it merged first, and after that we can open a dedicated issue.

Khady:
@zindel said he will have some time to look at it this Saturday. I'll see if I can help with anything.

src/frontend/lsp/ocamlmerlin_lsp.ml
zindel: Needed here, since I don't modify the dune file in `lsp`.

rgrinberg:
One thing that might be worth discussing is if it's even worth it to support building merlin-lsp on such old versions of OCaml. For dune 2.0 we've decided that dune will build only on relatively recent versions of OCaml. However, this shouldn't impact users of dune that are using an older version of compiler. Essentially, we've decoupled the version of OCaml used to build dune and to use dune build one's own project. Would it be possible to decouple merlin-lsp (and merlin itself) this way?

It just seems like if you require compiling all your tools on such old versions of OCaml, you are setting yourself up for a massive world of pain.

andreypopp:
@zindel needs rebasing on top of current `master`

andreypopp:
@rgrinberg as far as I understand merlin can only work within an OCaml environment of the same version it was built with. Thus in my mind it's useful to keep 4.02 compat at least for bucklescript users.

rgrinberg:
> as far as I understand merlin can only work within an OCaml environment of the same version it was built with. Thus in my mind it's useful to keep 4.02 compat at least for bucklescript users.

Thus I'm wondering what it takes to break this relationship. I know it's quite troublesome to maintain software that needs to build against a wide range of versions of OCaml, particularly when they're as old as 4.02. So I'd like to know what's preventing this approach from being pursued as opposed to this PR.

Khady:
While I do agree that being able to compile merlin with one version and use it with another would be nice, I still believe that this PR is nice to have and hopefully it can be merged soon:

- it migrates to a more modern ppx that is actively maintained
- it provides instant support of 4.02 (which would make the life of many people easier)

rgrinberg:
> it migrates to a more modern ppx that is actively maintained

Indeed. But that can come separately.

> it provides instant support of 4.02 (which would make the life of many people easier)

Is being usable in a branch not instant enough? I would think that users of 4.02 should be experienced enough to use a custom branch

Khady:
> I would think that users of 4.02 should be experienced enough to use a custom branch

I'd say that this is exactly the opposite. Many of the bucklescript users are not experienced.

trefis:
@rgrinberg: I'm happy to discuss dropping support for older versions of ocaml, or support for *building* merlin with other versions of ocaml if you want to open a dedicated issue. I have some thoughts about this / the future.
As for now, we still want to do things that way for merlin itself, and while it's a bit more annoying for merlin-lsp because of the ppx dep, it seems reasonable to do it too.
So I'd suggest to not discuss this point further in this PR (if you agree).

rgrinberg:
Yes, I’m fine with it. If it addresses some short term for the users, we might as well let this in.

I will open an appropriate issue as I feel that this is an important topic.

Risto-Stevcev:
@bobzhang Is there going to be a 4.06.1 release of bucklescript as an opam switch? this might help because I think the current version is based off of that. I don't know if this pings him or if I need to open a ticket in bucklescript

src/lsp/protocol.ml
```diff
@@ -137,21 +1227,96 @@ module DocumentHighlight = struct
...
 
   let kind_of_yojson = function
-    | `Int 1 -> Ok Text
-    | `Int 2 -> Ok Read
-    | `Int 3 -> Ok Write
-    | _ -> Error "expected int between 1 and 3"
+    | `Int 1 -> Text
+    | `Int 2 -> Read
+    | `Int 3 -> Write
+    | node -> yojson_error "kind expected to be int int between 1 and 3" node
```
andreypopp:
```suggestion
    | node -> yojson_error "kind expected to be an int between 1 and 3" node
```

src/lsp/protocol.ml
```diff
@@ -344,16 +2064,16 @@ module Completion = struct
...
 
   let completionTriggerKind_of_yojson = function
-    | `Int 1 -> Ok Invoked
-    | `Int 2 -> Ok TriggerCharacter
-    | `Int 3 -> Ok TriggerForIncompleteCompletions
-    | _ -> Error "invalid completion.triggerKind"
+    | `Int 1 -> Invoked
+    | `Int 2 -> TriggerCharacter
+    | `Int 3 -> TriggerForIncompleteCompletions
+    | v -> yojson_error "invalid completion.triggerKind, show be 1-3" v
```
andreypopp:
```suggestion
    | v -> yojson_error "invalid completion.triggerKind, should be equal to 1, 2 or 3" v
```

src/lsp/protocol.ml
```diff
@@ -447,10 +2167,12 @@ module Completion = struct
...
     | `Int v ->
       begin match completionItemKind_of_int_opt v with
-      | Some v -> Ok v
-      | None -> Error "invalid completion.kind"
+      | Some v -> v
+      | None -> yojson_error "completion.kind expected to be between 1 and 25"
+                  (`Int v)
       end
-    | _ -> Error "invalid completion.kind: expected an integer"
+    | node -> yojson_error "completion.kind expected to be int between 1 and 25"
```
andreypopp:
```suggestion
    | node -> yojson_error "completion.kind expected to be between 1 and 25"
```

src/lsp/protocol.ml
```diff
@@ -774,20 +3998,1380 @@ module Initialize = struct
...
+                                                         -> params) 
+and result_of_yojson =
+  (let _tp_loc = "src/lsp/protocol.ml.Definition.result" in
+   fun t -> list_of_yojson DefinitionLocation.t_of_yojson t : Ppx_yojson_conv_lib.Yojson.Safe.t
+                                                                -> result) 
+
+let _ = params_of_yojson
+and _ = result_of_yojson
+let yojson_of_params =
+  (TextDocumentPositionParams.yojson_of_t : params -
```
andreypopp:
```suggestion
  [@@deriving_inline yojson]  [@@yojson.allow_extra_fields]
```

andreypopp:
Addressed feedback by @Khady and fixed few compiler warnings (mostly unattached docstrings).

I've also did a sanity check with my vscode extension and it seems to work. Going to merge it after CI passes — apart from 4.02.1 support it enables merlin-lsp not to depend on ppx at build time which is a nice thing to have.

andreypopp:
@sync this branch indeed had autocompletion not working, fixed it now.

sync:
@andreypopp and all amazing work thank you 👍

PR #1021, opened 2019-09-18
Merlin imenu cleanup

rgrinberg:
@Wilfred @taquangtrung  could you have a look.

I'm also worried about these two lines:

```
;;; enable depth and size threshold for OCaml modules with big size
(setq max-specpdl-size 10000)
```

I'm really not a fan of changing settings like that in some random emacs lisp script. Can we rewrite the code to avoid mutual recursion instead?

### DIFF
--- a/emacs/merlin-imenu.el
+++ b/emacs/merlin-imenu.el
@@ -88,7 +88,7 @@
     (when type-list (push (cons "Type" type-list) index))
     index))
 
-;; enable Merlin to use the merlin-imenu module
+;;;###autoload
 (defun merlin-use-merlin-imenu ()
   "Merlin: use the custom imenu feature from Merlin"
   (interactive)
--- a/emacs/merlin-imenu.el
+++ b/emacs/merlin-imenu.el
@@ -11,7 +11,6 @@
 ;; URL:
 
 (require 'imenu)
-(require 'tuareg)
 (require 'subr-x)
 (require 'merlin)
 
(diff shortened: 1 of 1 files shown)


taquangtrung:
> I'm also worried about these two lines:
> 
> ```
> ;;; enable depth and size threshold for OCaml modules with big size
> (setq max-lisp-eval-depth 10000)
> (setq max-specpdl-size 10000)
> ```
> 
> I'm really not a fan of changing settings like that in some random emacs lisp script. Can we rewrite the code to avoid mutual recursion instead?

I remember encountering some errors related to these two variables when retrieving the outline a few years ago. I will test again with the newest Merlin.

taquangtrung:
@rgrinberg Somehow your modification was based on the old version of `merlin-imenu` (0.2). Can you update it to the newest version?

rgrinberg:
@taquangtrung oops, I fixed it now.

taquangtrung:
I tested with this ML module: https://github.com/llvm-mirror/llvm/blob/master/bindings/ocaml/llvm/llvm.mli, and `merlin-imenu` reports this error:

```
funcall: Variable binding depth exceeds max-specpdl-size
```

The default value of this `max-specpdl-size` variable is 1300. Its documentation says that: 

```
Limit on number of Lisp variable bindings and `unwind-protect's.
If Lisp code tries to increase the total number past this amount,
an error is signaled.
```

Do you think that this is because the `let-binding` in my code introduces too many local variables?

rgrinberg:
I don't know unfortunately. Perhaps @Wilfred knows?

Wilfred:
Absolutely agreed we shouldn't be changing those two variables. I was thinking about this on the last PR but I didn't want to be difficult when it was already a big (and worthwhile 🙂) change.

The max binding error almost always means a stack overflow. It's a limit on how many let variables you can bind, so if you have a deep stack you can you it.

I'll try to find time for a PR in the next few days if that's helpful.

PR #1024, opened 2019-09-20
Emacs: fix stack overflow on large files

Wilfred:
Don't recurse on the outline list, as large files have a long list and Emacs runs out of stack.

This ensures that merlin-imenu doesn't need to change Emacs stack depth settings (see discussion in #1021).

Also rename variables to ensure they're appropriately namespaced.

### DIFF
--- a/emacs/merlin-imenu.el
+++ b/emacs/merlin-imenu.el
@@ -16,9 +16,9 @@
 (require 'merlin)
 
 ;; lists of different outline items
-(defvar-local value-list nil)
-(defvar-local type-list nil)
-(defvar-local exception-list nil)
+(defvar-local merlin-imenu--value-list nil)
+(defvar-local merlin-imenu--type-list nil)
+(defvar-local merlin-imenu--exception-list nil)
 
 (defun merlin-imenu-compute-position (line col)
   "Get location of the item."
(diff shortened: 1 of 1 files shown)


emacs/merlin-imenu.el
```diff
@@ -15,14 +15,10 @@
...
-(setq max-lisp-eval-depth 10000)
-(setq max-specpdl-size 10000)
-
 ;; lists of different outline items
-(defvar-local value-list nil)
-(defvar-local type-list nil)
-(defvar-local exception-list nil)
+(defvar-local merlin-imenu-value-list nil)
+(defvar-local merlin-imenu-type-list nil)
+(defvar-local merlin-imenu-exception-list nil)
```
rgrinberg: If we're going to rename them, shall we just make them private using the `--` convention? (I don't know, I trust you as the elisp expert here)

Wilfred: Agreed, we should do that. I'll update the PR.

taquangtrung:
@Wilfred, @rgrinberg : awesome patch. Now I can use `merlin-imenu` normally with the `llvm.mli` with the default thresholds of `max-lisp-eval-depth` and `setq max-specpdl-size`.

Btw, I love what you guys are doing. Do you want to be added as the contributors of this package? (and feel free to do so)

Wilfred:
@taquangtrung you mean listed as contributors in the header of the file? I don't mind, as it's all in git-blame anyway :)

Wilfred:
Updated to use the private variable naming convention.

rgrinberg:
Did you forget to add the `--` separators? Or am I misunderstanding something?

Wilfred:
Argh, I messed up my force push. Should be correct now.

rgrinberg:
Thank you!

Issue #1022, opened 2019-09-18
error (end-of-file) trying to parse answer:


Fourchaux:
Merlin dev - Linux x86_64

Since the introduction of: https://github.com/ocaml/merlin/pull/1018

_merlin-locate_ (or _merlin-type-enclosing_ ...) gives:
```
 merlin: error (end-of-file) trying to parse answer:
```
Known problem?

brettgilio:
I can replicate this issue.

https://github.com/hlissner/doom-emacs/issues/1793

brettgilio:
This issue is persisting. Is there any temporary work-around?

osener:
@brettgilio see #1026

rgrinberg:
@trefis I've pushed a hotfix to alleviate the suffering of the users. Feel free to tweak it as you wish

Fourchaux:
:smiley:

PR #1026, opened 2019-09-21
[emacs] Fix merlin/call command

osener:
This change fixes a regression introduced in #1018 and reported in #1022:

```
# calling binary: "/home/osener/.opam/default/bin/ocamlmerlin" with arguments: ("server" "/home/osener/.opam/default/bin/ocamlmerlin" "-protocol" "sexp" ...).
```


rgrinberg:
Ah, I didn't notice this PR. I've made a similar fix to master. Thanks.

Fourchaux:
Yesss...
Thanks @osener and @rgrinberg.

PR #1035, opened 2019-10-16
[lsp] enable backtraces in server

rgrinberg:
Whenever the server crashes, it should output the backtrace by default.


andreypopp:
Thanks!

Issue #1032, opened 2019-09-26
Uncaught Syntaxerr.Error exception


nojb:
The following fragment
```
let f: type a. t:'a ttype -> unit = fun ~t -> ()
```
produces the following stack trace:
```
merlin/call: unknown answer: "exception":"Syntaxerr.Error(_)
Raised at file \"src/ocaml/parsing/408/ast_helper.ml\", line 86, characters 8-57
Called from file \"src/ocaml/parsing/408/ast_helper.ml\", line 93, characters 12-49
Called from file \"list.ml\", line 92, characters 20-23
Called from file \"src/ocaml/parsing/408/ast_helper.ml\", line 102, characters 35-52
Called from file \"src/ocaml/parsing/408/ast_helper.ml\", line 96, characters 30-44
Called from file \"src/ocaml/preprocess/parser_raw.mly\", line 351, characters 34-76
Called from file \"src/ocaml/preprocess/parser_raw.mly\", line 2450, characters 10-50
Called from file \"src/utils/menhirLib.ml\", line 1885, characters 18-44
Called from file \"src/utils/std.ml\", line 93, characters 17-25
Called from file \"src/kernel/mreader_recover.ml\", line 210, characters 10-57
Called from file \"src/kernel/mreader_recover.ml\", line 248, characters 28-38
Called from file \"src/kernel/mreader_parser.ml\", line 130, characters 21-35
Called from file \"src/kernel/mreader_parser.ml\", line 175, characters 20-48
Called from file \"src/kernel/mreader_parser.ml\", line 190, characters 6-42
Called from file \"src/ocaml/utils/misc.ml\", line 45, characters 8-15
Re-raised at file \"src/ocaml/utils/misc.ml\", line 57, characters 10-24
Called from file \"src/kernel/mreader_parser.ml\", line 204, characters 20-56
Called from file \"src/kernel/mreader.ml\", line 49, characters 15-77
Called from file \"src/kernel/mpipeline.ml\", line 103, characters 19-62
Called from file \"camlinternalLazy.ml\", line 29, characters 17-27
Re-raised at file \"camlinternalLazy.ml\", line 36, characters 4-11
Re-raised at file \"src/kernel/mpipeline.ml\", line 15, characters 34-49
Called from file \"camlinternalLazy.ml\", line 29, characters 17-27
Re-raised at file \"camlinternalLazy.ml\", line 36, characters 4-11
Called from file \"camlinternalLazy.ml\", line 29, characters 17-27
[...]

PR #1037, opened 2019-10-17
Fix issue 1032

let-def:
This PR addresses the bug reported by @nojb in #1032

### DIFF
--- a/src/ocaml/preprocess/402/parser_raw.ml
+++ b/src/ocaml/preprocess/402/parser_raw.ml
@@ -2,7 +2,7 @@
 (* This generated code requires the following version of MenhirLib: *)
 
 let () =
-  MenhirLib.StaticVersion.require_20190626
+  MenhirLib.StaticVersion.require_20190924
 
 module MenhirBasics = struct
   
@@ -27821,7 +27821,7 @@ end
 
 # 27823 "src/ocaml/preprocess/parser_raw.ml"
 
-# 269 "/Users/def/.opam/4.02.3/lib/menhir/standard.mly"
+# 269 "<standard.mly>"
   
 
 # 27828 "src/ocaml/preprocess/parser_raw.ml"
--- a/src/ocaml/preprocess/408/parser_raw.mly
+++ b/src/ocaml/preprocess/408/parser_raw.mly
@@ -71,8 +71,8 @@ let psig_value (vd, ext) =
   (Psig_value vd, ext)
 let psig_type ((nr, ext), tys) =
   (Psig_type (nr, tys), ext)
-let psig_typesubst ((nr, ext), tys) =
-  assert (nr = Recursive); (* see [no_nonrec_flag] *)
+let psig_typesubst ((_nr, ext), tys) =
+  (*assert (nr = Recursive);*) (* see [no_nonrec_flag] *)
   (Psig_typesubst tys, ext)
 let psig_exception (te, ext) =
   (Psig_exception te, ext)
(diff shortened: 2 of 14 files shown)


src/ocaml/preprocess/402/parser_raw.ml
```diff
@@ -27821,7 +27821,7 @@ end
 
 # 27823 "src/ocaml/preprocess/parser_raw.ml"
 
-# 269 "/Users/def/.opam/4.02.3/lib/menhir/standard.mly"
+# 269 "<standard.mly>"
```
trefis: :open_mouth:

src/ocaml/preprocess/408/parser_raw.mly
```diff
@@ -487,15 +487,15 @@ let class_of_let_bindings ~loc lbs body =
       lbs.lbs_bindings
   in
     (* Our use of let_bindings(no_ext) guarantees the following: *)
-    assert (lbs.lbs_extension = None);
+    (* assert (lbs.lbs_extension = None); *)
```
trefis: If that assertion is incorrect, is it due to something specific in merlin, or is it something we should report upstream?

src/ocaml/preprocess/408/parser_raw.mly
```diff
@@ -314,10 +314,10 @@ let bigarray_set ~loc arr arg newval =
...
 
 let lapply ~loc p1 p2 =
-  if !Clflags.applicative_functors
-  then Lapply(p1, p2)
-  else raise (Syntaxerr.Error(
-                  Syntaxerr.Applicative_path (make_loc loc)))
+  if not !Clflags.applicative_functors then
+    raise_error (Syntaxerr.Error(
+      Syntaxerr.Applicative_path (make_loc loc)));
+  Lapply(p1, p2)
```
trefis: Can you avoid changing the style of the code for patches like this one?
I think keeping the diff against upstream to a minimal is really beneficial.

src/ocaml/preprocess/408/parser_raw.mly
```diff
@@ -71,8 +71,8 @@ let psig_value (vd, ext) =
   (Psig_value vd, ext)
 let psig_type ((nr, ext), tys) =
   (Psig_type (nr, tys), ext)
-let psig_typesubst ((nr, ext), tys) =
-  assert (nr = Recursive); (* see [no_nonrec_flag] *)
+let psig_typesubst ((_nr, ext), tys) =
+  (*assert (nr = Recursive);*) (* see [no_nonrec_flag] *)
```
trefis: ?

src/ocaml/preprocess/408/parser_raw.mly
let-def: It is unlikely that the parser recovery introduces an extension here, but in theory that might happen. It is ok to leave the assertion if you feel better about that.

src/ocaml/preprocess/408/parser_raw.mly
let-def: I change the style because the control flow changes. Do you prefer to duplicate the `Lapply (p1, p2)`? I think this is less readable.

src/ocaml/preprocess/408/parser_raw.mly
let-def: Same remark.

src/ocaml/preprocess/408/parser_raw.mly
trefis: Ah, I had stupidly assumed that `raise_error` wasn't return, while in fact it is.
Nevermind then.

trefis:
W.r.t. all the disabled assertions, I guess the follow up question is: what happens if we do generate such a piece of parsetree during recovery? If the typechecker chokes on it, are we catching and hiding that error?

trefis:
Ah! Thanks.

PR #1039, opened 2019-10-21
Remove names and versions from opam files

rgrinberg:
* The names are already present in the opam filenames
* Manually specified versions eventually become out of date

### DIFF
--- a/merlin-lsp.opam
+++ b/merlin-lsp.opam
@@ -1,6 +1,4 @@
 opam-version: "2.0"
-name:         "merlin-lsp"
-version:      "3.1.0"
 maintainer:   "defree@gmail.com"
 authors:      "The Merlin team"
 homepage:     "https://github.com/ocaml/merlin"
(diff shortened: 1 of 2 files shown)


rgrinberg:
If it's ok to bump the minimum version to 1.11, I can also move the opam metadata to the dune-project file.

Khady:
It would be better for ahrefs if >= 1.11 was not mandatory.

trefis:
Also, can you tell us more about what would be the advantages of moving the metadata to the dune-project file? (Or link us to a place which already discusses this)

rgrinberg:
> Also, can you tell us more about what would be the advantages of moving the metadata to the dune-project file? (Or link us to a place which already discusses this)

I can give a short list here:

* You don't have to maintain build/test instruction opam files.
* We plan to add linting for this step that will catch some errors like incorrectly specified dependencies.

By the way, this is in no way required for this PR. The changes here are far more conservative and good to merge regardless of our future plans.

PR #1040, opened 2019-10-21
[LSP] Little cleanups

rgrinberg:
@andreypopp Just some minor cleanups I've applied while reading the source. Perhaps you might like `Document_store` to a submodule of `Document`, but I've left things as is for now. In the next PR I will try to add some custom commands for merlin functionality that isn't lsp based (e.g. polarity search).


src/frontend/lsp/ocamlmerlin_lsp.ml
```diff
@@ -227,20 +227,17 @@ let on_request :
     return (store, lsp_locs)
 
   | Lsp.Rpc.Request.DocumentSymbol {textDocument = {uri;}} ->
-    let module DocumentSymbol = Lsp.Protocol.DocumentSymbol in
```
andreypopp: btw, ocaml doesn't warn about this and I had a bug some time ago with moving things around and forgetting to move an alias — back then it resulted in an incorrect (unexpected) deriving generated.

PR #1041, opened 2019-10-21
Use gsed on OSX

rgrinberg:
The sed that comes with OSX doesn't work in this case. 

This RP is required to run the tests on OSX.

### DIFF
--- a/tests/merlin-wrapper-template
+++ b/tests/merlin-wrapper-template
@@ -1,12 +1,18 @@
 #!/usr/bin/env bash
 
+sed="sed"
+uname=$(uname -s)
+if [ "$uname" == "Darwin" ]; then
+    sed="gsed"
+fi;
+
 src_dir="$(dirname ${BASH_SOURCE[0]})/.."
 cd "$src_dir"
 src_dir=$(realpath -e $(pwd) 2>/dev/null || readlink -f $(pwd))
 cd - > /dev/null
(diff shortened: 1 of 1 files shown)


trefis:
But will `$(uname -s)` work on windows? :)

Jokes aside, I have neither a mac, nor a windows machine, so I can't really review this. If you have a mac and this works, I trust you. (And if @dra27 confirms this should be all fine on his side as well, I'm happy to trust him as well)

rgrinberg:
> But will $(uname -s) work on windows? :)

Running the test suite on windows is quite ambitious :)

I indeed test this PR and it seems to be enough to make it all pass on Mac. Quite surprising actually. Some longer term improvements could include:

* Removing the `jq` dependency.
* Getting rid of the `--force` when running `$ make test`. Why is there anyway? Are the dependencies not tracked properly?

trefis:
Yes, I believe the dependency on merlin's executable itself is not tracked properly.
(And possibly some others).

trefis:
I do take your point about jq, and the fact that that's probably also going to be a problem to run the tests on windows.

dra27:
It's a (considerable) while since I last checked, but the testsuite used to work on Windows... it's just AppVeyor was never switched on for this repo.

FWIW, there's normally a gnaw-your-arm-off-in-frustration Posix-sed compatible solution for these things, but if it's only for the testsuite, I can't imagine it's worth it.

PR #1042, opened 2019-10-22
Re-organize frontend code

rgrinberg:
The main goal of this PR is to remove copying modules. It was necessary in the beginning because dune lacked features, but now it has proper support for multi dir sources.


.gitignore
```diff
@@ -7,9 +7,9 @@ dune-workspace
 *.pyc
 *.cmly
 
-ocamlmerlin
-ocamlmerlin-server
-ocamlmerlin-lsp
+/ocamlmerlin
+/ocamlmerlin-server
+/ocamlmerlin-lsp
```
trefis: ?

rgrinberg: The current `.gitignore` ignores `ocamlerlin` in all subdirectories of the repositories. The leading `/` makes the .gitignore rule only apply to the root dir.

This is needed because I introduced an `ocamlmerlin` directory which would be ignored without this change.

rgrinberg:
> But I don't understand why src/frontend/new is being moved to src/frontend/ocamlmerlin/new/new (to be clear: it's the two levels of "new" that surprise me).

That's just a mistake. Will fix.

rgrinberg:
@trefis should be ready now

trefis:
I did not notice before, but the same double nesting exists for the `old/` directory :laughing:

rgrinberg:
@trefis, oops. I fixed that too.

trefis:
Thanks @rgrinberg !

PR #1043, opened 2019-10-23
Emacs: fix merlin-occurrences

Wilfred:
`M-x merlin-occurrences` was broken in #888, because `merlin-occurrences-populate-buffer` relied on the side effect of moving point.

This meant that selecting results in the `*merlin-occurrences*` did not move point to the relevant line.

Also fix the style of line numbers so they're distinct from source code.

Screenshot 2019-10-23 at 11 13 05

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1517,8 +1517,11 @@ Empty string defaults to jumping to all these."
         (occ-buff (merlin--get-occ-buff))
         (positions
          (mapcar (lambda (pos)
-                   (merlin--point-of-pos (assoc 'start pos))
-                   (cons (cons 'marker (point-marker)) pos))
+                   (cons
+                    (cons 'marker
+                          (copy-marker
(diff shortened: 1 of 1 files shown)


emacs/merlin.el
```diff
@@ -1487,6 +1487,7 @@ Empty string defaults to jumping to all these."
 
 (defun merlin--occurrence-text (line-num marker start end source-buf)
   (concat (propertize (format "%7d:" line-num)
+                      'font-lock-face 'shadow
```
trefis: What is this about?

Wilfred: This sets the style of the line number to use the `shadow` style, which makes it darker.

Otherwise it can be a little hard to read `123:foo bar` that `123:` is a line number. With this change, `123:` is using the `shadow` style and it's visually distinct.

This is also consistent with `M-x occur` :)

rgrinberg:
LGTM

PR #1044, opened 2019-10-24
Env.check_state_consistency: look at cached cmis, don't load new ones

trefis:


rgrinberg:
Will this fix some errors of this form that I've been getting recently:

```
Fatal error: exception Sys_error("/Users/rgrinberg/github/ocaml/dune/_build/default/src/memo/.memo.objs/byte/memo__Spec.cmi: No such file or directory")
Raised by primitive operation at file "stdlib.ml", line 398, characters 28-54
Called from file "stdlib.ml" (inlined), line 406, characters 2-47
Called from file "src/ocaml/typing/408/cmi_format.ml", line 42, characters 11-31
```

trefis:
Hard to say without a longer backtrace.

It fixes one place where this exception was being emitted yes, but if you look at the first commit: the exception is being caught somewhere, and doesn't lead to a fatal error.
So either:
- you're getting it from a different place
- you're using the lsp backend and it doesn't catch the exception.

trefis:
That's the one indeed. We're in luck.

PR #1047, opened 2019-10-25
Spurious inconsistent assumptions

trefis:
Some time ago we removed the `without_cmi` thingy, which we did not need anymore with the new version of `-short-paths` (that is still waiting to be upstreamed).

However recently (i.e. 4.08), other parts of the compiler have been trying to load arbitrary `.cmi` files, that do not go through the usual "give me the best path" code, here are two I found today: `Printtyp.rewrite_double_underscore_paths`, `Printtyp.Namespace.lookup`.
There might be others.

This draft PR adds a test for the first one, I still have to add a test for the second.

As for the fix, the simplest one would be to reintroduce the `without_cmi` nonsense that the compiler is using.
Was there a reason to remove it other than "we don't need it"? I have the vague feeling there might have been a reason, but I can't remember.


trefis:
There were issues with `without_cmis` in the past, but they seem to have been fixed upstream by https://github.com/ocaml/ocaml/pull/1223 .
So I reintroduced it in merlin, which fixes the issues (as shown by the test).

This is ready for review.

trefis:
I'm reasonably confident this is not going to cause any issue.
So I'm merging now, feel free to complain when you come back from vacation.

trefis:
Note: this broke the increasing of verbosity when printing expressions types; we sometimes do need to load cmis.
Right now in these situations we end up with this:
```
Not_found
Raised at file \"map.ml\", line 135, characters 10-25
Called from file \"src/ocaml/typing/408/env.ml\", line 1115, characters 6-28
Called from file \"src/analysis/type_utils.ml\", line 176, characters 17-39
Called from file \"src/ocaml/utils/misc.ml\", line 47, characters 8-15
Re-raised at file \"src/ocaml/utils/misc.ml\", line 59, characters 10-24
Called from file \"src/ocaml/utils/misc.ml\", line 73, characters 10-14
Re-raised at file \"src/ocaml/utils/misc.ml\", line 75, characters 38-45
Called from file \"src/ocaml/typing/408/env.ml\", line 645, characters 10-105
Called from file \"src/utils/std.ml\", line 679, characters 8-12
Re-raised at file \"src/utils/std.ml\", line 681, characters 30-39
Called from file \"src/frontend/query_commands.ml\", line 342, characters 12-132
Called from file \"list.ml\", line 96, characters 20-25
Called from file \"src/utils/local_store.ml\", line 29, characters 8-12
Re-raised at file \"src/utils/local_store.ml\", line 37, characters 4-15
Called from file \"src/kernel/mocaml.ml\", line 40, characters 8-38
Re-raised at file \"src/kernel/mocaml.ml\", line 48, characters 4-15
Called from file \"src/frontend/ocamlmerlin/new/new_commands.ml\", line 65, characters 15-53
Called from file \"src/utils/std.ml\", line 679, characters 8-12
Re-raised at file \"src/utils/std.ml\", line 681, characters 30-39
Called from file \"src/ocaml/utils/misc.ml\", line 47, characters 8-15
Re-raised at file \"src/ocaml/utils/misc.ml\", line 59, characters 10-24
Called from file \"src/frontend/ocamlmerlin/new/new_merlin.ml\", line 99, characters 18-54
```

Obviously at that point it is *fine* to load cmis, unfortunately we're already under a `wrap_printing_env`.
So we should either turn cmi loading back on (which `Env.with_cmis` doesn't actually do), or push down the call to `wrap_printing_env`.

trefis:
More generally: `type_utils.ml` should be looked at in details now that we've reintroduced `without_cmi`.

PR #1051, opened 2019-11-04
Mark file-cache tests as related to merlin package

andreypopp:
While building merlin-lsp dune evals `(deps ...)` stanza and fails on
references to `%{bin:ocamlmerlin}` and `${bin:ocamlmerlin-server}` which
(by design) are not present in merlin-lsp package.

This commit marks `run-server-test` alias as being related to merlin
package so it's not considered while building merlin-lsp at all.


trefis:
Damn I’m an idiot; that makes sense.
Thanks!

Issue #1003, opened 2019-07-22
Merlin doesn't correctly resolve modules overriding stdlib ones.


ejgallego:
In OCaml 4.07/4.08, if I have a file:
```ocaml
let () =
  let s = List.foo in
  print_endline s
```
where `list.ml` is also in my project, and exports `foo`, my project will build correctly [under dune], however merlin insists to resolve the `List` module to the one in `Stdlib`, thus I get:
```
Unbound value List.foo
```

This is creating some pain with existing projects and OCaml 4.08 due to the new modules in that release.

trefis:
That's a known bug with our 4.08 backend yes.
And we're planning to resolve that "soon".

I'm a bit surprise that there is a problem on 4.07 though, that is: I believe on 4.07 the compiler behaves as merlin.

ejgallego:
Hi @trefis , thanks for looking into this. I can reproduce with 4.07.1 .

trefis:
Hm. Interesting.
Then it's not the bug I expected, and there must be dune shenanigans in the mix.

I'll investigate.

ejgallego:
Solved in merlin 3.3.3, thanks a lot!

PR #1052, opened 2019-11-08
Update creation of initial environment

trefis:
Merlin and the compiler behaved differently since 4.07 (and ocaml/ocaml#1010) (cf. #1003).
This PR fixes that, and also updates 4.08 to use the new `Load_path` module from the compiler distribution.

While I'm confident about the 4.07 fix, the 4.08 might degrade performances, I'll be monitoring that at janestreet over the next few weeks before merging this PR.


src/ocaml/utils/408/load_path.ml
```diff
@@ -0,0 +1,83 @@
...
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+
+module SMap = Misc.StringMap
+
+(* Mapping from basenames to full filenames *)
+type registry = string SMap.t ref
+
+let state = Local_store.new_bindings ()
```
let-def: The `state` value represents a list of dynamically-scoped variables.
With `srefk` create new variables, but at no point in time you redefine these variables.
See https://github.com/ocaml/merlin/blob/master/src/kernel/mocaml.ml#L30 for instance.

The original design assumed that there might be different notion of global scopes, hence the distinction between `Btype.state` and `Env.state`. 
However I think a single `typechecker_state` might be sufficient, and so this notion of binding could be removed.

I will write a PR to simplify this.

src/ocaml/typing/408/typemod.ml
```diff
@@ -189,7 +188,7 @@ let initial_env ~loc ~safe_string ~initially_opened_module
           match after with
           | [] -> None
           | units :: after ->
-              if String.Set.mem m units then
+              if StringSet.mem m units then
```
let-def: In https://github.com/ocaml/merlin/commit/828a79d6121913a61294753555526a7d1739797e (part of the 4.09 support) I propose to keep `String.Set` to stay close to the upstream typechecker style.

With this patch, `Std.String.Set` and `Std.String.Map` are the labelled versions (incompatible with the typechecker code) while `Misc.String.Set` and `Misc.String.Map` are the unlabelled ones.
It works as long as compiler modules open `Misc` module and Merlin modules open `Std` after `Misc`.

src/ocaml/utils/408/load_path.ml
trefis: Thanks, I rebased and this should now be fixed.

src/ocaml/typing/408/typemod.ml
trefis: 828a79d is a good idea, I extracted it from your PR and pushed it directly on master.
I'll rebase this PR once more (and I guess you'll have to rebase yours)

src/ocaml/utils/408/load_path.ml
trefis: Actually, the last commit (which uses `Local_store.Typechecker`) breaks the tests.
I guess there's something I missunderstood.

PR #1054, opened 2019-11-11
Get rid of hooks

let-def:
This simple PR remove hooks from all OCaml versions.
Hooks are not used by Merlin and were removed from OCaml 4.09. This simplification is not necessary but prepare the 4.09 support.


trefis:
The commit "Get rid of with_cmis" seems unrelated to the aim of this PR. Also, it breaks verbosity expansion, cf. https://github.com/ocaml/merlin/pull/1047#issuecomment-553851353 and https://github.com/ocaml/merlin/commit/adcae5ff2896d8992ec19bfa784dc1a3143b45b0.

PR #1059, opened 2019-11-20
Rework global state

let-def:
This changes the way state from the compiler is managed.
It now mimics a much simpler form of dynamic-scoping were bindings cannot be nested and there is a unique "binding point" that snapshots the whole compiler state.

If this doesn't break the cache (in correctness nor performance), it will help further changes (path management that @trefis worked on) and can be the basis of a hypothetical upstreaming of state management.

### DIFF
--- a/src/kernel/mocaml.ml
+++ b/src/kernel/mocaml.ml
@@ -1,13 +1,37 @@
 open Std
+open Local_store.Compiler
 
+(* Instance of environment cache & btype unification log  *)
+
+type typer_state = Local_store.scope
+
+let bound = srefk None
+
+let new_state () =
+  let scope = Local_store.fresh compiler_state in
+  Local_store.with_scope scope (fun () -> bound := Some scope);
+  scope
+
+let with_state state f =
+  if Local_store.is_bound compiler_state then
+    failwith "Mocaml.with_state: another instance is already in use";
+  match Local_store.with_scope state f with
+  | r -> Cmt_format.clear (); r
+  | exception exn -> Cmt_format.clear (); reraise exn
+
+let is_current_state state = match !bound with
+  | Some state' -> state == state'
+  | None -> false
 
 (* Build settings *)
 
 let setup_config config = (
+  assert Local_store.(is_bound compiler_state);
   let open Mconfig in
   let open Clflags in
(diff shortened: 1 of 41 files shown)


src/utils/local_store.ml
```diff
@@ -1,10 +1,17 @@
 open Std
 
 type ref_and_reset = F : 'a ref * (unit -> 'a) -> ref_and_reset
-type bindings = { mutable refs: ref_and_reset list }
+type bindings = { mutable refs: ref_and_reset list; is_bound: bool ref }
 
 let new_bindings () =
-  { refs = [] }
+  let is_bound = ref false in
+  { refs = [F (is_bound, (fun () -> true))]; is_bound }
```
trefis: `reset (new_bindings ())` sets `is_bound` to `true`. I don't understand why.

src/kernel/mocaml.ml
```diff
@@ -1,13 +1,37 @@
 open Std
+open Local_store.Compiler
 
+(* Instance of environment cache & btype unification log  *)
+
+type typer_state = Local_store.scope
+
+let bound = srefk None
```
trefis: I think the code would be slightly easier to read if this was named `current_state`.

src/kernel/mpipeline.ml
```diff
@@ -29,7 +29,7 @@ module Cache = struct
     in
     match List.assoc key !cache with
     | state ->
-      cache := List.remove_assoc key !cache;
+      cache := (key, state) :: List.remove_assoc key !cache;
```
trefis: The only reason I can see for doing this is "it's going to be quicker to access next time".
If that's the case, let's perhaps put a comment saying that before the line.

But also: the list seems be limited to a length of 6; is the change really worth it?

let-def: It's a quick'n'dirty implementation of a LRU (least recently used) cache: we cache the last 6 recently used instances of the typechecker.

If we didn't move the state to the head of the list, that would implement another caching policy... :man_shrugging:

let-def:
I implemented the changes you suggested.

I also added a check that we don't create new references after an instance has been created.

PR #1063, opened 2019-12-05
Merlin lsp opam file changes

rgrinberg:
* Add dependency on merlin
* Dune isn't a just build dependency

### DIFF
--- a/merlin-lsp.opam
+++ b/merlin-lsp.opam
@@ -10,8 +10,7 @@ build: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
-  "merlin"
-  "dune" {build & >= "1.8.0"}
+  "dune"
   "ocamlfind" {>= "1.5.2"}
   "yojson" {>= "1.6.0"}
   "ppx_yojson_conv_lib"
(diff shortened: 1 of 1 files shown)


andreypopp:
I don’t think a dependency on merlin is needed (merlin doesn't install any libs). Why dune isn’t just a build dependency?

rgrinberg:
Oh okay. I removed merlin as a dependency.
> Why dune isn’t just a build dependency?
it is a build dependency, in the sense that it’s only required at build time. However, the way opam inteprets build dependencies is by assuming that whenever build dependencies change, their artifacts remain constant. That’s clearly wrong and any package that builds with dune should be rebuilt when dune is updated.

andreypopp:
Thanks!

merlin-lsp.opam
```diff
@@ -10,8 +10,7 @@ build: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
-  "ocaml" {>= "4.02.1" & < "4.09"}
```
anuragsoni: was removing the ocaml dependency intentional?

rgrinberg:
It was intentional. The lsp binary itself doesn’t prescribe any version of OCaml, the restriction is inherited from merlin itself.

PR #1067, opened 2019-12-11
Move go-to typedef to merlin

rgrinberg:
The lsp frontend implements the ability to go to the type definition of
an expression. This requires some low level machinery, so it's moved to
merlin itself.

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -557,24 +557,35 @@ let find_source ~config loc =
    [find_source] doesn't like the "-o" option of the compiler. This hack handles
    Jane Street specific use case where "-o" is used to prefix a unit name by the
    name of the library which contains it. *)
-let find_source ~config loc =
-  match find_source ~config loc with
-  | Found _ as result -> result
-  | failure ->
-    let fname = loc.Location.loc_start.Lexing.pos_fname in
-    match
-      let i = String.first_double_underscore_end fname in
-      let pos = i + 1 in
-      let fname = String.sub fname ~pos ~len:(String.length fname - pos) in
-      let loc =
-        let lstart = { loc.Location.loc_start with Lexing.pos_fname = fname } in
-        { loc with Location.loc_start = lstart }
-      in
-      find_source ~config loc
-    with
+let find_source ~config loc path =
+  let result =
+    match find_source ~config loc with
     | Found _ as result -> result
-    | _ -> failure
-    | exception _ -> failure
+    | failure ->
+      let fname = loc.Location.loc_start.Lexing.pos_fname in
+      match
+        let i = String.first_double_underscore_end fname in
+        let pos = i + 1 in
+        let fname = String.sub fname ~pos ~len:(String.length fname - pos) in
+        let loc =
+          let lstart = { loc.Location.loc_start with Lexing.pos_fname = fname } in
+          { loc with Location.loc_start = lstart }
+        in
+        find_source ~config loc
+      with
+      | Found _ as result -> result
+      | _ -> failure
+      | exception _ -> failure
+  in
+  match result with
+  | Found src -> `Found (Some src, loc.Location.loc_start)
+  | Not_found f -> File.explain_not_found path f
+  | Multiple_matches lst ->
+    let matches = String.concat lst ~sep:", " in
+    `File_not_found (
+      sprintf "Several source files in your path have the same name, and \
+               merlin doesn't know which is the right one: %s"
+        matches)
 
 let recover _ =
   match Fallback.get () with
@@ -728,38 +739,25 @@ let from_longident ~config ~env ~lazy_trie ~pos nss ml_or_mli ident =
       locate ~config ~ml_or_mli ~path:tagged_path ~lazy_trie ~pos ~str_ident loc
 
 let from_path ~config ~env ~local_defs ~pos ~namespace ml_or_mli path =
+  let str_ident = Path.name path in
   if Utils.is_builtin_path path then
     `Builtin
   else
-    let str_ident = Path.name path in
     let browse = Mbrowse.of_typedtree local_defs in
     let lazy_trie =
       lazy (Typedtrie.of_browses ~local_buffer:true
               [Browse_tree.of_browse browse])
     in
     let nss_path = Namespaced_path.of_path ~namespace path in
-    let loc =
-      match Env_lookup.loc path namespace env with
-      | None -> Location.none
-      | Some loc -> loc
-    in
-    match
-      locate ~config ~ml_or_mli ~path:nss_path ~lazy_trie ~pos ~str_ident loc
-    with
-    | `Not_found _
-    | `File_not_found _ as err -> err
-    | `Found (loc, _) ->
-      match find_source ~config loc with
-      | Found src -> `Found (Some src, loc.Location.loc_start)
-      | Not_found f ->
-        let path = Path.name path in
-        File.explain_not_found path f
-      | Multiple_matches lst ->
-        let matches = String.concat lst ~sep:", " in
-        `File_not_found (
-          sprintf "Several source files in your path have the same name, and \
-                   merlin doesn't know which is the right one: %s"
-            matches)
+    match Env_lookup.loc path namespace env with
+    | None -> `Not_in_env str_ident
+    | Some loc ->
+      match
+        locate ~config ~ml_or_mli ~path:nss_path ~lazy_trie ~pos ~str_ident loc
+      with
+      | `Not_found _
+      | `File_not_found _ as err -> err
+      | `Found (loc, _) -> find_source ~config loc str_ident
 
 let from_string ~config ~env ~local_defs ~pos ?namespaces switch path =
   let browse = Mbrowse.of_typedtree local_defs in
@@ -806,18 +804,7 @@ let from_string ~config ~env ~local_defs ~pos ?namespaces switch path =
     match from_longident ~config ~pos ~env ~lazy_trie nss switch ident with
     | `File_not_found _ | `Not_found _ | `Not_in_env _ as err -> err
     | `Builtin -> `Builtin path
-    | `Found (loc, _) ->
-      match find_source ~config loc with
-      | Found src -> `Found (Some src, loc.Location.loc_start)
-      | Not_found f -> File.explain_not_found path f
-      | Multiple_matches lst ->
-        let matches = String.concat lst ~sep:", " in
-        `File_not_found (
-          sprintf "Several source files in your path have the same name, and \
-                   merlin doesn't know which is the right one: %s"
-            matches
-        )
-
+    | `Found (loc, _) -> find_source ~config loc path
 
 let get_doc ~config ~env ~local_defs ~comments ~pos =
   let browse = Mbrowse.of_typedtree local_defs in
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -705,7 +705,7 @@ let from_longident ~config ~env ~lazy_trie ~pos nss ml_or_mli ident =
     else
       locate ~config ~ml_or_mli ~path:tagged_path ~lazy_trie ~pos ~str_ident loc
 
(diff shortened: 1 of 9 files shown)


rgrinberg:
I haven't added supported in the traditional frontends, it doesn't seem like it's necessary for now.

src/frontend/query_protocol.ml
```diff
@@ -130,6 +130,7 @@ type _ t =
        | `Not_found of string * string option
        | `No_documentation
        ] t
+  | Type_definition : Msource.position -> (string option * Lexing.position) list t
```
andreypopp: Maybe it should be called `Locate_type` (as we have `Locate` which does the same for values).

src/frontend/query_commands.ml
```diff
@@ -359,6 +359,49 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+    let typer = Mpipeline.typer_result pipeline in
+    let structures = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
+    let pos = Mpipeline.get_lexing_pos pipeline pos in
+    let path = Mbrowse.enclosing pos [structures] in
+    let path =
+      let rec resolve_tlink env ty =
+        match ty.Types.desc with
+        | Tconstr (path, _, _) -> Some (env, path)
+        | Tlink ty -> resolve_tlink env ty
+        | _ -> None
```
trefis: You should be calling `Ctypes.repr` before looking at the type desc. That will clean up `Tlink` for you.

src/frontend/query_commands.ml
```diff
@@ -359,6 +359,49 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+    in
+    Std.List.filter_map path ~f:(fun (env, path) ->
+      Locate.log ~title:"debug" "found type: %s" (Path.name path);
+      let local_defs = Mtyper.get_typedtree typer in
+      match
+        Locate.from_string
+          ~config:(Mpipeline.final_config pipeline)
+          ~env ~local_defs ~pos ~namespaces:[`Type] `MLI
+          (* FIXME: instead of converting to a string, pass it directly. *)
+          (Path.name path)
```
trefis: I agree with the FIXME: that's not how one should do things.
Instead of calling `Locate.from_string`, you'll want to expose (and rename?) `Locate.from_completion_entry` which works on a `Path.t` directly.

src/analysis/locate.ml
```diff
@@ -809,6 +809,33 @@ let from_longident ~config ~env ~lazy_trie ~pos nss ml_or_mli ident =
...
 
+let from_path ~config ~local_defs ~pos ~namespace ml_or_mli path =
+  let str_ident = Path.name path in
+  let browse = Mbrowse.of_typedtree local_defs in
+  let lazy_trie =
+    lazy (Typedtrie.of_browses ~local_buffer:true
+            [Browse_tree.of_browse browse])
+  in
+  let nss_path = Namespaced_path.of_path ~namespace path in
+  let loc = Location.none in
```
rgrinberg: This is wrong but I'm not sure where to get the right location here.

src/frontend/query_commands.ml
rgrinberg: I tried fixing this but it wasn't so simple. Will likely need some tweaks (or a rewrite?) from you.

src/frontend/query_commands.ml
rgrinberg: Thanks for the tip. I pushed a commit to address this.

src/frontend/query_protocol.ml
rgrinberg: Renamed

rgrinberg:
I think I need to write an alternative frontend here after all. Otherwise, it doesn't seem like it's easy to add tests.

rgrinberg:
The frontend is now complete and I've added a simple test.

rgrinberg:
@trefis gentle ping on this.

src/frontend/query_commands.ml
```diff
@@ -359,6 +359,43 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
     List.map ~f:Mbrowse.node_loc path
 
+  | Locate_type pos ->
+    let typer = Mpipeline.typer_result pipeline in
+    let structures = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
+    let pos = Mpipeline.get_lexing_pos pipeline pos in
+    let path = Mbrowse.enclosing pos [structures] in
+    let path =
+      Std.List.filter_map path ~f:(fun (env, node) ->
+        Locate.log ~title:"debug" "inspecting node: %s" (Browse_raw.string_of_node node);
```
trefis: Nitpick: I'd change `"debug"` for `"query_commands Locate_type"` or something of the sort.

src/frontend/query_commands.ml
```diff
@@ -359,6 +359,43 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+        match node with
+        | Browse_raw.Expression {exp_type = ty; _}
+        | Pattern {pat_type = ty; _}
+        | Core_type {ctyp_type = ty; _}
+        | Value_description { val_desc = { ctyp_type = ty; _ }; _ } ->
+          begin match (Ctype.repr ty).desc with
```
trefis: :heart_eyes:

src/frontend/query_commands.ml
```diff
@@ -359,6 +359,43 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
     in
     List.map ~f:Mbrowse.node_loc path
 
+  | Locate_type pos ->
+    let typer = Mpipeline.typer_result pipeline in
+    let structures = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
+    let pos = Mpipeline.get_lexing_pos pipeline pos in
+    let path = Mbrowse.enclosing pos [structures] in
+    let path =
+      Std.List.filter_map path ~f:(fun (env, node) ->
```
trefis: This looks weird to me. IIUC you are call locate on the type of all the enclosing nodes … is that really useful / what you want?

Admittedly the API to get the deepest node around the cursor is … nonexistent (it's something that we should probably improve), but you can get inspiration from the way `Case_analysis` is handled lower down in this file.

src/frontend/query_commands.ml
```diff
@@ -359,6 +359,43 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+          ~local_defs ~pos ~namespace:`Type `MLI
+          path
+      with
+      | exception Env.Error _ -> None
+      | `Found (path, lex_position) -> Some (path, lex_position)
+      | `File_not_found _
+      | `Invalid_context
+      | `Missing_labels_namespace
+      | `Not_found _
+      | `Not_in_env _ -> None)
```
trefis: I assume you're swallowing the error because of the `filter_map` call, but that means that when you don't find any result, the user won't know why.

src/analysis/locate.ml
```diff
@@ -809,6 +809,33 @@ let from_longident ~config ~env ~lazy_trie ~pos nss ml_or_mli ident =
...
+  match
+    locate ~config ~ml_or_mli ~path:nss_path ~lazy_trie ~pos ~str_ident loc
+  with
```
trefis: I think you want to do something similar to what `from_longident` does before calling `locate`, i.e. calling `Utils.is_builtin_path`.

src/analysis/locate.ml
trefis: I think you'll want to do something similar to what `Env_lookup.in_namespaces` does: call the correct `Env.find_...` function depending on what the namespace is. (this can probably be an auxiliary function in `Env_lookup`).

src/frontend/query_commands.ml
rgrinberg: Indeed, it's a bit odd. That's how it was originally.

@andreypopp Was there a reason you looked up the location for the type of all enclosing nodes?

src/analysis/locate.ml
rgrinberg: Actually, I can just get the location directly from the node. Does that work?

This location is only used for:

```
    if not (Utils.is_ghost_loc loc) then Fallback.set loc;
```

Which I don't quite understand. What' a fallback location?

trefis: The location from the node is the location of the use, not the location of the definition.
Fallback locations are basically … checkpoints in a video game? If you die, you'll start over from here.

Or more seriously: we're trying to walk back from the use to the definition of the identifier, but we might get stuck at any point (because some cmt/cmti might be unavailable). In those cases we don't want to completely fail, we want to go back to the last place we looked at.
For instance if you have
```ocaml
include Foo

let f = something
```
And something comes from `Foo`, but its cmt is missing. Merlin will still jump you to the `include` line, to indicate that `something` comes from there.

rgrinberg: I see. Thanks for the explanation. I've pushed a commit that attempts to find the right location.

src/frontend/query_commands.ml
rgrinberg: > Admittedly the API to get the deepest node around the cursor is … nonexistent (it's something that we should probably improve), but you can get inspiration from the way Case_analysis is handled lower down in this file.

Isn't the deepest node around the cursor just:
```
List.hd (Mbrowse.enclosing cursor [structures]
```

I've looked into `Case_analysis`, but from what I see the filtering there is necessary to find a node that spans the user's selection. 

So it seems like I could just `find_map` instead of `filter_map`?

andreypopp: IIRC the intention was to make the query fallback to locating the outer node if (for some reason) it cannot do this for the one under the cursor. Now I don't think this is useful.

Though my thinking back then was that it can be useful if we'd return the range of the node we've located the type for, so we can highlight it in an editor (I think vscode can preview go-to-definition location in a popup and thus you can see both the original and located positions at the same time).

rgrinberg: Yeah, that seems like a good thing to have, but I think that the current implementation has some surprising side effects.

For example, consider this fragment from query_commands.ml (around line 629):

```ocaml
    let nodes =
      List.drop_while nodes
        ~f:(fun (_,t) ->
          let {Location. loc_start; loc_end; _} = Mbrowse.node_loc t in
                                        (* cursor on the [pos_start] below *)
          Lexing.compare_pos loc_start pos_start > 0 || Lexing.compare_pos loc_end pos_end < 0)
```

Attempting to locate the type will go to definition of list in `listLabels.mli`. That seems very confusing to me

trefis: > Isn't the deepest node around the cursor just: `List.hd (Mbrowse.enclosing cursor [structures])`

Probably :)

src/frontend/query_commands.ml
rgrinberg: Fixed

rgrinberg:
@trefis, have another look. I'm now only trying to locate the type for the deepest node.

rgrinberg:
@trefis little ping :) I'd like to get this PR in before updating merlin in lsp. Would prefer to do all the updates in batch.

src/analysis/locate.ml
```diff
@@ -705,6 +727,40 @@ let from_longident ~config ~env ~lazy_trie ~pos nss ml_or_mli ident =
...
+      | Found src -> `Found (Some src, loc.Location.loc_start)
+      | Not_found f ->
+        let path = Path.name path in
+        File.explain_not_found path f
+      | Multiple_matches lst ->
+        let matches = String.concat lst ~sep:", " in
+        `File_not_found (
+          sprintf "Several source files in your path have the same name, and \
+                   merlin doesn't know which is the right one: %s"
+            matches)
```
trefis, line 762: Looks like that is just duplicated in `from_string`.

rgrinberg:
@trefis Thanks for the review (and the rebase). I've addressed your comments in the last commit.

trefis:
Thanks!

PR #1068, opened 2019-12-11
Automatically generate dune rules for tests

trefis:
Inspired by dune's own setup.


trefis:
The error is due to me not generating a `(package merlin)` field in the rule stanza. I'll fix that tomorrow and then it should be good to merge.

tests/dune
```diff
@@ -11,3 +11,14 @@
...
+(include dune.inc)
+
+(rule
+ (targets dune.inc.gen)
+ (deps    (source_tree .))
+ (action  (with-stdout-to %{targets} (run ./gen/gen_tests.exe))))
+
+(alias
+ (name   runtest)
+ (action (diff dune.inc dune.inc.gen)))
```
rgrinberg: Ddi you try `(mode promote)` in the `rule` above and change the target to `dune.inc`? That seems a bit simpler.

trefis: I just copied what dune itself is doing.
I think there was a reason why simply generating `dune.inc` wouldn't work, but I don't remember. Perhaps @aalekseyev will know?

trefis:
Note: I ended up not generating the `(package merlin)` field, since LSP moved out.

trefis:
Here is something I should have seen coming: just reading the tests directories to generate the rule means that the order in which rules are generated is not deterministic… which means travis is unhappy.

rgrinberg:
> Here is something I should have seen coming: just reading the tests directories to generate the rule means that the order in which rules are generated is not deterministic… which means travis is unhappy.

In dune, I think we sort the dirs first.

PR #1070, opened 2019-12-15
Test fixes

rgrinberg:
* Deterministic order of tests. Sort tests before generating stanzas.
* Restore accidentally deleted .merlin file

### DIFF
--- a/tests/dune.inc
+++ b/tests/dune.inc
@@ -1,12 +1,12 @@
 
 (alias
  (name runtest)
- (deps (:t ./test-dirs/misc/external-arity.t)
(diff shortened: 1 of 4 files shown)


trefis:
Thanks!

And good catch for the broken test, I had missed it. :)

trefis:
Actually: I believe you forgot to git add the .merlin file for the broken test. So it's still broken :smile: 
Perhaps instead of committing that file, we should update the `.t` file to create it at the beginning of the test?

rgrinberg:
Good call. I updated the PR to generate the .merlin inside the test.

trefis:
Thanks again!

PR #1071, opened 2019-12-19
Add -kind support to complete-prefix

rgrinberg:
So this is why it wasn't working in my tests


rgrinberg:
This fix is trivial so I merged it

PR #1073, opened 2019-12-20
Print argument parsing errors

rgrinberg:
This should save people time debugging. It also caught some silent errors
in some tests. The commit history describes includes the details

### DIFF
--- a/tests/test-dirs/locate/includes/test.t
+++ b/tests/test-dirs/locate/includes/test.t
@@ -6,7 +6,7 @@ in the environment to fallback to:
 Test when the include is a name, this should directly redirect us to the right
 thing.
 
-  $ $MERLIN single locate -look-for-mli -position 4:17 -filename test.ml < test.ml
+  $ $MERLIN single locate -look-for mli -position 4:17 -filename test.ml < test.ml
   {
     "class": "return",
     "value": {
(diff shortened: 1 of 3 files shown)


trefis:
I don't think that's the right way to do things. Printing on stderr means that garbage will appear over your buffer if you're using vim (I'm not sure what happens in emacs)

We have a dedicated "notification" (or something like that) field in the json message we print on stdout as a result, failures should go in there.

trefis:
> We have a dedicated "notification" (or something like that) field in the json message we print on stdout as a result, failures should go in there.

Btw, I'm actually working on that particular area at the moment (though I have nothing to show for it yet), with the aim to also report errors when reading the configuration for .merlin files or when trying to reach the build system.
So I think you could put this work on hold for the next couple of weeks, I hope to push something in the new year.

rgrinberg:
Fine with me. How about we just merge the fixes to the tests?

trefis:
Yes, that sounds good :)

rgrinberg:
The PR now contains only test fixed

let-def:
Actually, printing on stderr is now fine with both vim and emacs. 
In vim it is redirected to an hidden logging buffer (displayed with `:b *merlin-log*` or `:MerlinDebugEnable`).
In Emacs it is dropped  unless logging is enabled, in which case it is redirected to the logging buffer.

In the future, it is fine to use stderr for debug/log/notifications.

PR #1074, opened 2019-12-20
Detect deprecation from attributes

rgrinberg:
LSP clients are able to display deprecated completion suggestions as ~~strike
through~~. Hence it's useful to pass this information through.

The current implementation is a bit too simplistic. We need to make sure that it
works for various completion suggestions.

### DIFF
--- a/tests/test-dirs/polarity-search/recent.t
+++ b/tests/test-dirs/polarity-search/recent.t
@@ -8,22 +8,21 @@ A few simple tests that show all the things we want to preserve or improve:
 
   $ echo "" | $MERLIN single search-by-polarity -query "-float +int64" \
   > -position 1:0 -filename test.ml | \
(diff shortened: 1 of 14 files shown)


tests/test-dirs/polarity-search/recent.t
```diff
@@ -49,49 +51,49 @@ A few simple tests that show all the things we want to preserve or improve:
   > tr '\n' ' ' | jq '.value.entries[] | del(.info) | del(.kind)' | head -n48
```
trefis: Let's maybe increase from 48 to whatever is required to have exactly the same elements as before?

rgrinberg: I removed the deprecated field from these results and also changed the `head` to a jq array slice. That should make this test less sensitive.

Issue #629, opened 2017-03-27
MerlinTypeOf fails to display in MacVim


nilsbecker:
since a while i observe that `:MerlinTypeOf` messages are displayed only for a fraction of a second or not at all in the status buffer. this happens whether or not i invoke it via a key map. 

i was thinking this is a bug in MacVim, which could be the case. however , `YankLatestType` always succeeds in showing a correct message, which suggests that this might be a bug in merlin, or a bug in MacVim that could be worked around easily by doing the display in the same way as `YankLatestType`. 

the behavior is unfortunately not always reproducible. fullscreen mode increases the chances of seeing it.

raphael-proust:
I can't reproduce on Ubuntu (default Terminal, default up-to-date vim) with

```
$ ocamlmerlin -version
The Merlin toolkit version 2.5.3, for Ocaml 4.02.3
```

and with

```
$ grep merlin .vimrc
execute "set rtp+=" . g:opamshare . "/merlin/vim"
  let l:dir = s:opam_share_dir . "/merlin/vim"
let s:opam_configuration['merlin'] = function('OpamConfMerlin')
let s:opam_packages = ["ocp-indent", "ocp-index", "merlin"]
```

nilsbecker:
yes, i also cannot reproduce in terminal vim. i only see it in gui macvim on os x. the symptom is this: focus switches to the status buffer quickly (which i see by a color change) then back to the buffer, but nothing is visible in the status buffer. 

it's likely to be some bug on the macvim side, but as i said above since YankLatestType works it might be possible to work around it easily.

let-def:
Could it be some extensions you are using which is interacting weirdly?
Can you share a listing of `:scriptnames` and `:au`.

nilsbecker:
sorry for the delay, i somehow missed the reply. 

first off: to reproduce, i need to start macvim in gui mode, then do a split with two vertically stacked windows. then from the upper window, invoke `:MerlinTypeOf`. focus will jump down to the status line and then back up, leaving an empty status line. (from the lower window it works). unfortunately, that does not reproduce the bug every time...

i have a ton of extensions and autocommands. scriptnames is 150 lines. 
probably it makes sense to try debugging only when reproducing with a minimal `.vimrc`. how would a minimal `.vimrc` look for Merlin?

let-def:
```viml
set nocompatible
syntax on
filetype plugin on 
set rtp+=~/.opam/xxx/share/merlin/vim
```

[...]

PR #1077, opened 2019-12-24
Add initial support for VIM 8.2's popupwin

xguerin:
This PR contains an initial proposal to display the result of `MerlinTypeOf` in a popup window as made available since Vim 8.2.

It is automatically enabled when the 'popupwin' feature is detected. An option to control its activation externally may be desired.

It reuses merlin's type history buffer but skips the processing in `s:RecordType` to avoid having extraneous blank lines at the beginning and at the end of the buffer.

### DIFF
--- a/vim/merlin/autoload/merlin_type.vim
+++ b/vim/merlin/autoload/merlin_type.vim
@@ -103,51 +103,71 @@ endfunction
 
 function! merlin_type#Show(type, tail_info)
 
-  let l:user_lazyredraw = &lazyredraw
-  if l:user_lazyredraw ==# 0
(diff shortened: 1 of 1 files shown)


let-def:
Sorry for not looking at your PR earlier. Did you give up on this feature or do you have other plans?

xguerin:
After seeing that comment https://github.com/ocaml/merlin/issues/629#issuecomment-568987544 I thought the change would not be welcomed :)

I switched to `lsp` in the meantime, but the change was working, so if you think it is still relevant I’d be glad to reopen it and add whatever you deem missing.

PR #1078, opened 2019-12-26
Add failing test for destruct

rgrinberg:
Invalid destruct calls aren't handled properly.

### DIFF
--- a/tests/test-dirs/destruct/basic.t
+++ b/tests/test-dirs/destruct/basic.t
@@ -132,34 +132,13 @@ pprintast).
     "notifications": []
   }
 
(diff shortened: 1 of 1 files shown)


trefis:
The backtrace is not stable on different versions of caml.
You might want to `$ export CAMLRUNPARAM=b=0` before that test.

rgrinberg:
> You might want to $ export CAMLRUNPARAM=b=0 before that test.

I don't think this works because merlin sets `Printexc.record_backtrace true`. I added some ugly hacks with grep to get rid of the stacktrace. It's a bit ugly, but once the bug is fixed, it will not be necessary anyway.

PR #1082, opened 2020-01-08
Fix short-paths for 409

lpw25:
The 4.09 short-paths code had a FIXME. This PR fixes it.

Without this PR, `-short-paths` does not have access to any information from .cmi files -- so it doesn't work very well.

### DIFF
--- a/src/ocaml/typing/409/env.ml
+++ b/src/ocaml/typing/409/env.ml
@@ -631,44 +631,11 @@ type persistent_module = {
 
 (* Short paths basis *)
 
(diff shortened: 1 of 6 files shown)


trefis:
I haven't thought about it at all, but: do you know if there are some regression tests we could add (and what they'd look like) so that doesn't happen again in the future?

trefis:
More precisely: we supposedly already have some regression tests for `-short-paths`, but they didn't catch this, so...

lpw25:
I've added a simple test

trefis:
Thanks!

PR #1085, opened 2020-01-27
Include the stdlib in the source path

lpw25:
Currently locate won't jump to locations in the standard library because it is not part of the source path.

Arguably the stdlib directory is a build directory not a source directory, but since it includes the `.mli` files alongside the `.cmti` files it makes sense to treat it as if it were a source directory.


lpw25:
Note that it actually will jump to locations in the stdlib if there are no other source files with the same name, because the algorithm falls back to trying the path of the `cmti` file. However, if there is another source file with the same name then the fallback path is not tried and merlin just jumps to the wrong file.

PR #1086, opened 2020-01-27
Always consider the cmti's path when looking for its source file

lpw25:
This is an alternative fix to the issue described in #1085. This PR makes sure that the directory containing the `cmti` file is always included in searches for its corresponding source file. Previously it was only used as a fallback when there were no results from the source path -- this could cause the wrong file to get selected. Now it is always considered, which means we use the more accurate digest test if there are other files with the same name in the source path.

I think this PR should be included regardless of whether #1085 is merged as the current behaviour is fragile. #1085 should maybe also be merged depending on what else the source path is used for in merlin.

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -274,11 +274,11 @@ module Utils = struct
     in
     List.fold_left ~f:try_dir ~init:[] path
 
-  let find_all_matches ~config ?(with_fallback=false) file =
+  let find_all_matches ~config ~dir ?(with_fallback=false) file =
     let files =
       List.concat_map ~f:(fun synonym_pair ->
         find_all_in_path_uncap ~src_suffix_pair:synonym_pair ~with_fallback
-          (Mconfig.source_path config) file
+          (dir :: (Mconfig.source_path config)) file
       ) Mconfig.(config.merlin.suffixes)
     in
     List.uniq files ~cmp:String.compare
@@ -581,15 +581,12 @@ let find_source ~config loc =
     | "" -> dir
     | cwd -> Misc.canonicalize_filename ~cwd dir
   in
-  match Utils.find_all_matches ~config ~with_fallback file with
-  | [] ->
-    log ~title:"find_source" "failed to find %S in source path (fallback = %b)"
-       filename with_fallback ;
-    log ~title:"find_source" "looking for %S in %S" (File.name file) dir ;
-    begin match Utils.find_file_with_path ~config ~with_fallback file [dir] with
-    | Some source -> Found source
-    | None ->
-      log ~title:"find_source" "Trying to find %S in %S directly" fname dir;
+  match Utils.find_all_matches ~dir ~config ~with_fallback file with
+  | [] -> begin
+      log ~title:"find_source"
+        "failed to find %S in source path or %S (fallback = %b)"
+        filename dir with_fallback ;
+      log ~title:"find_source" "Trying to find %S in %S directly" fname dir ;
       try Found (Misc.find_in_path [dir] fname)
       with _ -> Not_found file
     end


trefis:
Disclaimer: I haven't looked at this / thought about it much yet.

This is indeed a fragile area, that required a lot of tuning in the past, and I'm a priori reluctant to change it. The change you propose could have the following effect for instance: assume the build system copies source files from the src dir to its build dir, and builds there. Then merlin might always look at the .ml next to the .cmt, which is not the file you want.

I remember having issues like that in the past, where I would jump somewhere, edit the file, rebuid: and discover that my changes have been lost because I edited the build dir and the build system copied the source file over my changes.

Could you give an example where the current heuristics have proven problematic?

lpw25:
> Could you give an example where the current heuristics have proven problematic?

Create a project with an `array.mli` and try to jump to the `Array.make` from the stdlib. You should find yourself on line 41 of your `array.mli` rather than the one in the stdlib install directory. 

Essentially merlin should prefer to go to an `.mli` next to the `.cmti` file with the correct digest rather than an `.mli` file from the source path with the wrong digest.

I think that between multiple paths with the correct digest it would be better to go to the one from the source path -- that would handle the example you gave properly -- and this PR does not currently implement that.

lpw25:
From Jane Street's perspective #1085 is sufficient since the stdlib is the only place where `.cmti` files are next to `.mli` files. So if you want to merge that one for now and consider the other cases later that seems fine to me.

lpw25:
Having thought about this some more, I think that the underlying problem that your example illustrates is that there are really 3 kinds of directory that the merlin will have access to:

1. Source directories -- these contain the sources of the files which are currently being worked on (e.g. the `src` directory of the current project)

2. Build directories -- these contain the build artifacts for the files which are currently being worked 
on (e.g. `dune`s build directory)

3. Install directories -- these contain the build artifacts and some source files for libraries which have been installed (e.g. `lib` directories in an opam switch)

You want to read cmti files from 2 and 3. You want to locate source files from 1 and 3 but not 2.

My understanding is that currently tools which generate `.merlin` files mark 1 as source directories and mark both 2 and 3 as build directories. This clearly prevents treating 2 and 3 differently.

I see two reasonable solutions:
a) Add an "install directory" option in `.merlin` files allowing build systems to explicitly mark directories as 3. Then remove the code that will try to look in the directory of the cmti file for source files. You could also restrict the search in various ways -- you could require that cmti's in build directories have sources in a source directory, whilst cmti's in install directories should have source files in the same directory.

b) Specify that build systems should mark 3 directories as both build and source directories. Then remove the code that will try to look in the directory of the cmti file for source files. 

(b) requires the least change in merlin, but it may double the size of `.merlin` files (I don't know if that is a concern or not).

Whichever path is taken it is clear that the stdlib directory should be treated as a 3.

trefis:
> Having thought about this some more, I think that the underlying problem that your example illustrates is that there are really 3 kinds of directory that the merlin will have access to:
> 1. Source directories -- these contain the sources of the files which are currently being worked on (e.g. the src directory of the current project)
> 2. Build directories -- these contain the build artifacts for the files which are currently being worked
    on (e.g. dunes build directory)
> 3. Install directories -- these contain the build artifacts and some source files for libraries which have been installed (e.g. lib directories in an opam switch)
>
> You want to read cmti files from 2 and 3. You want to locate source files from 1 and 3 but not 2.

That sounds right, and actually we already have the notion of "package path" for 3. In the .merlin language 1 maps to `S`, 2 maps to `B` (or a pair of `CMI` and `CMT` if one needs to be precise) but 3 is not exposed properly, instead we have `PKG`, which takes a findlib library name.

Adding some syntax for the notion of package path (your option `a`) would be reasonable, but in my work to make .merlin files disappear I've actually removed that notion of package paths from merlin. In that world, asking the build system to mark 3 as both source and build paths seems reasonable (since nothing will be stored on disk).

As for the case of the stdlib, I'll just merge #1085.

trefis:
tl;dr for @rgrinberg and @voodoos: for *install directories* (directories which contain the build artifacts and some source files for libraries which have been installed (e.g. `lib` directories in an opam switch)) dune currently only generates `B` directives in `.merlin` files.
It should be updated to generate `S` directives as well.

rgrinberg:
@voodoos would you like to fix this on the dune side?

voodoos:
Using this simple dune file:

```dune
(executable
 (name main)
 (libraries yojson))
```

Dune appears to be already generating the correct `.merlin` file:
```
B ....../.opam/default/lib/biniou
B ....../.opam/default/lib/easy-format
B ....../.opam/default/lib/yojson
B _build/default/.main.eobjs/byte
S ....../.opam/default/lib/biniou
S ....../.opam/default/lib/easy-format
S ....../.opam/default/lib/yojson
S .
FLG -w @1..3@5..28@30..39@43@46..47@49..57@61..62-40 -strict-sequence -strict-formats -short-paths -keep-locs
```

So I guess there is nothing more to do ?

trefis:
Yep, seems I spoke to fast and dune is already doing the right thing.
Having fixed the stdlib handling on merlin's side, there's indeed nothing left to do.

lpw25:
I think that this part:

> Then remove the code that will try to look in the directory of the cmti file for source files.

still remains to be done. I think this code is fragile and will mask cases where folders are missing from the source directories list.

A concrete example is merlin's own handling of the stdlib: without this code it would have been obviously wrong from the beginning, with this code it was a confusing bug that only arose when a project had a filename that matched one in the stdlib.

PR #1089, opened 2020-01-28
Improve completion tests

rgrinberg:
Add tests for the application context


rgrinberg:
I guess I can disable the tests for older versions of OCaml to fix the CI.

PR #1091, opened 2020-02-04
Fix wrong type information for constructors under type coercions

voodoos:
## Bug description:

Given the following file:

```ocaml
type t = U
type t' = U

let f : t  = U
```

The type hint given by Merlin for the `U` occuring in `f` is `t'` instead of `t`.
The same problem happens in patterns.

## Solution:

- Refactor the `Browse_raw.fold_node` so that `exp_extra` nodes are considered children of their corresponding expression and not parents.
- Short the heuristic for type enclosings when the type can be directly extracted from the node.

### DIFF
--- a/src/frontend/query_commands.ml
+++ b/src/frontend/query_commands.ml
@@ -284,6 +284,27 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
       let exprs = reconstruct_identifier pipeline pos expro in
       let env, node = Mbrowse.leaf_node (Mtyper.node_at typer pos) in
       let open Browse_raw in
+      let ident_opt =
+        let longident_to_string id = try
+          String.concat ~sep:"." (Longident.flatten id)
+          with Misc.Fatal_error _ -> ""
+        in
+        let ret typ = Mbrowse.node_loc node, `Type (env, typ), `No in
+        match node with
+        | Expression e ->
+          (match e.exp_desc with
+          | Texp_construct ({ Location. txt; loc=_ }, cdesc, _) ->
+            Some(longident_to_string txt, ret cdesc.cstr_res)
+          | Texp_ident (_, { Location. txt; loc=_ }, vdes) ->
+            Some(longident_to_string txt, ret vdes.val_type)
+          | _ -> None)
+        | Pattern p ->
+          (match p.pat_desc with
+          | Tpat_construct ({ Location. txt; loc=_ }, cdesc, _) ->
+            Some(longident_to_string txt, ret cdesc.cstr_res)
+          | _ -> None)
+        | _ -> None
+      in
       let include_lident = match node with
         | Pattern _ -> false
         | _ -> true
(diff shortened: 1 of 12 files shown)


src/frontend/query_commands.ml
```diff
@@ -281,40 +281,67 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+           Moreover the `of_exprs` heuristic  would fail for coerced constructors *)
+        let ret typ = Some (Mbrowse.node_loc node, `Type (env, typ), `No) in
+        match node with
+        | Expression e ->
+          (match e.exp_desc with
+          | Texp_construct (_, cdesc, _) ->
+            (*Printf.eprintf "toto";
+            Printtyp.wrap_printing_env env ~verbosity (fun () ->
+              Printtyp.type_expr (Format.err_formatter) cdesc.cstr_res);
+            Format.pp_print_flush Format.err_formatter ();*)
```
trefis: :thinking:

tests/test-dirs/type-enclosing/cons.t
```diff
@@ -0,0 +1,62 @@
...
+        "col": 13
+      },
+      "end": {
+        "line": 4,
+        "col": 14
+      },
+      "type": "t",
+      "tail": "no"
+    }
+  ]
```
trefis: Perhaps you should add a comment explaining why the output is duplicated, and that we aim to fix that in the future.

src/frontend/query_commands.ml
```diff
@@ -281,9 +281,30 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
     let result = List.filter_map ~f:aux path in
     (* enclosings of cursor in given expression *)
     let small_enclosings =
-      let exprs = reconstruct_identifier pipeline pos expro in
-      let env, node = Mbrowse.leaf_node (Mtyper.node_at typer pos) in
       let open Browse_raw in
+      let env, node = Mbrowse.leaf_node (Mtyper.node_at typer pos) in
+      let exprs = (reconstruct_identifier pipeline pos expro) in
```
trefis: This didn't need to change :)

voodoos:
If switching back to the original browse tree the wrong type is returned:

```diff
Done: 1608/1621 (jobs: 4)File "tests/test-dirs/type-enclosing/cons.t", line 1, characters 0-0:
         git (internal) (exit 1)
(cd _build/default && /usr/bin/git diff --no-index --color=always -u tests/test-dirs/type-enclosing/cons.t tests/test-dirs/type-enclosing/cons.t.corrected)
diff --git a/tests/test-dirs/type-enclosing/cons.t b/tests/test-dirs/type-enclosing/cons.t.corrected
index 2eb1f699..e04a6da2 100644
--- a/tests/test-dirs/type-enclosing/cons.t
+++ b/tests/test-dirs/type-enclosing/cons.t.corrected
@@ -13,7 +13,7 @@ Various parts of the cons.ml:
         "line": 4,
         "col": 14
       },
-      "type": "t",
+      "type": "t'",
       "tail": "no"
     },
     {
@@ -108,7 +108,7 @@ We aim to fix that in the future.
         "line": 15,
         "col": 15
       },
-      "type": "M.t",
+      "type": "M.u",
       "tail": "no"
     },
     {
```

PR #1096, opened 2020-02-13
Disable some tests for < 4.09

rgrinberg:
The application_context.t and cons.t tests give slightly different
results on older versions of OCaml. Therefore we disable them.

This should fix ci


trefis:
I'm not fond of that solution. I'll try to have a look to see why they differ and whether that can be fixed.

Issue #714, opened 2017-08-30
Merlin process opens short-lived cmd windows on Windows Emacs


leviroth:
To reproduce:

1. Install Ocaml on Windows 10 via https://fdopen.github.io/opam-repository-mingw/. I'm using version 4.04.0+mingw64.
   
2. Install merlin via opam. I am currently pinned to 8ea5c48d, but also saw this issue on 2.5.5.

2. Install Emacs via https://sourceforge.net/projects/emacsbinw64/. (I didn't test with other builds.) Version: GNU Emacs 25.1.1 (x86_64-w64-mingw32) of 2016-09-21

3. Add merlin to .emacs by copying the code at https://github.com/ocaml/merlin/wiki/emacs-from-scratch#installing-merlin-and-its-emacs-mode-with-opam.
   
3. Open Emacs via [ocaml-env-win](https://fdopen.github.io/opam-repository-mingw/ocaml-env/).
   
4. Open an Ocaml buffer and save. A cmd window will open very briefly any time merlin is invoked.
   
In contrast, ocp-indent works without a hitch, so this seems to be a problem with Merlin specifically.

The actual checking behavior works as expected, which is great. The cmd window a) is a visual annoyance and b) captures keystrokes, since it pops up during completion.

leviroth:
After investigating, it looks like this is caused by the server invocation [here](https://github.com/ocaml/merlin/blob/d301f4971ae8cdeb7345db585283d3cfaa58aae2/emacs/merlin.el#L487). If I change `"server"` to `"single"`, the problem goes away. Is it possible that the server is opening a new command window and then exiting immediately?

let-def:
I don't know why windows would open a command window (wrong compilation flags?!), @dra27 can maybe help? :)

bryphe:
@jordwalke and I were spending time debugging this - I used [Process Monitor](https://docs.microsoft.com/en-us/sysinternals/downloads/procmon) to grab a trace.

I found a few clues:
- This only reproduces when I am using `ppx`. For example, the [`hello-reason`](https://github.com/esy-ocaml/hello-reason) project does not exhibit this behavior, but the [`merlin-language-server`](https://github.com/bryphe/merlin-language-server) project does. (`hello-reason` uses no ppx, while `merlin-language-server` does).
- With the process monitor trace, I see calls to `ppx.exe` spawned using `cmd.exe /c`. The problematic process creations occurred in this form:

```
[...]

PR #1101, opened 2020-03-03
Fix windows popup 

jordwalke:
I pieced together many of the different suggestions discussed in [this issue](https://github.com/ocaml/merlin/issues/714).

We need to use the system shell to correctly interpret ppxs, however, when doing so, it makes windows pop up an annoying cmd window on every keystroke. @bryphe's suggestion was to wrap it in a `create_process("cmd.exe", [| theCommand |])`- that got rid of the popup, but it had one problem which we couldn't explain:  The same exact command would not work. I dug into it and it looks like any time you send a command to `cmd.exe` through `create_process`, some quotes get escaped unnecessarily. It doesn't happen with `Sys.command`. I almost gave up hope but at the last minute I thought of a hack: Pass the information through environment variables which will bypass the escaping!

### DIFF
--- a/src/ocaml/driver/pparse.ml
+++ b/src/ocaml/driver/pparse.ml
@@ -51,7 +51,7 @@ end
 module WinCmd : PpxCmd = struct
   type command_input = string * string list
   let normalize_first_segment_as_windows_path s =
-    let separated = String.split_on_char ' ' s in
+    let separated = String.split_on_char_ ' ' s in
     let rec first_path cur_str rest =
       match (cur_str, rest) with
       | ("", []) -> ("", [])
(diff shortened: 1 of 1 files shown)


Chimrod:
I'm just adding a comment to tell you that your solution is working for me. I'm using your patch since about a month without any problem.

I hope this will be merge in the main branch !

Chimrod:
Any news on the patch integration ? The branch requires now some minor modification to work against tag 3.4.2, but the the correction still works and correct a major annoyance (when we use Merlin with autocompletion, the pop ups raise each time we press a key).

MisterDA:
I've tested the PR on Emacs 28+Windows 10, rebased on the 411 branch (currently 4c8f9f8c8b9b9448d4212c262197bee453ee26f9). This PR fixes the problem. Emacs is almost unusable without it when editing a file with ppx. I suggest squashing the PR and hot-fixing it ;)
I think I notice a bit of added latency when typing just after autocompletion with this patch.

voodoos:
Let's close this now that  #1270 has been merged.

PR #1104, opened 2020-03-04
Type-enclosing gives a module when it should give a module type

lpw25:
Given an `.mli` with:

```ocaml
module type T = sig type a end

module T : sig type b end

include T
```

type-enclosing on the last `T` will return the module type `sig type b end` when it should give `sig type a end`.

This PR adds a test demonstrating this issue.


trefis:
Cc @voodoos, this should be fixed by your work on context detection.

PR #1106, opened 2020-03-04
Type-enclosing gives wrong type on shadowing let

lpw25:
Given:
```ocaml
let def = 6

let def : float = float_of_int def
```
Merlin will say `int` when asked for the type on the second `def`.

This PR adds a test demonstrating the issue.


trefis:
Merged manually, thanks.

PR #1107, opened 2020-03-04
Complete-prefix ignoring some infix operators

lpw25:
Given:
```ocaml
module Z = struct
  let (>>) = 0
  let (|+) = 0
  let (|-) = 0
  let (>>=) = 0
  let (>>|) = 0
end

let _ = Z.
```
complete-prefix offers all the operators as candidates except `(|-)`.

This PR adds this example as a test case.


trefis:
Fixed by #1142 .

PR #1108, opened 2020-03-04
Refactor `Type_enclosing` and make use of `Context` information

### LINKED ISSUE #864: merlin-type-enclosing picks up variable from scope instead of record field
In all of the examples below I run emacs, place point on "z" after the dot, and invoke (merlin-type-enclosing).
In all cases the expected behavior is to highlight the entire rhs and say that the type is int.
Currently example `f3` breaks.

```ocaml
module X = struct
  type t = { z : int }
end

let f1 _            (x : X.t) = x.z   (* highlights [x.z], reports type [int] *)
let f2 (z : string) (x : X.t) = x.X.z (* highlights [x.X.z], reports type [int] *)
let f3 (z : string) (x : X.t) = x.z   (* highlights just [z], reports type [string] *)
```

voodoos:
This PR:
- Extracts the module `Context` from `analysis/locate.ml`
- Extracts some of the `Type_enclosing` logic from `query_commands.ml`
- Makes `Context` provide richer infos
- Use that information in `Type_enclosing`

This fixes the bug highlighted in #1104 
 Should also fix #864


trefis:
Actually, can we split this PR in two?
Let's just keep the extraction commits (the first two) that are obviously good and mergeable, and open a separate one for the use of context in type_enclosing.

Thanks!

voodoos:
## Status of this PR
It is still partial, and when context information is not satisfying it fallback to the old behaviour: try to type in the various namespaces until success. Tests need to be more exhaustive but I believe it should give identical or better results than master.

## The weird transition from 405 to 406
As we discussed with @trefis and @let-def, something happened between these two versions.
With this PR the tests [`context.t`](https://github.com/ocaml/merlin/blob/234c896ac483f32d7ad6c391b60063d3897a1a4a/tests/test-dirs/type-enclosing/context_correct_pre_406.t) and [`let.t`](https://github.com/ocaml/merlin/blob/234c896ac483f32d7ad6c391b60063d3897a1a4a/tests/test-dirs/type-enclosing/let_correct_pre_406.t), which were added recently as issues, are indeed behaving as expected for ocaml `<= 405`. This is an improvement of this PR.
However past version 406 they do not behave as expected any more.

## Todo
- [ ] Solve the discrepancy between 405 and 406 (may be in another PR)
- [ ] Use context information to type partial paths (for now it falls back to previous behaviour)

voodoos:
## What changed since last comment:
- Added more test
- Cleaner `Type_enclosing` signature
- Removed some need for the fallbacks thanks to better context longident analysis
- Test `cons.t` is now correct from version 406
- Record labels are now correctly typed
- Fixed a bug in locate (527f4a7)
- [ ] Tests `let.t` still wrong for all versions along `cons.t` for version prior to 406

voodoos:
## Summary
- Fixes #1104  and #864
- Fixes #1109 but post-405 results are not as good as before
- Fixes undeclared issues such as record labels typing
- Does not fix the issue with GADT (solution similar to #1091 and #1121)
- Does not fix issue #1124 with objects' methods (the context is not yet aware of objects 🤯)
- The typing function still fall backs sometime to the legacy behaviour. Necessary due #1128

src/analysis/context.mli
```diff
@@ -41,6 +41,21 @@ type t =
...
 val to_string : t -> string
 
+(**
+  [inspect_browse_tree lid ~cursor mbrowse] tries to provide contextual
+  information given the selected identifier, the position of the cursor and the
+  typed tree. It is used by Locate and Type_enclosing.
+
+  [Mbrowse.enclosing cursor mbrowse] is used to obtain the node enclosing the
+  cursor's position un the typed tree. This node is then filtered to deduce
+  information on the enlosing context.
```
trefis, line 51: Is this information really necessary to use the function?

src/analysis/type_enclosing.ml
```diff
@@ -23,41 +32,18 @@ let from_nodes path =
...
-          | _ -> None)
-        | Pattern p ->
-          (match p.pat_desc with
-          | Tpat_construct ({ Location. txt; loc=_ }, cdesc, _) ->
-            Some(longident_to_string txt, ret cdesc.cstr_res)
-          | _ -> None)
-        | _ -> None
-
-let from_reconstructed verbosity exprs env node =
+let from_reconstructed ~get_context ~verbosity env node exprs =
```
trefis, line 40: Why is `get_context` passed as a parameter? My first impression is that it makes the code harder to reason about; `from_reconstructed` is called only once, so `get_context` is always the same why not simply inline it?

voodoos:
@trefis do you think this is ready for merging ?

It opens the door to more work, but this can be the subject of other PR's:
- [ ] Solve the discrepancy between 405 and 406 (may be in another PR)
- [ ] `cons.t` test still wrong from 406. 
- [ ] when cursor at (M.|A 3), the enclosing node returned is const 3, thus
  breaking the context inference (but the fallback catch that for now...)

These are not regressions but already present issues. 
The last one is the one that prevent us for having a fully context-based type inference. (For now the old fallback behaviour is still in place.)

PR #1117, opened 2020-03-27
Support for OCaml 4.10

trefis:
- `-short-paths` is working
- the only output in `tests/` is present both with and without `-short-paths` and looks like an improvement to me.

@let-def can you skim over this and make sure there's no "FIXME" or anything of the sort left?


let-def:
Looking at it!

let-def:
I updated opam constraints to reenable CI.

trefis:
Some tests are broken on "old" versions of OCaml, but these tests were added recently, and are already broken on master.
So I believe this PR is good to squash & merge :)

let-def:
Wunderbar! :fireworks:

PR #1120, opened 2020-04-03
Stricter typechecker caching policy

let-def:
Instances of the typechecker are cached based on configuration (values of type `Mconfig.t`).

Older versions of Merlin ignored some components resulting in possible mismatches between the internal configuration of the typechecker (loadpath, global modules visible from the environment) and Merlin configuration.

This patch just use the whole configuration as caching key.

### DIFF
--- a/src/kernel/mpipeline.ml
+++ b/src/kernel/mpipeline.ml
@@ -24,9 +24,7 @@ module Cache = struct
 
   let get config =
     let title = "pop_cache" in
-    let key =
-      Mconfig.(config.query.directory, config.query.filename, config.ocaml)
-    in
+    let key = config in
     match List.assoc key !cache with
     | state ->
       cache := (key, state) :: List.remove_assoc key !cache;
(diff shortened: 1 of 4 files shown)


let-def:
Using too much configuration is not satisfying either (https://github.com/ocaml/merlin/blob/master/src/kernel/mconfig.mli).
For instance `verbosity` is now part of the cache, which is certainly wrong :P.
I will establish a finer grained list of fields to cache.

let-def:
The values used as a key for the cache should:
- allow to maximize reuse; associating a single typechecker instance to a filename and directory is natural, but keying also based on verbosity makes no sense
- prevent reuse in different environments (if there is a change in loadpath, a new typechecker should be produced).

It would be better to guarantee that the typechecker was well-behaved when the loadpath changes (so that we can reusing the same instance, and let the typechecker figure which part of its internal state should be invalidated).
However we already had many bug related to that.  There are subtle changes in the type checker behavior accross the different versions of OCaml.
It is simpler to create new instances upfront.

PR #1123, opened 2020-04-06
External config readers

voodoos:
Twin PR: https://github.com/ocaml/dune/pull/3395

This PR is based on @trefis [external-config](https://github.com/ocaml/merlin/tree/external-config) branch.

This branch adds an external reader for config files to merlin. Merlin communicates with this reader to get configuration information.

This PR implements a simple communication protocol between a config provider and merlin via standard inputs and outputs.

Processed config files are encoded as S-expressions, for exemple:
```
(("S" "/.../..../") ("B" "/.../.../") ...)
```

This S-expression is then sent as a _canonical s-expression_ by the config provider to merlin.

This choice of protocol should allow an easy communication with a future "dune/merlin communicator".

## Todo
- [x] Improve error handling


src/frontend/ocamlmerlin/old/old_command.ml
```diff
@@ -99,9 +96,14 @@ let configure (state : buffer) =
...
-      ~filenames:(Option.cons (Option.map ~f:Filename.dirname state.path)
-                    (Option.value ~default:[] state.dot_merlins))
+  let config =
+    match state.dot_merlins with
+    | Some (first :: _) -> (* ignore anything but the first one... *)
+      Mconfig.get_external_config first config
+    | None | Some [] ->
+      match state.path with
+      | None -> config
+      | Some p -> Mconfig.get_external_config p config
```
voodoos, line 106: @trefis 

I had to change that part (`Mconfig.gec (Filename.dirname p) config` -> `Mconfig.gec p config`) for things to work with the old protocol.

However I am not sure of what is happening here: If `state.path` contains the path to the file, while would we prefer the path to an arbitrary chosen `.merlin` file ?

The new config reader expect the path to a source file, not a dot merlin file. Especially the Dune one which will be able to give file-specific configs.

src/frontend/ocamlmerlin/old/old_command.ml
```diff
@@ -197,9 +199,13 @@ let dispatch_sync config state (type a) : a sync_command -> a = function
       Mconfig.(config.merlin.flags_to_apply)
 
   | Project_get ->
+    (*
     let pipeline = make_pipeline config state in
     let config = Mpipeline.final_config pipeline in
     (Mconfig.(config.merlin.dotmerlin_loaded), `Ok) (*TODO*)
+    *)
+    (* FIXME: ??? *)
+    ([], `Ok)
```
voodoos, line 208: @trefis what should be happening here ?

src/frontend/query_commands.ml
```diff
@@ -618,6 +618,15 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
          (if parsing then parser_errors else []) @
          (if typing then typer_errors else []))
     in
+    (* Add configuration errors *)
+    let errors =
+      let cfg = Mpipeline.final_config pipeline in
+      let failures =
+        (* FIXME: the source is known *)
+        List.map ~f:(Location.error ~source:Unknown) cfg.merlin.failures
```
voodoos, line 626: @voodoos TODO

src/frontend/ocamlmerlin/new/new_commands.ml
```diff
@@ -564,8 +564,11 @@ The return value has the shape:
     begin fun pipeline () ->
       let config = Mpipeline.final_config pipeline in
       `Assoc [
+        (* TODO: print the configuration context. *)
```
voodoos, line 567: @voodoos TODO
@trefis  Maybe you can clarify what is expected here ? Just printing the .merlin content ?

src/kernel/mconfig_dot.ml
```diff
@@ -220,213 +74,154 @@ let parse_suffix str =
...
-    try Some(resolve_path (Findlib.package_property predicates pkg "ppx"))
-    with Not_found -> None
-  and ppxopts =
+type context = string * Configurator.t
+
+let get_config (dir, cfg) path =
+  try
+    let p = Configurator.get_process ~dir cfg in
+    let cmd = Dot_protocol.Commands.make_f path in
+    (* TODO: ensure [path] is absolute, or that it is relative to dir, and not the cwd. *)
```
voodoos, line 191: @trefis The only call to that function is in [Mconfig](https://github.com/ocaml/merlin/pull/1123/files#diff-4fa8910acf164e080175042c6ad98d87R212) and is preceded by a call to `canonicalize_filename`. Thus it should always be absolute, shouldn't it ?

src/kernel/mconfig_dot.mli
```diff
@@ -28,34 +28,29 @@
...
-   string list (* failures *))
+val find_project_context : string -> context option
+(** [find_project_config dir] searches for a "project configuration file" in dir
+    and its parent directories. Stopping on the first one it finds and returning
+    its path, returning None otherwise (if '/' was reached without finding such
+    a file).
 
-val list_packages : ?conf:string -> ?path:string list -> ?toolchain:string -> unit -> string list
+    A project configuration files is one of:
+    - .merlin
```
voodoos, line 55: @voodoos TODO Update

src/frontend/ocamlmerlin/new/new_commands.ml
voodoos: Path vers .merlin + lister les erreurs

src/kernel/mconfig_dot.ml
voodoos: Yes

src/frontend/ocamlmerlin/old/old_command.ml
voodoos: In some cases caller can pass an explicit .merlin file via a command line argument. In this case we should always use the `dot-merlin-reader`.

src/frontend/ocamlmerlin/old/old_command.ml
voodoos: In the old protocol this shoudl return the list of all .merlin
    files loaded for current buffer and a list of failures that might have
    happened during loading (missing package for instance, ill-formed .merlin,
    etc)

voodoos:
@trefis , @let-def : This PR should be ready for review. 

If you want to test it you need to pin the corresponding dune PR:
https://github.com/voodoos/dune.git#ocaml-merlin

From a user perspective this new version of Merlin does not require the latest dune as `.merlin` files will be loaded as usual (but with the new `dot-merlin-reader`) when they are present.
However the test suite will fail without the new dune PR because most of the tests do not have `.merlin` files and the closest config file in the tree is a `dune` file. Thus merlin tries to start the new dune configuration server and reports a failure.

How do you think we should mitigate this issue that will last until dune is upgraded ?

voodoos:
@trefis  @let-def 
This PR is now ready for review !

## Future steps

1. The next step are to merge this and to release of merlin. Merlin will then use the new `dot-merlin-reader` to get its configuration info.

2. Then it would be nice to update `ocaml-lsp-server`. A small patch is needed (https://github.com/voodoos/ocaml-lsp/commit/a5ebb66a86480b7da1e658cfe173f27028ddbf1c). @rgrinberg do you know if other projects are vendoring Merlin or should be notified of the disparition of the `.merlin` files ?

3. The final step will be to merge and release this Dune PR: https://github.com/ocaml/dune/pull/3554
After that dune will stop generating `.merlin` files and Merlin will start using `dune ocaml-merlin`  to get its configuration info.

As of today I am using a fork of `ocaml-lsp-server` with the new Merlin along with the new Dune and it works well. I will keep using it to catch any remaining issue.

tests/test-dirs/config/flags/invalid.t
```diff
@@ -2,7 +2,9 @@
   {
     "class": "return",
     "value": {
-      "dot_merlins": [],
+      "dot_merlins": [
+        "tests/test-dirs/config/flags/.merlin"
+      ],
```
trefis, line 7: Why did you add a `.merlin` file here?

.travis.yml
```diff
@@ -5,6 +5,9 @@ install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.t
 script: |
   bash -ex .travis-opam.sh
 env:
+  global:
+  - EXTRA_DEPS="dot-merlin-reader"
+  - PINS="dot-merlin-reader.dev:. merlin.3.3.7:. dune:https://github.com/voodoos/dune.git#merlin-abs-paths"
```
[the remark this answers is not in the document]
voodoos: I initially planned to keep it until the appropriate dune release. It was needed for a test that I added that called dune explicitly. 
However I removed this test because it would have caused the same issue as with release 3.3.5.
The pin should not be useful any more, I will remove it.

src/dot-merlin/dot_merlin_reader.ml
```diff
@@ -0,0 +1,488 @@
...
+open Std
+
+let () =
+  try Findlib.init ()
+  with exn ->
+    let message = match exn with
+      | Failure message -> message
+      | exn -> Printexc.to_string exn
+    in
+    prerr_endline ("Error during findlib initialization: " ^ message);
```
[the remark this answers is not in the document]
voodoos: It won't. And it does not happen at a time when Merlin expects output from the reader.

I can store that result and append an error message to followings answers of the reader. In that case they would be treated by Merlin as "config errors", not sure if all editor extensions show these to the user.

tests/test-dirs/config/flags/invalid.t
voodoos: To prevent triggering the dune reader and having a path not correctly sanitized by the wrapper:

`"/home/ulysse/tmp/merlin/dune-project"` 

I could also have fixed that by adding a sanitization rule to the merlin wrapper.

trefis: Oh right, that makes sense. Thanks.
It's fine like this.

src/dot-merlin/dot_merlin_reader.ml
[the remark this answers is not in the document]
trefis: > I can store that result and append an error message to followings answers of the reader

Yes, though perhaps append it only if people try to use findlib packages?
So there is no noise for people who don't need findlib.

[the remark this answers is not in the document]
voodoos: Should be solved by eef4a7959b216905e7e85588f8e0f10d0061e597

Issue #1125, opened 2020-04-08
Badly resolved ambiguity when a GADT constructor's name match the one of a module. Since OCaml 4.08


voodoos:
Given the following code:
```ocaml
type _ term =
  | Int : int -> int term
end
```

When asked the type of  `Int` Merlin wrongly answers `(module Stdlib__int)` since OCaml versions >= 4.08.

voodoos:
As with previous issues, the enclosing node shows a `core_type` first while `constructor_declaration` was expected.

```
current node is: [[ core_type; constructor_declaration; type_kind; type_declaration; structure_item; structure ]]
```

trefis:
Duplicate of #1226.

Issue #1124, opened 2020-04-08
Object type is returned when asking for only the type of a method since OCaml 4.08


voodoos:
Given the following code where `|` symbolizes the cursor:
```ocaml
let s = object
  val mutable v = [0; 2]
  method p|ush hd = v <- hd :: v
end
```

Merlin gives the type of the whole object  `< push : int -> unit >` instead of the type of the `push` method `int -> unit`.

When using OCaml versions < 4.08 the returned type is the correct one.

trefis:
Looking at the testsuite, this issue is still present on OCaml ≥ 4.08.

Issue #1109, opened 2020-03-04
Wrong result on a function named 'unit'


Julow:
Hi,

```ocaml
let rec unit x = x

and f : Unknown_module.t -> _ =
 fun x -> x

let g x = unit x
```

In this file, asking `:MerlinTypeOf` on any of the two `unit` occurrences gives:

```
type unit = ()
```

This happens only if `Unknown_module` is not found.

trefis:
@voodoos can you give an update on the status of this one?

voodoos:
I would say it is mostly fixed ?

https://github.com/ocaml/merlin/tree/master/tests/test-dirs/type-enclosing/recovery

Merlin gives two answers, the first one is correct, the second not really:
```

  $ $MERLIN single type-enclosing -position 5:11 -verbosity 0 \
  > -filename ./issue1109.ml < ./issue1109.ml | jq ".value[0:2]"
  [
    {
      "start": {
        "line": 5,
        "col": 10
      },
      "end": {
        "line": 5,
        "col": 14
      },
      "type": "'a -> 'b",
      "tail": "no"
    },
    {
      "start": {
        "line": 5,
        "col": 10
      },
      "end": {
        "line": 5,
        "col": 16
      },
      "type": "'a",
      "tail": "no"
    }
  ]
```

trefis:
Meh.
But yeah, let's close this for now. This should get better with a better recovery.

PR #1126, opened 2020-04-08
Add non-regression tests (and illustrates some bugs)

voodoos:
This PR adds a bunch of non-regression tests. The goal is to improve confidence
when tinkering with the type enclosing function of Merlin.

FIXME comments are added to tests that have wrong or unsatisfying outputs.

It also adds tests for the following issues:
#864, #1109,  #1124, #1125

Most of these issues and others should be fixed by PR #1108


trefis:
The travis failure on 4.08 is due to https://github.com/ocaml/merlin/commit/4f141992815109c64ebb0c399edbcff671f4f49b which updated menhir (Bad!). Otherwise everything looks good. I'm merging.

Thank you for this!

PR #1129, opened 2020-04-09
[Based on #1108] Use context to prevent typing of small enclosings in strings

### LINKED ISSUE #1116: Strings that contain typable names get wrongly typed
Hi !
In a code like this one for example:
```
let some_int = 5
let x = "some_int"
```

When hovering inside the string, the given type is `int` (first element returns by `type_enclosing`). Merlin should probably only return `string`. I do not know if this is desired behaviour or not.

See [this issue](https://github.com/ocaml/ocaml-lsp/issues/66) of ocaml-lsp for more details :)

Thank you !

voodoos:
This PR fixes #1116

If the context of a small enclosing is a string we do not try to type it anymore.

### DIFF
--- a/src/analysis/context.ml
+++ b/src/analysis/context.ml
@@ -41,6 +41,7 @@ type t =
   | Module_type
   | Patt
   | Type
+  | Constant
   | Unknown
 
 let to_string = function
@@ -50,6 +51,7 @@ let to_string = function
   | Module_path -> "module path"
   | Module_type -> "module type"
   | Patt -> "pattern"
+  | Constant -> "constant"
   | Type -> "type"
   | Unknown -> "unknown"
 
@@ -116,6 +118,7 @@ let inspect_expression ~cursor ~lid e : t =
       Expr
     else
       Module_path
+  | Texp_constant _ -> Constant
   | _ ->
     Expr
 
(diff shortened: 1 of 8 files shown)


voodoos:
For the record, to fix this issue in a satisfying way we must work on an other issue first:
Given the expression `(M|.A 3)` where `|` represents the cursor, Merlin will consider that the closest enclosing node is `const 3` which is wrong.

rgrinberg:
@voodoos anything blocking this PR?

voodoos:
> @voodoos anything blocking this PR?

I just rebased it. It is a bit hackish though, maybe @trefis will see a better solution ?

rgrinberg:
If you don't mind entertaining my curiosity, could you explain why it's hackish? The fix in this PR looks simple and straightforward to me.

voodoos:
Well, this specific change is not really hackish in itself you are right.
It's been a while but what I meant was that we would not need that patch if identifier reconstruction was working as intended.
(And when I was working on improving the Context detection I had to keep a series of fallbacks in `Type_utils.type_in_env` because of this.)

I think we should merge this, but keep in mind that there is more cleaning to do in `Type_enclosing`.

PR #1134, opened 2020-04-23
RFC: don't fail on directives when the file starts with `#!`

### LINKED ISSUE #1023: Since 4.08 merlin no longer likes toplevel directives
< 4.08 merlin didn't complain about `#use` and `#require` directives in ocaml scripts. Since 4.08 it seems it does ("Syntax error"). 

I didn't test 4.09.

hnrgrgr:
This MR tries to fix #1023 .

When a file starts with `#!` it is now parsed with `Parser_raw.Incremental.use_file` instead of the usual `Parser_raw.Incremental.implementation`.

In the first patch, all the toplevel directives are ignored. The second patch adds handling of `#require`. The third and last patch drafts an implementation of `#use` and `#mod_use`.

### DIFF
--- a/src/kernel/mreader.ml
+++ b/src/kernel/mreader.ml
@@ -16,6 +16,12 @@ type result = {
   no_labels_for_completion : bool;
 }
 
+let rec process_directives acc = function
+  | [] -> List.rev acc
+  | Parsetree.Ptop_dir _ :: phrases -> process_directives acc phrases
+  | Parsetree.Ptop_def items :: phrases ->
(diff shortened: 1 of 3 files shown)


hnrgrgr:
Ok, I clearly missed the point on handling multiple compiler version. And I missed the tricks used until ocaml-4.07 to handle `#require`.

What's the policy about modifying `parser_raw.mly` ? Would it be best to modify recent parser versions or best ot try backporting some compatibility layer to old ones ? I suspect the latter is more future proof.

PR #1137, opened 2020-05-15
Add keyword completion kind

rgrinberg:
This will allow the LSP server to request keyword completion explicitly. This seems to be a highly desired feature: https://github.com/ocaml/ocaml-lsp/issues/173

### DIFF
--- a/src/analysis/completion.mli
+++ b/src/analysis/completion.mli
@@ -56,14 +56,17 @@ val branch_complete
                -> [> `Found of string ])
   -> ?target_type:Types.type_expr
   -> ?kinds:Compl.kind list
+  -> keywords:string list
   -> string
   -> Mbrowse.t
   -> raw_info Compl.raw_entry list
 
 val expand_prefix
-  : global_modules:string list
+   : global_modules:string list
   -> ?kinds:Compl.kind list
-  -> Env.t -> string
+  -> keywords:string list
+  -> Env.t
+  -> string
   -> raw_info Compl.raw_entry list
 
 val application_context : prefix:Asttypes.label -> Mbrowse.t ->
--- a/src/frontend/query_commands.ml
+++ b/src/frontend/query_commands.ml
@@ -415,9 +415,10 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
         Some (Locate.get_doc ~config ~env ~local_defs
                 ~comments:(Mpipeline.reader_comments pipeline) ~pos)
     in
+    let keywords = Mpipeline.reader_keywords pipeline in
     let entries =
       Printtyp.wrap_printing_env env ~verbosity @@ fun () ->
-      Completion.branch_complete config ~kinds ?get_doc ?target_type prefix branch |>
+      Completion.branch_complete config ~keywords ~kinds ?get_doc ?target_type prefix branch |>
       print_completion_entries ~with_types config source
     and context = match context with
       | `Application context when no_labels ->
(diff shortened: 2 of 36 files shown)


avsm:
This would really help the VSCode plugin, as the immediate keyword completion of `in` to `in_channel` is the first thing that gets in the way.

jberdine:
Perhaps unrelated, but this makes me think that maybe this could help with avoiding the annoying behavior of `with` autocompleting to e.g. `with_return` in emacs.

src/frontend/query_commands.ml
```diff
@@ -243,6 +243,13 @@ let reconstruct_identifier pipeline pos = function
         aux acc (succ i) in
     aux [] offset
 
+let keywords pipeline kinds =
+  match kinds with
+  | []
+  | _ :: _ when List.mem `Keywords ~set:kinds ->
```
trefis, line 249: That's a bit weird. Why not make it an `if`?

src/kernel/mreader.ml
```diff
@@ -14,6 +14,7 @@ type result = {
   comments      : comment list;
   parsetree     : parsetree;
   no_labels_for_completion : bool;
+  keywords      : string list;
```
trefis, line 17: I don't think this needs to be stored in there. That list is going to stay the same across all the parse / buffers / etc.

src/analysis/completion.mli
```diff
@@ -47,6 +47,10 @@ val raw_info_printer : raw_info ->
   | `Concat of string * Extend_protocol.Reader.outcometree
   ]
 
+type keywords =
+  | Exclude
+  | Suggest of string list
```
trefis, line 52: Is there a good reason to pass a list of keywords as argument?
It seems we're always passing the same list (all the keywords), so perhaps the function should just be parametrized by `suggest_keywords:bool`?

src/ocaml/preprocess/411/lexer_raw.mll
```diff
@@ -140,6 +140,10 @@ let keyword_table : keywords =
     "asr", INFIXOP4("asr");
 ]
 
+let ocaml_keywords =
+  let l = lazy (Hashtbl.fold (fun k _ acc -> k :: acc) keyword_table []) in
+  fun () -> Lazy.force l
```
trefis, line 145: Did you notice an impact on startup time if you remove the layer of lazy + closure? That is: if you just compute the list of keywords during module initialization.
There aren't that many of them, so hopefully it should be fine.

trefis (changes requested):
This seems fine as a feature, but a couple of things need to happen before this gets merged:
- it needs rebasing (the blame is definitely on our side, sorry for not looking at this sooner). Hopefully this won't be too much work: most of the backends have disappeared!
- I don't think we need the keywords to bubble up through `Mreader_lexer`, `Mreader_parser`, etc. We can probably just access the lexer directly. And it's probably a bit cleaner if we get the list from `Completion` and not `Query_commands`
- I also left a couple of inline comments/questions

src/frontend/query_commands.ml
rgrinberg: No longer present in the up to date patch

src/analysis/completion.mli
rgrinberg: No longer present in the up to date patch

src/ocaml/preprocess/411/lexer_raw.mll
rgrinberg: As you mentioned, it wouldn't work without custom keywords. Although it makes sense to remove custom keywords altogether? I don't think anyone uses camlp4 anymore, so I'm not sure what the functionality is useful for .

rgrinberg:
@trefis I've updated the patch to the latest version actually used in lsp. I know you're working on your own patch, so this is just for your reference.

voodoos:
I guess we can close this PR now that #1243 is merged.

PR #1143, opened 2020-06-08
dune: remove duplicated rules for profile=release

avsm:
Currently, attempting to build in profile release mode will
result in dune complaining about duplicate rules:

```
$ dune build --profile=release
Error: Multiple rules generated for
_build/default/src/ocaml/preprocess/parser_recover.ml:
- _build/.dune/default/src/ocaml/preprocess/dune:14
- _build/.dune/default/src/ocaml/preprocess/dune:39
```

This fixes compilation of merlin in a dune monorepo with
`dune build --profile=release`.  I'm not quite clear on why
the opam merlin package worked, as I thought `-p` activates
the release profile...

### DIFF
--- a/src/ocaml/preprocess/dune
+++ b/src/ocaml/preprocess/dune
@@ -41,11 +41,13 @@ Printf.ksprintf J.send {|
 
 (menhir
  (modules parser_raw)
+ (enabled_if (<> %%{profile} "release"))
  (mode    (promote-into %s))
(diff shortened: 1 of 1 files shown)


nojb:
> I'm not quite clear on why the opam merlin package worked, as I thought `-p` activates the release profile...

It works because `-p` silently disables all rules with `(mode promote)`.

PR #1149, opened 2020-06-26
Support controlling ocamlmerlin socket with environment variable

lpw25:
Whilst benchmarking short-paths a few years ago, I found it useful to be able to control which socket ocamlmerlin used. It let me open several merlin servers in parallel without them interfering with each other. Someone else just mentioned they had a similar use case, so I thought I'd make a PR of the code. Fair warning: I haven't tested it, or even compiled it, since it was rebased.


ceastlund:
+1, I would find this useful.

PR #1151, opened 2020-07-03
Check if dune is compatible with new external reader

voodoos:
Merlin should check if the installed dune is recent enough to have the `dune ocaml-merlin` subcommand.

This is done in this PR by reading the output of the `dune ocaml-merlin` command: if it exited with status 0 and the printed version is `>= 2.7.0` then we enabled the functionality.

This check is done only once when building the initial configuration.

### DIFF
--- a/src/kernel/mconfig.ml
+++ b/src/kernel/mconfig.ml
@@ -120,6 +120,7 @@ let dump_merlin x =
       );
     "log_file"     , Json.option Json.string x.log_file;
     "log_sections" , Json.list Json.string x.log_sections;
+    "dune_support", `Bool x.dune_support;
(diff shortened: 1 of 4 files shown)


trefis:
I don't like this.

Can't we just fail gracefully when calling `dune` to get the config results in an error?

voodoos:
The thing is that if `dune ocaml-merlin` is not an option we want Merlin to switch to dot-merlin-reader if there is a `.merlin` file in a parent folder.

Thus if we detect this at the moment of launch we will have to redo the recursive search for a `.merlin`. 

Also dune 2.6 ships with an `ocaml-merlin` which is not fully operational. 

We should probably introduce some kind of protocol versioning.

PR #1153, opened 2020-07-13
Support 4.11

trefis:
This
- builds with 4.11 and all the features are implemented except `destruct` which needs a bit more work.
- does not build with any other version than 4.11
- is based of the latest released version, not master: so we can do a new release of that version with this 4.11 support

Currently a completion test is broken because some extra parentheses are added around some candidates (I don't know why).

Apart from that (ignorable) issue, and the missing destruct support (which is actually unlikely to appear before 4.11's release), I believe this might be good to merge.


trefis:
... should have made this a draft PR.
This should not be merged in master yes, it is purely FYI. The release should be made from the 411 branch.

Before merging in master the build should be fixed with older versions, this mainly implies unifying the way patterns are handled.
Which means:
- fixing destruct for 4.11
- updating `Raw_compat.Pattern` on all version to look like `Typedtree.pattern` in 4.11
- reintroduce the use of `Raw_compat.Pattern` in every place this PR currently removes it

trefis:
OK, all versions have now been reconciled.
So the remaining tasks are:
- fixing destruct
- understanding why `tests/test-dirs/completion/parenthesize.t` is broken on 4.11

voodoos:
I added the changelog entry for 4.11 support.

I guess we can, once reviewed, release directly from that branch which is based on 464df1a60775fd2bb70f4efff4f8603eaf834b74 (and thus does not include the new external config readers.

voodoos:
> 
> 
> Is a `(targets destruct.ml)` missing here?

It builds fine with latest dune, I believe the targets are automatically inferred ?

However it is possible that this behaviour is different in older version of dune. I will check that.

Did you encounter a build failure ?

voodoos:
> Is a `(targets destruct.ml)` missing here?

There was, indeed, a change in behaviour in dune 2.0.0 regarding target inference for the long form of the `rule` stanza:
https://github.com/ocaml/merlin/issues/1157
https://github.com/ocaml/dune/issues/3748

trefis:
Everything has been backported into master.

PR #1156, opened 2020-08-19
[Experiment] Use new dune's cram stanza

voodoos:
This experimental PR replace merlin testing plumbing, based on `mdx`, by the new dune's cram stanza.

### Pros
- No more `dune.inc` to generate
- Some free sanitization (we might remove some from merlin-wrapper)
- It looks like the separation of the various tests execution environments might be better. There is a small chance that this would solve the race issues in opam CI.
- No more test dependency on mdx. 
- The next standard for binary testing in OCaml ?

### Cons
- Requires dune 2.7
- Structure is a bit more rigid: two different tests cannot share the same data
- Dune outputs weird diffs with /dev/null which are not very useful... probably a bug in dune
- Had to reorganize and debug the tests

## TODO
- [ ] Decide if we want it
- [ ] Run the tests on OCaml < 410 to debug the test files specific to older versions


trefis:
I'd add another item to the "cons" list: it requires dune>=2.

rgrinberg:
> I'd add another item to the "cons" list: it requires dune > 2.

One way around this issue is to move the tests into a separate `dune-project`. The only downside is that it's quite ugly.

voodoos:
I rebased this PR and removed the tests for unsupported versions.
The upgrade to Dune 2.7 had an unexpected side effect: Dune now fails while trying to build the emacs plugin bytecode if the emacs binary is not installed.

It is a reasonable failure, what is weird is that this rule is executed even if it's target, the merlin bytecode file is not appearing as a dependency anywhere else. @rgrinberg : do you know if that behaviour is expected ?

Also, is there a way to condition the triggering a rule to the existence of a binary ?

rgrinberg:
> even if it's target, the merlin bytecode file is not appearing as a dependency anywhere else. @rgrinberg : do you know if that behaviour is expected ?

Are you using `@all` by chance?

voodoos:
> Are you using `@all` by chance?

The Ci was using `dune build` and thus `@all`. I was able to fix the problem, thanks.

tests/test-dirs/type-enclosing/constructors_and_paths.t/run.t
```diff
@@ -30,6 +30,8 @@ Various parts of the cons.ml:
     }
   ]
 
+".value[0:2]"
```
trefis, line 33: That seems weird

tests/test-dirs/file-cache.t/run.t
```diff
@@ -0,0 +1,66 @@
+The server might already be running, we kill it to make sure we start from a
```
trefis, line 1: This test was removed on master by 4885781c361177dd93b96deb35d13be94d6e38a0
I guess you missed that in the rebase and that's why you're reintroducing it. But it could be deleted.

PR #1158, opened 2020-08-28
Limit test jobs to one

voodoos:
This change may fix the OPAM CI which otherwise runs tests in parallel on 31 jobs.

### DIFF
--- a/merlin.opam
+++ b/merlin.opam
@@ -7,7 +7,7 @@ dev-repo: "git+https://github.com/ocaml/merlin.git"
 build: [
   ["dune" "subst"] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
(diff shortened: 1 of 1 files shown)


rgrinberg:
I'm not sure this is the right way to fix this. If the tests need to be run sequentially, why don't you make them acquire a lock?

voodoos:
I agree that this is not the right way to fix this.

However simply adding a lock to all tests is not a great solution either as it will speed down the tests when developing.
On my machine the number of jobs is 8 the issue never happen. The OCaml CI machine has 31 jobs.

We should identify precisely which tests need to be run sequentially and make these acquire a lock. 
And that can be done by dissecting OCaml CI logs. I will try to do that.

PR #1160, opened 2020-08-28
[Experiment] Add Github Actions workflow

voodoos:
Once the OCaml installation cached the CI should be faster than travis.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -1,0 +1,59 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    strategy:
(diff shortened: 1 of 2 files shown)


main.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          # Version of the OCaml compiler to initialise
+          ocaml-version: ${{ matrix.ocaml-version }}
+
+      - run: opam pin add merlin.dev . --no-action
+      - run: opam pin add dot-merlin-reader.dev . --no-action
+      - run: opam depext merlin --yes --with-test
+      - run: opam install . --deps-only --with-test
+      - run: opam install menhir
+      - run: opam exec -- dune build
```
trefis, line 58: I think with just this two lines, we might see some diff in the generated parsers.
Do we expect that to be an issue?

If yes, either we should install a precise version of menhir, or we should build in release mode.

voodoos:
First run without cache: 
- One build: 14m53s
- CI total: 17m23

Second run with cache of `.opam`:
- One build: 1m31s 
- CI total: 1m54s :tada:

GitHub will remove any cache entries that have not been accessed in over 7 days.
Max cache size is 5gb. This allows for roughly 10 OCaml installations.

I will add a cron to trigger the CI every 6 days...

PR #1165, opened 2020-08-31
Destruct tests 2

voodoos:
- Enable OCaml 4.05 and 4.06 in the CI
- Edit dune workspace, 4.11-beta -> 4.11
- Upstream "raise new exceptions on Destruct error" from ocaml-lsp
- Add some destruct tests


dune-workspace.template
```diff
@@ -8,4 +8,4 @@
 (context (opam (switch 4.08.1)))
 (context (opam (switch 4.09.0)))
 (context (opam (switch 4.10.0) (merlin)))
-(context (opam (switch ocaml-variants.4.11.0+beta2)))
+(context (opam (switch 4.11.0)))
```
trefis, line 11: Can we have 4.11.1 there or not yet?

voodoos: I guess so, this file is just for devs, the CI does not use it.

Issue #971, opened 2019-06-13
LSP : codelenses types are not using -short-path


jorisgio:
It seems that while merlin uses (a similar feature as ?) -short-path to display type of expressions uppon requests, types of top-level bindings displayed as codelenses are displayed with long path, which make them quite noisy. 
This happens with neovim + coc.nvim, but i assume it is indifferent of the editor. Any reason not to use short-path there too ?

jorisgio:
I had a quick look at this and my understanding is that it is not easy to tweak. As far as i understand, the reason of different behavior is that codelenses use outline that types every symbol in the context of the module, while type on hover uses the local type env.
So Outline types are shown in an env that has no opens/local opens. I can try to work on this but finding a good solution to solve it is way beyond my current understanding of how things work.

Bannerets:
@trefis 

> Can you give some example of what you're talking about? (i.e. for a given toplevel item, can you reproduce the type displayed in the "code lense", and the one returned by "type on hover"?)

If I understood the issue correctly:

![image](https://user-images.githubusercontent.com/10106819/68532279-51365a80-0313-11ea-8b5f-3fb90887b433.png)

(merlin `2b0ce3a42`)

etanter:
Hi, I'm the one who posted the linked issue from the VCode Ocaml Platform extension, which was tracked down to this Merlin issue. I see it has been open for over a year already -- is there any hope of seeing this fixed anytime soon?
Thanks!

PR #1166, opened 2020-09-02
Use local env when printing outline to benefit from -short-paths

let-def:
This should fix issue #971 : outline types are now printed using the local environment.
I don't know enough about LSP to test in this setting.


trefis (approved):

voodoos:
OCaml LSP is just calling Merlin's `outline` so if the paths are short there they will be short in the "code-lens" and "outline" features.

PR #1167, opened 2020-09-04
Testsuite: Add compatibility with clang when used as main C++ compiler

kit-ty-kate:
This fixes the issue encountered when testing merlin on FreeBSD:
```
#=== ERROR while compiling merlin.3.3.9 =======================================#
# context     2.0.6 | freebsd/x86_64 | ocaml-base-compiler.4.09.1 | file:///usr/home/travis/build/ocaml/opam-repository
# path        /usr/home/travis/.opam/ocaml-base-compiler.4.09.1/.opam-switch/build/merlin.3.3.9
# command     /usr/home/travis/.opam/ocaml-base-compiler.4.09.1/bin/dune runtest -p merlin -j 1
# exit-code   1
# env-file    /usr/home/travis/.opam/log/merlin-10058-d2816f.env
# output-file /usr/home/travis/.opam/log/merlin-10058-d2816f.out
### output ###
# [...]
- File "tests/test-dirs/pp/simple-pp.t", line 1, characters 0-0:
-          git (internal) (exit 1)
- (cd _build/default && /usr/local/bin/git diff --no-index --color=always -u tests/test-dirs/pp/simple-pp.t tests/test-dirs/pp/simple-pp.t.corrected)
- diff --git a/tests/test-dirs/pp/simple-pp.t b/tests/test-dirs/pp/simple-pp.t.corrected
- index 639d0d2..ee0d5c8 100644
- --- a/tests/test-dirs/pp/simple-pp.t
- +++ b/tests/test-dirs/pp/simple-pp.t.corrected
- @@ -28,6 +28,7 @@
-    > let x : int = 42 \
-    > #endif \
-    > EOF
- +  cpp: warning: argument unused during compilation: '-E' [-Wunused-command-line-argument]
-    {
-      "class": "return",
-      "value": [
- @@ -55,6 +56,7 @@
-    > let x : int = 42 \
-    > #endif \
-    > EOF
- +  cpp: warning: argument unused during compilation: '-E' [-Wunused-command-line-argument]
-    {
-      "class": "return",
-      "value": [],
```
*Testsuite entirely tested successfully using `FreeBSD 13.0-CURRENT arm64` locally*

PR #1168, opened 2020-09-04
Enable tests on MacOS

kit-ty-kate:
This is a follow-up of https://github.com/ocaml/merlin/pull/1167#issuecomment-686876059. I tested this on MacOS using Travis with this PR + https://github.com/ocaml/merlin/pull/1167 (see full logs: https://travis-ci.org/github/ocaml/opam-repository/jobs/724010830) and it works.

@rgrinberg what was the issue you were hitting in https://github.com/ocaml/merlin/pull/1041?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
   global:
   - EXTRA_DEPS="dot-merlin-reader"
   - PINS="dot-merlin-reader.dev:."
+  - ASSUME_ALWAYS_YES=YES
   matrix:
   - OCAML_VERSION=4.11
     PACKAGE=merlin
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: c
-sudo: true
 dist: xenial
 install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-opam.sh
 script: |
@@ -13,5 +12,9 @@ env:
     PACKAGE=merlin
 os:
   - linux
-  - macos
+  - osx
   - freebsd
+jobs:
+  include:
+  - os: linux
+    env: OCAML_VERSION=4.02 PACKAGE=merlin TESTS=false
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,27 +9,9 @@ env:
   - EXTRA_DEPS="dot-merlin-reader"
   - PINS="dot-merlin-reader.dev:."
   matrix:
-  - OCAML_VERSION=4.02
-    PACKAGE=merlin
-    TESTS=false
-  - OCAML_VERSION=4.03
-    PACKAGE=merlin
-  # - OCAML_VERSION=4.04
-  #   PACKAGE=merlin
-  # - OCAML_VERSION=4.05
-  #   PACKAGE=merlin
-  # - OCAML_VERSION=4.06
-  #   PACKAGE=merlin
-  # - OCAML_VERSION=4.07
-  #   PACKAGE=merlin
-  - OCAML_VERSION=4.08
-    PACKAGE=merlin
-  - OCAML_VERSION=4.09
-    PACKAGE=merlin
-  - OCAML_VERSION=4.10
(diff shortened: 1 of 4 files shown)


rgrinberg:
Could you add a macos job in github CI?

There are quite a few differences between gnu sed and bsd sed, so I don't recall what was the exact issue. I will try it again once I get some spare cycles.

voodoos:
Hum, the matrix is too big for the cache limit (5gb ~ 10 switches).
I think we should restrict the macos tests to the latest version.

kit-ty-kate:
That should be all good now. Everything should work out of the box without installing `coreutils` on the side. MacOS from Travis and MacOS from Github Actions were setup slightly differently (the later didn't have `readlink`) so I used the more portable (included in `bash` as builtin) `pwd -P`.

.travis.yml
```diff
@@ -32,3 +32,4 @@ env:
     PACKAGE=merlin
 os:
   - linux
+  - macos
```
voodoos, line 35: I believe the plan is to get rid of Travis sooner or later (@trefis ?), maybe we should not add any more load to it ? 

Or at least only on the latest version of OCaml because testing all versions would take ages...

rgrinberg:
I tested this PR on my mac and it seems to work. Thanks Kate!

rgrinberg:
There seems to be a travis failure:

```
- File "tests/test-dirs/destruct/exceptions_since_411.t", line 1, characters 0-0:
1217-          git (internal) (exit 1)
1218- (cd _build/default && /usr/local/bin/git diff --no-index --color=always -u tests/test-dirs/destruct/exceptions_since_411.t tests/test-dirs/destruct/exceptions_since_411.t.corrected)
1219- diff --git a/tests/test-dirs/destruct/exceptions_since_411.t b/tests/test-dirs/destruct/exceptions_since_411.t.corrected
1220- index cf23839..9fd6a86 100644
1221- --- a/tests/test-dirs/destruct/exceptions_since_411.t
1222- +++ b/tests/test-dirs/destruct/exceptions_since_411.t.corrected
1223- @@ -6,7 +6,7 @@
1224-    >   | exception _ -> () \
1225-    >   | Some 3 -> () \
1226-    > EOF
1227- -  {
1228- +  cannot open "/tmp/mlog2" for logging: Sys_error("/tmp/mlog2: Operation not permitted"){
1229-      "class": "error",
1230-      "value": "Destruct not allowed on computation pattern",
1231-      "notifications": []
1232- @@ -18,7 +18,7 @@
1233-    >   | exception _ -> () \
1234-    >   | Some _ -> () \
1235-    > EOF
1236- -  {
1237- +  cannot open "/tmp/mlog2" for logging: Sys_error("/tmp/mlog2: Operation not permitted"){
1238-      "class": "return",
1239-      "value": [
1240-        {
1241- @@ -58,7 +58,7 @@ FIXME: `Some 0` certainly is a missing case but we can do better
1242-    >   | Some 3 -> () \
1243-    >   | exception _ -> () \
1244-    > EOF
1245- -  {
1246- +  cannot open "/tmp/mlog2" for logging: Sys_error("/tmp/mlog2: Operation not permitted"){
1247-      "class": "return",
1248-      "value": [
1249-        {
1250[ERROR] The compilation of merlin failed at
1251        "/Users/travis/.opam/opam-init/hooks/sandbox.sh build dune runtest -p
1252        merlin -j 1".
```

voodoos:
My bad, I forgot to remove the `-log-file` in the destruct tests, and visibly travis doesn't allow writting to `/tmp`.
I will do that in another PR, it is not blocking this one.

voodoos:
Without `ASSUME_ALWAYS_YES` set to `YES` travis' FreeBSD build is stuck on `opam depext`.

It looks like this issue is already known... and fixed ?

In the CI script:
https://github.com/ocaml/ocaml-ci-scripts/blob/master/.travis-ocaml.sh#L315-L317

Merge opam PR:
https://github.com/ocaml-opam/opam-depext/pull/123

PR #1180, opened 2020-09-25
Add Ident.currentstamp to Local_store

trefis:
Minimal practical interest (perhaps we could have overflowed?), but it makes the usage of the store more regular.


let-def:
I fixed the build failures on 4.06, 4.07. They were due to predefined identifiers that are allocated before user idents (they start at 999).
Not a beautiful fix but goodenough... However it is not clear to me why it works on earlier versions, <=4.05

PR #1184, opened 2020-10-06
new Local_store API

trefis:
Some pre-upstreaming simplifications.
There are now two kinds of things that can be bound in the local store:
- tables (which are mutable) and must come with an initializer
- refs to immutable values: these take an initial value but get snapshoted once the bindings are frozen

This reimplement #1180 and a bit more using these new primitives.

### DIFF
--- a/src/kernel/mocaml.ml
+++ b/src/kernel/mocaml.ml
@@ -5,7 +5,7 @@ open Local_store.Compiler
 
 type typer_state = Local_store.scope
 
-let current_state = srefk None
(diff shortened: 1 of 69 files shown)


let-def:
Looks good to me, but  src/ocaml/typing/408/ident.ml.rej  should probably be removed.

trefis:
Woops :sweat_smile: 
Removed, thanks.

let-def:
Seal of approval

let-def (approved):

PR #1192, opened 2020-10-29
Modify naming convention for destruct sources

rgrinberg:
Use `destruct.%s.ml` rather than `destruct.ml.%s`. This convention has the
advantage of having editors automatically pick up the file type based on the
extension.


voodoos (approved):
Looks fine to me, I will merge it as soon as my next destruct-patch is done (#1194) .

voodoos:
Cherry-picked to master.

PR #1193, opened 2020-10-31
Replace appending list in a fold

rgrinberg:
It's faster to gather a list of lists and flatten later.

It's unlikely this optimization is very impactful, but it doesn't complicate the
code either.

### DIFF
--- a/src/analysis/completion.ml
+++ b/src/analysis/completion.ml
@@ -537,9 +537,10 @@ let complete_prefix ?get_doc ?target_type ?(kinds=[]) ~prefix ~is_label
           (kinds : kind list :> kinds)
       in
       let add_completions acc kind =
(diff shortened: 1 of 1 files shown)


src/analysis/completion.ml
```diff
@@ -537,9 +537,10 @@ let complete_prefix ?get_doc ?target_type ?(kinds=[]) ~prefix ~is_label
           (kinds : kind list :> kinds)
       in
       let add_completions acc kind =
-        get_candidates ?get_doc ?target_type ?prefix_path ~prefix kind ~validate env branch @ acc
+        get_candidates ?get_doc ?target_type ?prefix_path ~prefix kind ~validate env branch :: acc
```
voodoos, line 540: We could have taken the opportunity to break this line.

voodoos (approved):
Looks good to me

PR #1194, opened 2020-11-03
Destruct ill typed

voodoos:
This PR introduces type checking to the destruct `complete` feature.
This allows Merlin to filter out ill-typed witnesses obtained from `Parmatch.exhaust` (GADT example in `complete.t` tests). 

### Todo
- [x] Make compatible changes for other versions of OCaml
- [x] There is a small regression: record pattern contain useless `; _` field. (see [`complete.t` tests](https://github.com/ocaml/merlin/pull/1194/files?file-filters%5B%5D=.t#diff-644644c6605ed2c7f04f912d50bbb13e26017e20c7a68d15ec4c30558b6bf144R117)) (@trefis do you have an idea why this happen ?)

### DIFF
--- a/tests/test-dirs/destruct/complete.t
+++ b/tests/test-dirs/destruct/complete.t
@@ -89,7 +89,7 @@ Test 1.3 : with type constructor
 ## RECORDS ##
 #############
 
-Test 2.1 : FIXME : useless _ in record pattern
+Test 2.1
 
   $ cat >typ.ml <<EOF \
   > type a = A | B of string \
--- a/tests/test-dirs/destruct/complete.t
+++ b/tests/test-dirs/destruct/complete.t
@@ -114,7 +114,7 @@ Test 2.1 : FIXME : useless _ in record pattern
           "col": 19
         }
       },
-      "|{ a = B _;_} -> (??)"
+      "|{ a = B _ } -> (??)"
     ],
     "notifications": []
   }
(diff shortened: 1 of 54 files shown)


trefis:
I think if I had written this, I would have done the unmangling from inside parmatch itself, as I find these "renaming tables" a bit ugly (but I already found them ugly in the compiler), but I guess having our code *outside* of parmatch makes sense as well, it's good to limit our additions there I guess.
At least to start with.

This seems to work well, so I'd be in favor of merging if you can replicate the fix on older versions :)

trefis:
> There is a small regression: record pattern contain useless `; _` field.

One question: does this happen on all versions?
Apart from that, I guess it could just be some flag somewhere that gets set from `Open` to `Closed` (or vice-versa) during one of the  `typedree -> parsetree -> typedtree -> parsetree` transformations (given that 2 of them are introduced in this PR and weren't there before).

voodoos:
Yes it happen on all version, the testfile is common to all.

tests/test-dirs/destruct/complete.t
```diff
@@ -89,7 +89,7 @@ Test 1.3 : with type constructor
 ## RECORDS ##
 #############
 
-Test 2.1
+Test 2.1 : FIXME : useless _ in record pattern
```
trefis, line 92: isn't that fixed now?

trefis (approved):

PR #1195, opened 2020-11-10
Add test for type alias

voodoos:
Check that Merlin is not the cause of ocaml/ocaml-lsp#303

### DIFF
--- a/tests/dune.inc
+++ b/tests/dune.inc
@@ -1445,6 +1445,23 @@
 (alias (name runtest) (deps (alias type-enclosing-recovery-issue1109_pre_406)))
 
 (alias
+ (name type-enclosing-types-alias)
+ (package merlin)
+ (deps (:t ./test-dirs/type-enclosing/types/alias.t)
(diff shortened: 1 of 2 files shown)


voodoos:
@let-def @trefis : is it expected that the `-short-paths` option change the output on that example ?

```
Test 2.1

  $ $MERLIN single type-enclosing -position 2:5 -filename type_alias2.ml <<EOF \
  > type foo = int  \
  > let x = 1 \
  > EOF
  {
        "type": "int",
  }

Test 2.2 with shortpaths

  $ $MERLIN single type-enclosing -short-paths -position 2:5 -filename type_alias3.ml <<EOF \
  > type foo = int  \
  > let x = 1 \
  > EOF
  {
        "type": "foo",
  }
```

let-def:
@voodoos : This is the intended behavior I think.

```ocaml
utop # type foo = int;;
type foo = int
utop # 1;;
- : foo = 1
```

PR #1200, opened 2020-11-19
Wrap merlin_utils

rgrinberg:
It contains some commonly named modules that are best prefixed.

### DIFF
--- a/src/analysis/dune
+++ b/src/analysis/dune
@@ -1,6 +1,7 @@
 (library
  (name merlin_analysis)
  (wrapped false)
+ (flags -open Merlin_utils)
  (libraries
   config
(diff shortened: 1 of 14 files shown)


voodoos:
Looks good to me.

voodoos (approved):
Looks good to me

trefis:
I'm sorry, I should have been more careful, but this PR (and more recent follow up PRs) with its `(flags -open Foo)` instead of `(flags :standard -open Foo)` turned off all of dune standard flags (including a bunch of warnings) on a considerable part of the codebase.

I noticed this because I stumbled on some dead code which wasn't being reported by the compiler.

PR #1204, opened 2020-11-25
Add 4.12 support

trefis:
TODO:
- [x] make sure I haven't forgotten any patch
- [x] destruct is only offering one more case when called, instead of make the match exhaustive. That's due to the changes to parmatch making the exploration lazy. We should force the `Seq.t` to gather more branches (all of them? a fixed number? spend up to `X` ms on it?)
- [x] warnings output change to display the name of the warning. We could update the tests output but the info is redundant with the message, and clutters the display in some editors (e.g. vim), so we should probably just turn off that part of the printing.


trefis:
I need to update the CI to use 4.12, but apart from that this looks all done and is ready for review.
I changed the printing of warnings to be as compact as it was before (so not very compact) and I updated destruct so that it fully forces the sequence it gets from parmatch (this is strictly equivalent to what was done before).

trefis:
Note: I haven't pushed the update to `magic_numbers.ml`, so this is currently *not* usable.
I'll do so shortly.

kit-ty-kate:
Just a heads-up, OCaml 4.12.0~beta1 is going to be released tomorrow. Should a `3.4.3~4.12preview1` or something similar be released right after? (if you have time of course)

kit-ty-kate:
If you want to make the CI happy you can just pull https://github.com/ocaml/merlin/pull/1241 and it should be all green.

Issue #1202, opened 2020-11-25
Usability of MerlinLocate


mrmr1993:
Currently, `:MerlinLocate` in vim seems to ignore `g:merlin_locate_preference="ml"` when the target module has an interface file (at least when building with dune). In particular, locating a value `M.t` in a module with an interface file goes to the declaration in the signature instead of the definition.

This has never been particularly useful to me, since the value's information in the interface file can normally be gleaned from `:MerlinTypeOf` and `:MerlinDocument`. Worse still, if the interface file `include`s a signature that provides that particular item, merlin will jump to the declaration in the signature file, so the command may not leave you in the proximity of the implementation either.

For example, with the following setup, locating `Test.f` will land in `intf.ml`.
`test.ml`:
```ocaml
let f x = x
```
`test.mli`:
```ocaml
include Intf.S
```
`intf.ml`:
```ocaml
module type S = sig
  val f : int -> int
end
```

It seems that this is a consequence of the `.cmti` files generated from modules with interfaces: the locations for the signature items point to the interface file, not the implementation file, so that location is displayed.

It would be much more useful to modify the command (or add a new one? `:MerlinLocateDef`?) to identify the global module and locate the definition in its `.ml` file instead. Sadly, this seems to require merlin to parse and typecheck the file to find the definition, but should in theory be possible.

I would be willing to work towards this if it is viable.

#### Potential scope creep 1

Even nicer would be a stepping version of `:MerlinLocate` that remembers the current path that is being located. In the example above, this would find
* `test.mli`, at the `include` line
* `intf.ml`, at the `val f` declaration
* `test.ml`, at the `let f` definition

This would fully reveal all of the references that might be relevant, which I would find especially useful when modifying existing code. It would still also be nice to have a straight-to-definition version as above, even if this extension is worth exploring.

This naturally also has an impact on and increased complexity from module inclusion. For example, replacing `test.ml` with
```ocaml
include Test_impl
```
(where `Test_impl` defines the value `f`) adds an additional location to this list, as well as potentially more for any interface files and their inclusions if there is a `test_impl.mli`.

#### Potential scope creep 2

In code with heavy use of functors, it can be awkward to find the definition of a function. For example, if the above `Test_impl` was implemented as
```ocaml
module F (T : sig val f : 'a -> 'a end) = struct
  include T
end
module T = struct
  let f x = x
end
include F (T)
```
[...]

PR #1208, opened 2020-11-30
Add MerlinLocateImpl, MerlinLocateIntf commands

mrmr1993:
I added these as part of my debugging for #1202, but they have turned out to be generally useful and low-cost to add.

The commands behave as `MerlinLocate`, but jumping to either the implementation or interface regardless of the contents of `g:merlin_locate_preference`.

### DIFF
--- a/vim/merlin/autoload/merlin.vim
+++ b/vim/merlin/autoload/merlin.vim
@@ -385,23 +385,17 @@ function! merlin#Locate(...)
 endfunction
 
 function! merlin#LocateImpl(...)
-  if (a:0 > 1)
-    echoerr "Locate: too many arguments (expected 0 or 1)"
-  elseif (a:0 == 0) || (a:1 == "")
-    MerlinPy merlin.vim_locate_choice_under_cursor("implementation")
-  else
-    MerlinPy merlin.vim_locate_choice_at_cursor(vim.eval("a:1"), "implementation")
-  endif
+  let l:pref = g:merlin_locate_preference
+  let g:merlin_locate_preference = 'implementation'
+  call call("merlin#Locate", a:000)
+  let g:merlin_locate_preference = l:pref
 endfunction
 
 function! merlin#LocateIntf(...)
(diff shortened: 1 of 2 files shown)


vim/merlin/autoload/merlin.vim
```diff
@@ -376,14 +376,34 @@ endfunction
 
 function! merlin#Locate(...)
   if (a:0 > 1)
-    echoerr "Locate: to many arguments (expected 0 or 1)"
+    echoerr "Locate: too many arguments (expected 0 or 1)"
```
trefis, line 379: :sweat_smile:

vim/merlin/autoload/merlin.vim
```diff
@@ -376,14 +376,34 @@ endfunction
...
   elseif (a:0 == 0) || (a:1 == "")
     MerlinPy merlin.vim_locate_under_cursor()
   else
     MerlinPy merlin.vim_locate_at_cursor(vim.eval("a:1"))
   endif
 endfunction
 
+function! merlin#LocateImpl(...)
```
trefis, line 387: Naïve question (because I haven't touched vimscript in years), why can't this function be implemented as:
```vim
let l:pref = g:merlin_locate_preference
let g:merlin_locate_preference = 'ml'
call merlin#Locate(...)
let g:merlin_locate_preference = l:pref
```

trefis (commented):
These new functions seem redundant with the variable that's already there.
But if people really think they make for a more convenient user interface then why not...

mrmr1993:
Pushed a small change to enable completion for the commands.

vim/merlin/autoload/merlin.vim
mrmr1993: Updated to do roughly this. Vimscript being vimscript, your elegant call syntax needed to be replaced with something much uglier and less obvious:
```vim
call call("merlin#Locate", a:000)
```

trefis:
Thanks!

PR #1211, opened 2020-11-30
Remove unused Emacs rule

### LINKED ISSUE #1210: Current master requires emacs for compilation
It seems that the change of the dune language version from 1.8 to 2.7 in [this commit](https://github.com/ocaml/merlin/commit/89c23b36692d4b2599523be3ccce24d2efb24b6f) causes merlin to fail to compile if no emacs is available, since the emacs folder is now build too.

voodoos:
Fixes #1210 

The target of the incriminating rule is not used anywhere else in the project. It is very unlikely that is is used by anyone.
This PR removes the rule.

### DIFF
--- a/emacs/dune
+++ b/emacs/dune
@@ -9,7 +9,3 @@
         (merlin-xref.el as emacs/site-lisp/merlin-xref.el)
         (merlin.el as emacs/site-lisp/merlin.el)))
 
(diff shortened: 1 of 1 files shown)


bschommer:
With this change the build works for me again.

PR #1213, opened 2020-12-02
Fix for Neovim's CursorMoved semantics

ddickstein:
Neovim's CursorMoved semantics differ from Vim's:

Vim: After the cursor was moved in Normal or Visual mode.
Neovim: After the cursor was moved in Normal or Visual mode **or to another window**.

Because we do `exe l:cur . "wincmd w"` after setting up the auto-commands, in Neovim they would fire immediately, causing the type history window to instantly close. This patch defers the auto-command setup until after we've run `wincmd w`.


bfredl:
Be aware that this can happen in vim 8 as well. Neovim just chooses to be consistent.

trefis:
Thanks!

PR #1217, opened 2020-12-06
Add OCAMLMERLIN_PATH env var to facilitate Merlin development

ddickstein:
Allow OCAMLMERLIN_PATH env var to override the computed path for the ocamlmerlin
binary. This facilitates iterative development on Merlin.


ddickstein:
No longer convinced this is a good idea. Closing.

PR #1218, opened 2020-12-06
Fix [List.uniq] to dedup non-adjacent elmements.

ddickstein:
[List.uniq] only dedups adjacent duplicates. This is sufficient for
[List.sort_uniq], but can be surprising when invoked directly. This commit
patches [List.uniq] to dedup the whole list. It is also now used when retrieving
source paths, which can be duplicates but for which we might want to preserve
ordering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _build
+_opam
 .merlin
 jbuild-workspace
 dune-workspace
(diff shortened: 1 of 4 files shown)


trefis:
The rename of "uniq" to "dedup_adjancent" seems good.
As for the new implementation of `uniq` that you propose, I believe that's already implemented as `filter_dup`.

rgrinberg:
ping @ddickstein

rgrinberg:
LGTM. @voodoos shall we merge?

Issue #894, opened 2018-12-28
MerlinLocate not resolving definition of core modules


ddickstein:
If I have an OCaml file as follows:
```
open Core
module L = List
...
```
and I try to `MerlinLocate` on `List`, I get the following error:
```
Several source files in your path have the same name, and merlin doesn't know which is the right one:
/usr/local/lib/ocaml/list.ml, /usr/local/lib/ocaml/list.mli, /Users/dani/.opam/system/lib/core_kernel/
list.ml, /Users/dani/.opam/system/lib/core_kernel/list.mli, /Users/dani/.opam/system/lib/base/list.ml,
 /Users/dani/.opam/system/lib/base/list.mli
```
Why is Merlin having trouble selecting the right one? I'm using Merlin version 3.2.2.

(Following up on issue 2 in https://github.com/ocaml/merlin/issues/863. The behavior I had been seeing was that this issue occurred when trying to go to definition a second time, as though the locate was cached and there was an issue with the cached version, but now the issue happens even on the first locate request)

b0o:
@ddickstein I know this issue is really old, but did you ever figure out a solution? I am experiencing the same problem.

ddickstein:
My understanding is that [duniverse](https://github.com/avsm/duniverse), when it arrives, will solve this problem.

ddickstein:
My understanding is that duniverse <https://github.com/avsm/duniverse>,
when it arrives, will solve this problem.

trefis:
I don't think "the duniverse" will help at all with this.
You have several "list.ml" in your path right now, that'll still be the case with the duniverse.

That being said, I just tried to reproduce the issue right now and wasn't able to. (Though I used `Base` and not `Core`).
There have been a lot of fixes on locate since this issue was created. And perhaps this one is fixed and @b0o is actually encountering a different problem?

trefis:
@ddickstein: can you confirm whether you can reproduce the issue or not?

ddickstein:
Yes, this problem is still occurring for me (using merlin 3.3.2).

I'm using dune 1.9 (set in my dune-project file), an empty dune workspace file, and my dune file itself where I'm testing the jump to definition looks like this:

```
(library
  (name rocket_game)
  (inline_tests)
  (libraries core sexplib)
  (preprocess (pps ppx_jane)))
```

[...]

PR #1219, opened 2020-12-06
Fix [locate] to look for non-pp source files.

ddickstein:
When ppx is used, the cmt sourcefile that is recorded is the *.pp.mli? file,
and we compare its digest to the digests of the files in the source directory.
Often the .pp.mli? file isn't there, and even when it is we probably do not
want to prefer it over the normal .mli? file for locate. This patch checks if
the file is a .pp.mli? file, and if so uses the digest of the regular .mli?
file instead.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@ git version
     - filter dups in source paths (#1218)
     - improve load path performance (#1323)
     - fix handlink of ppx's under Windows (#1413)
+    - locate: look for original source files before looking for preprocessed
+      files (#1219 by @ddickstein, fixes #894)
 
 merlin 4.4
 ==========
@@ -151,7 +153,7 @@ Tue Feb  2 03:13:37 PM CET 2021
     - emacs: add missing mandatory argument for define-obsolete-function-alias
       (#1250, by Atharva Shukla, fixes #1234)
     - emacs: use "opam var" instead of "opam config var" (#1249, by Raja Boujbel)
-    - vim: fix CursorMoved semantics (#1213, by Daniel Dickstein)
+    - vim: fix CursorMoved semantics (#1213, by @ddickstein)
     - vim: add :MerlinLocateImpl and :MerlinLocateIntf (#1208 by Matthew Ryan)
   + test suite
     - replace mdx usage by dune's cram mechanism
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -316,15 +316,24 @@ let move_to_root root cmt_infos =
   let digest =
     (* [None] only for packs, and we wouldn't have a trie if the cmt was for a
        pack. *)
-    let sourcefile = Option.get cmt_infos.Cmt_format.cmt_sourcefile in
-    match String.split_on_char ~sep:'.' sourcefile |> List.rev with
+    let sourcefile_in_builddir =
+      Filename.concat
+        (cmt_infos.Cmt_format.cmt_builddir)
+        (Option.get cmt_infos.cmt_sourcefile)
+    in
+    match sourcefile_in_builddir |> String.split_on_char ~sep:'.' |> List.rev with
     | ext :: "pp" :: rev_path ->
-      (* If the source file was a post-processed file (.pp.mli?), use the regular
-         .mli? file for locate. *)
-      let builddir = cmt_infos.cmt_builddir in
-      let sourcefile = (ext :: rev_path) |> List.rev |> String.concat ~sep:"." in
-      (match Misc.exact_file_exists ~dirname:builddir ~basename:sourcefile with
-       | true -> Digest.file (Filename.concat builddir sourcefile)
+      (* If the source file was a post-processed file (.pp.mli?), use the
+         regular .mli? file for locate. *)
+      let sourcefile_in_builddir =
+        (ext :: rev_path) |> List.rev |> String.concat ~sep:"."
+      in
+      (match
+         Misc.exact_file_exists
+           ~dirname:(Filename.dirname sourcefile_in_builddir)
+           ~basename:(Filename.basename sourcefile_in_builddir)
+       with
+       | true -> Digest.file sourcefile_in_builddir
        | false -> Option.get cmt_infos.cmt_source_digest)
     | _ -> Option.get cmt_infos.cmt_source_digest
   in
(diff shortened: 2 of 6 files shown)


ddickstein:
This patch resolves https://github.com/ocaml/merlin/issues/894.

I've attached a small dune project that reproduces the problem. Inside bin/main.ml, if you try jumping to definition on `Good` it will work, but on `Bad` it will fail w/ "Several source files in your path have the same name, and merlin doesn't know which is the right one." With this patch, jumping to definition on `Bad` now works too. I also confirmed that this resolves the failure flagged in the issue with jumping to the definition of `Core.List`.

ddickstein:
Note: If you modify `lib/core_kernelish/dune` by removing `(preprocess (pps ppx_jane -check))`, you will then be able to jump to definition of `Bad` in `bin/main.ml`.

trefis:
Disclaimer: I haven't looked at your reproduction case yet.
In the general case, I don't think your idea of jumping to the `.ml` corresponding to a `.pp.ml` would be correct: some preprocessing will have happened, things will have moved. So you are going to jump to the right (perhaps) file but not to the right location.
But I guess that primarily, I'm confused regarding why a `.pp.ml` would be fed to the compiler in the first place.

ddickstein:
@trefis have you had a chance to take a closer look at the repro case? Is a `.pp.ml` being fed to the compiler a dune bug?

rgrinberg:
> Is a .pp.ml being fed to the compiler a dune bug?

This behavior is intended. When dune preprocesses a module, it saves the result into `module.pp.ml`. This is done to share the preprocessing step between ocamlc & ocamldep.

trefis:
@rgrinberg and how does that work w.r.t. error messages locations? Do you introduce line directives in the `.pp.ml`?
Couldn't you simply save a serialized AST instead of a text file?

rgrinberg:
> Do you introduce line directives in the .pp.ml

Well behaving preprocessors will insert such directives.

> Couldn't you simply save a serialized AST instead of a text file?

We do if the preprocessor is a ppx. The ppx driver will emit the serialized AST.

Sorry for the late reply. Slipped by my inbox.

trefis:
> We do if the preprocessor is a ppx. The ppx driver will emit the serialized AST.

Even if it's not, I think a majority of the reasonable preprocessors are able to emit a serialized AST.

I finally looked at @ddickstein's example (sorry for the delay!) and there is indeed a `.pp.ml`, which contains a serialized ast!
So the locations should indeed be correct and the approach proposed here seems reasonable.

I just pushed a (probably incorrectly) minimized version of your reproduction case to master, as a test (which you can build with `dune build @dot-pp-dot-ml`), which exposes the behavior you describe before the PR, but gives the following error with the patch proposed in this PR:
```
I/O error: $TESTCASE_ROOT/_build/libb/dep.ml: No such file or directory
```

I say "probably incorrectly minimized" because it's lacking one level of indirection (:exploding_head: :gun:) so we're probably going to be missing on crucial digest along the way.
But right now we're not even reaching that point. I'm not quite sure what is happening, it could be that the call to `Digest.file` is failing (but it'd be surprising, because the file we get an error on does exist at the time).
I don't quite have the time to debug this right now, so you guys should feel free to have a go at it.

rgrinberg:
I finally had a look. Sorry for the delay.

> which exposes the behavior you describe before the PR, but gives the following error with the patch proposed in this PR:

Probably because your test doesn't copy the source files to the build directory like dune does.

Regarding the PR, it seems like an improvement but it should probably be resilient to the source being absent from the build dir and fail more gracefully.

ddickstein:
Any advice on how to get the new test to play nicely on Windows? It works on my Linux machine.

Also, if "A" and "B" in the test are replaced with (), you'll see https://github.com/ocaml/merlin/issues/1403

ddickstein:
Update: fixed the issue on MacOS -- now only failing on Windows.

ddickstein:
@voodoos Do you have any suggestions on how to get the Windows build working? I tried to use [act](https://github.com/nektos/act) to run the workflows locally but I ran into a "Parameter token or opts.auth is required" error that I think may be due to my repo being a fork and not having access to the Github secrets.

voodoos:
> Do you have any suggestions on how to get the Windows build working? 

Not sure what is the issue here, my guess would be that there is an problem with the path starting by `/tmp` in:
```
 (cd _build/default && C:\Windows\system32\cmd.exe /c "/tmp/project/prep.exe libb/dep.ml") > _build/default/libb/dep.pp.ml
```

However we don't have a strict policy for Windows testing yet. If patching the test is too complicated you can simply disable it (see https://github.com/ocaml/merlin/pull/1298).

voodoos:

ddickstein:
@voodoos sure - how's "locate: look for original source files before looking for preprocessed files (#1219, fixes #894)"

voodoos:
> @voodoos sure - how's "locate: look for original source files before looking for preprocessed files (#1219, fixes #894)"

Sounds good to me

voodoos:
And you can add your github handle to it, we usually credit explicitly external contributors :-)

ddickstein:
I'm not sure why debian-11-4.13_x86_32 is now failing. Is that architecture test newly added or is the test flaky?

voodoos:
Looking again at the test I think we should avoid escaping the sandbox.

However I tried to rework it but it doesn't work quite as expected.
My idea is to have `.merlin` files to use `dot-merlin-reader` and build with `--root=.`.
But the last test does not pass.
Interestingly, it doesn't work either when reproducing the test myself in another folder 🤔 

I won't have the time to look more closely for the next week, here is a gist with my attempt if you want to play with it:
<script src="https://gist.github.com/voodoos/4192730f056ffb5ef939e900255ed93d.js"></script>

voodoos:
@ddickstein I just force-pushed to your branch to be able to benefit from recent changes to merlin-wrapper.

I rewrote the test to stay in the sandbox (that is not using the /tmp directory). However the last test is still not working. Would you mind reading again the test case to check that I did not add any mistake ? I am wondering if that PR actually solves the issue or not...

ddickstein:
> I'm not sure why debian-11-4.13_x86_32 is now failing. Is that architecture test newly added or is the test flaky?

This is still happening with the latest sandboxing changes.

voodoos (approved):
Looks good to me, thank you !
I can merge by squashing everything or if you want you can rebase first and squash some of the commits. Like the iterations on the tests, the back and forth in merlin-wrapper etc.

ddickstein:
Squashing everything seems fine to me. Thank you!

PR #1225, opened 2020-12-16
Give expected type of holes when using `type-enclosing`

voodoos:
This PR adds new constructors for the parsetree and typedtree expressions that represents holes in a program. 
The concrete syntax for theses holes in expressions is `_`. (this can be easily changed if needed)

This way holes are no more identifiers and Merlin can easily give their type, that is the type expected in the hole.


voodoos:
Closing in fa or of #1242

Issue #1216, opened 2020-12-04
hover over type with mutual recursion fails silently


ulugbekna:
Hover over a type using a mutually recursive type crashes ocaml-lsp silently (ocaml-lsp usually shows thrown exceptions, but not in this case). Say, we have 

```ocaml
type t = s
and s = string
```

hovering over `s` crashes ocaml-lsp with no stack trace, so possibly segfaulting 🤷‍♂️

Note: the problem is discovered using ocaml-lsp, so that might not be merlin's fault; sorry for the noise if that's the case.

trefis:
I don't know what hovering is supposed to trigger in lsp; two reasonable guesses would be "show the type" or "show occurences", both work for me on your example.

ulugbekna:
Thanks for a prompt response!

ulugbekna:
Example I gave in the issue description fails with a stack overflow for me when I run command 

```sh
ocamlmerlin single type-enclosing -position 1:10 -verbosity 1 -\
  > -filename ./test.ml < ./test.ml | jq
```
"hovering over" `s` in 

```ocaml
type t = s and s = string;;
```

@trefis  were you running the same merlin command when you said it was working well for you? 

---

Btw, ocaml-lsp uses `type enclosing` query command for hovering feature.

---

Strangely, for this case, `utop` on `#show t` returns `type nonrec t = t`, which is wrong (?), but `ocaml` returns `type nonrec t = s`. Utop's behavior seems to hint why we could be getting a stack overflow.

trefis:
I had the `verbosity` flag set at 0 (which is the default), but indeed when you increase the value, then you get a stack overflow.
Thanks for the report, we'll look into it.

Btw: is `-verbosity 1` the default in lsp? That seems like it could be a mistake.

ulugbekna:
> Btw: is -verbosity 1 the default in lsp? 

Yes, see this [line](https://github.com/ocaml/ocaml-lsp/blob/cb8aba2238ba5b0e5a6fe0a3180f25077496f75a/ocaml-lsp-server/src/document.ml#L136)

---

> That seems like it could be a mistake.

I don't fully understand what `verbosity` does, but sometimes it can return strange types. Fro example, for a file `t.ml` with contents:

```ocaml
type 'a t = 'a
```

we can run `type-enclosing`  and see a strange type `type 'a t = 'b t` as the head of the list `value`

```sh
[...]

PR #1240, opened 2021-01-18
type-enclosing result on type expressions

trefis:
This PR is in response to #1230 (cc @ulugbekna).

First, I discovered that after rebasing past #1232 (which was merged last week), all but one of the FIXMEs were indeed fixed (hurrah!).
The diff is visible in https://github.com/ocaml/merlin/commit/5125e54a7aa033d444514c3af1122076845d57b6.

As for the remaining FIXME, it doesn't look like a bug, but rather like the expected behavior of `-short-paths`.
You can see that if we remove that flag, the result is the one you seem to expect @ulugbekna.

However, going back to the initial discuss of `verbosity`'s value (started in #1216), you can also notice that with that flag, the result stays the same regardless of the verbosity level.

### DIFF
--- a/tests/test-dirs/type-enclosing/types.t/run.t
+++ b/tests/test-dirs/type-enclosing/types.t/run.t
@@ -113,6 +113,45 @@ FIXME: A type with a type param shouldn't equal itself - aliasing a list type
     }
   ]
 
+Same result regardless of verbosity:
+
+  $ $MERLIN single type-enclosing -short-paths -position 11:9 -verbosity 1 \
+  > -filename ./types.ml < ./types.ml | jq ".value"
+  [
+    {
+      "start": {
+        "line": 11,
+        "col": 0
+      },
+      "end": {
+        "line": 11,
+        "col": 19
+      },
+      "type": "type 'a l = 'a l",
+      "tail": "no"
+    }
+  ]
+
+OK without -short-paths
+
+  $ $MERLIN single type-enclosing -position 11:9 -verbosity 0 \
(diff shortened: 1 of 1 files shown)


ulugbekna:
Thanks for taking time to investigate this! :-)

Did I get you right that presence of `short-paths` flag "disables" `verbosity` levels?

trefis:
> Did I get you right that presence of short-paths flag "disables" verbosity levels?

Only when the type is a simple alias, if there is more structure, then verbosity still comes into play:
```
$ cat sp.ml
type ('a, 'b) t = fst:'a -> int -> snd:'b -> unit

type 'a foo = unit -> (float, 'a) t
$ ocamlmerlin single type-enclosing -position 3:8 -index 0 -filename sp.ml -verbosity 0 < ./sp.ml | jq ".value[].type"
"type 'a foo = unit -> (float, 'a) t"
% ocamlmerlin single type-enclosing -position 3:8 -index 0 -filename sp.ml -verbosity 1 < ./sp.ml | jq ".value[].type"
"type 'a foo = unit -> fst:float -> int -> snd:'a -> unit"
% ocamlmerlin single type-enclosing -position 3:8 -index 0 -filename sp.ml -short-paths -verbosity 0 < ./sp.ml | jq ".value[].type"
"type 'a foo = unit -> (float, 'a) t"
% ocamlmerlin single type-enclosing -position 3:8 -index 0 -filename sp.ml -short-paths -verbosity 1 < ./sp.ml | jq ".value[].type"
"type 'a foo = unit -> fst:float -> int -> snd:'a -> unit"
```

PR #1242, opened 2021-01-22
[Construct] Part 0 - add AST holes, fix typing of holes and add a new `holes` command

voodoos:
This PR is a preliminary for the future `construct` command.

- add test for the typing of identifiers with `_` in their names
- adds a new AST node `Pexp_hole` whose (non-definitive) concrete syntax is `_`.
- fix the typing of these holes (they used to always be typed as `'a` which is not very useful)
- add a new command `holes` that returns the list of the locations of all the holes of a file

~~There is an issue that neeeds to be fixed. These changes created some conflicts in the parser:~~ FIXED
```
Warning: 2 states have shift/reduce conflicts.
Warning: 2 shift/reduce conflicts were arbitrarily resolved.
```


voodoos:
I fixed the parsing conflict.

voodoos:
Maybe this PR should not change the holes syntax (ie keep `(??)`) as `_` probably won't be the definitive one.
The important thing is to add an AST node for holes, the syntax can be easily changed.

src/analysis/destruct.new.ml
```diff
@@ -74,7 +74,7 @@ module Predef_types = struct
 end
 
 let placeholder =
-  Ast_helper.Exp.ident (mk_id "??")
```
trefis, line 77: This file is a left-over from when we had several backends, it should be deleted (perhaps as a preliminary PR?)

src/ocaml/merlin_specific/browse_raw.ml
```diff
@@ -894,3 +894,9 @@ let node_of_binary_part env part =
   | Partial_module_type x ->
     Module_type x
 
+let rec all_holes (env, node) =
+  let f env node acc = match node with
+    | Expression { exp_desc = Texp_hole; exp_loc; _}  -> exp_loc :: acc
+    | _ -> List.rev_append (all_holes (env, node)) acc
```
trefis, line 900: I'm a bit surprised to see this `rev_append` inside a fold.
Why aren't we just folding down?

src/ocaml/preprocess/parser_raw.mly
```diff
@@ -651,8 +651,7 @@ let expr_of_lwt_bindings ~loc lbs body =
   let default_loc = ref Location.none
 
   let default_expr () =
-    let id = Location.mkloc "merlin.hole" !default_loc in
-    Exp.mk ~loc:!default_loc (Pexp_extension (id, PStr []))
+    Exp.mk ~loc:!default_loc (Pexp_hole)
```
trefis, line 654: I don't think I like that diff, `default_expr` is inserted in place of expression nodes we couldn't parse. So if you insert `Pexp_hole`s the result of the `holes` command will contain garbage on files that don't parse.

I agree that these nodes and `(??)` inserted by the user were indistinguishable before, but now they probably should be?

trefis (commented):
This mostly looks good, modulo the comment I left in the parser which worries me a little.

src/ocaml/preprocess/parser_raw.mly
voodoos: I pushed a commit (a955e71) reverting the changes to the recovered "holes".
They are represented by `merlin.hole` extension points as before.

src/ocaml/merlin_specific/browse_raw.ml
voodoos: `fold_node` is not really folding on all the nodes but all the "surface" nodes. To go deeper we need additional recursive calls and thus some lists concatenation.

src/ocaml/typing/typecore.ml
```diff
@@ -2463,7 +2464,7 @@ let check_partial_application statement exp =
             | Texp_setinstvar _ | Texp_override _ | Texp_assert _
             | Texp_lazy _ | Texp_object _ | Texp_pack _ | Texp_unreachable
             | Texp_extension_constructor _ | Texp_ifthenelse (_, _, None)
-            | Texp_function _ ->
+            | Texp_function _ | Texp_hole ->
```
trefis, line 2467: I'm not sure that's right, I think this could result in spurious warnings (spurious because the code hasn't been written yet).
I think `| Texp_hole -> ()` would probably be better.

src/ocaml/typing/typecore.ml
```diff
@@ -3837,22 +3838,29 @@ and type_expect_
...
+                          val_type = ty_expected;
+                          val_kind = Val_reg;
+                          val_loc = loc;
+                          val_attributes = [];
+                          val_uid = Uid.internal_not_actually_unique;
+                        });
+          exp_loc = loc; exp_extra = [];
+          exp_type = instance ty_expected;
+          exp_attributes = attr :: sexp.pexp_attributes;
+          exp_env = env }
```
trefis, line 3863: Can you make the (1 space) indentation change disappear? :)

tests/test-dirs/recovery.t
```diff
@@ -10,7 +10,7 @@ type error.
   }
 
 This incomplete expression should generate only a parser error.
-The hole is filled with merlin.hole.
+The hole is filled with Pexp_hole.
```
trefis, line 13: This hunk should be reverted.

tests/test-dirs/type-enclosing/underscore-ids.t
```diff
@@ -0,0 +1,309 @@
...
+          "line": 3,
+          "col": 6
+        },
+        "type": "unit -> int",
+        "tail": "no"
+      }
+    ],
+    "notifications": []
+  }
+
```
trefis, line 63: I'm a bit confused by all the tests below: I'm sure they're exercising the identifier reconstruction code, but we can't really observe the result, can we?

voodoos: Because all that count is the return type annotation ?
I should define the value first and check the type like the first test I guess.

trefis (approved):
LGTM.

PR #1243, opened 2021-01-22
completion: add "keyword" kind

trefis:
Alternative implementation of #1137

The code is very similar, but it's on an up-to-date merlin.
Also, it made me realize that all the plumbing @rgrinberg had implemented is actually necessary if you want to also list keywords added by extensions (which you can't if you just access `Lexer_raw.keyword_table`).

### DIFF
--- a/src/analysis/completion.ml
+++ b/src/analysis/completion.ml
@@ -354,6 +354,7 @@ let get_candidates ?get_doc ?target_type ?prefix_path ~prefix kind ~validate env
         1000 - cost
     in
     let of_kind = function
+      | `Keywords -> [] (* cannot happen after a dot. *)
       | `Variants ->
         let add_variant name param candidates =
           if not @@ validate `Variant `Variant name then candidates else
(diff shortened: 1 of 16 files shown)


rgrinberg:
I just realized that the code in #1137 is out of date. The latest version lives in https://github.com/rgrinberg/merlin/tree/lsp

It's quite similar to #1137 but it adds keyword completion by context courtesy of @mnxn: https://github.com/rgrinberg/merlin/commit/30d1c87fd00c3f1414df90e1d8e822e2b2e01c3f

trefis:
I'm not fond of the way the context filter is implemented, so I'll merge this PR as it is now.
And we can revisit the filtering later, hopefully that's alright with you.

bluddy:
OCaml's `in` keywords and their lack of completion are a very serious language-writing issue given LSP. This needs to be sorted out. It's hitting me in neovim and I thought it was some kind of bug.

PR #1244, opened 2021-01-22
adds module, module type, and class imenu items

ivg:
These are important syntactic categories, which I am missing a lot. In fact, the modules category is the most important as it enables to quickly move between submodules and significantly reduces cognition burden during developing OCaml programs. The class category fits into the same niche, except when the OOP style is employed. There is also a constructor category, though I am not really sure if is useful, and given that it can pollute the outline I decided to leave it.

### DIFF
--- a/emacs/merlin-imenu.el
+++ b/emacs/merlin-imenu.el
@@ -18,6 +18,9 @@
 (defvar-local merlin-imenu--value-list nil)
 (defvar-local merlin-imenu--type-list nil)
 (defvar-local merlin-imenu--exception-list nil)
+(defvar-local merlin-imenu--module-list nil)
+(defvar-local merlin-imenu--signature-list nil)
+(defvar-local merlin-imenu--class-list nil)
 
 (defun merlin-imenu-compute-position (line col)
   "Get location of the item."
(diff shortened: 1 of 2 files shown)


ivg:
The dot-merlin-read lint failure looks irrelevant to the changed code, as far as I can see.

emacs/merlin-imenu.el
```diff
@@ -58,6 +61,12 @@
              (setq merlin-imenu--value-list (cons marker merlin-imenu--value-list)))
             ((string= kind "Type")
              (setq merlin-imenu--type-list (cons marker merlin-imenu--type-list)))
+            ((string= kind "Module")
+             (setq merlin-imenu--module-list (cons marker merlin-imenu--module-list)))
+            ((string= kind "Modtype")
```
mndrix, line 66: If I'm reading the output of `(merlin-call "outline")` correctly, I think `"Modtype"` should be `"Signature"`. The entries for `"Module"` and `"Class"` look correct though.

emacs/merlin-imenu.el
```diff
@@ -78,6 +90,12 @@
       (push (cons "Exception" merlin-imenu--exception-list) index))
     (when merlin-imenu--type-list
       (push (cons "Type" merlin-imenu--type-list) index))
+    (when merlin-imenu--module-list
+      (push (cons "Module" merlin-imenu--module-list) index))
+    (when merlin-imenu--modtype-list
+      (push (cons "Modtype" merlin-imenu--modtype-list) index))
```
mndrix, line 96: If my above comment about `"Signature"` is correct, it might be worth using the same terminology here for consistency.

mndrix (commented):
I would really enjoy having modules, module types, and classes in imenu. I haven't tried running with this patch yet, so my comments are based strictly on inspection.

emacs/merlin-imenu.el
ivg: Done, as well as the variable names.

emacs/merlin-imenu.el
ivg: I can confirm that your reading is correct. Fixed now.

voodoos:
@ivg could you rebase and add a changelog entry ?

ivg:
@voodoos, done!

voodoos:
Thank you !

PR #1241, opened 2021-01-20
Make OCaml-CI happy

kit-ty-kate:
`ocaml-dune-lint` used by OCaml-CI checks that every libraries explicitly used by a project is listed in the opam file.

PR #1248, opened 2021-01-27
Build cleanup

trefis:
The first patch is an alternative to #1241 (thanks to @kit-ty-kate for noticing the issue) which removes the dependency on menhirLib from dot-merlin-reader (instead of reflecting it in the opam file).
The second commit cleans up some legacy support that I noticed while working on the first patch.


trefis:
Ah, I was wrong, it's not a complete alternative to #1241, which also fixed merlin.opam.
It's an alternative just for the dot-merlin-reader part. Anyway...

PR #1253, opened 2021-02-02
Fix tests in release mode

kit-ty-kate:
This should fix https://github.com/ocaml/opam-repository/pull/18074#issuecomment-771721700
I'm still unsure why it happens but at least it fixes the issue.

### DIFF
--- a/tests/dune
+++ b/tests/dune
@@ -20,8 +20,7 @@
  (name test-deps)
  (deps
   %{bin:merlin-wrapper}
-  (package merlin)
-  (package dot-merlin-reader)))
+  (package merlin)))
 
 (cram
  (applies_to :whole_subtree)


kit-ty-kate:
If you're fine with the fix it might need to be cherry-picked to master too

trefis:
I don't think that fix makes sense, at least: I don't understand how it could.
Our tests depend on (and cover) both merlin and dot-merlin-reader.

kit-ty-kate:
but merlin depends on `dot-merlin-reader` already so by transitive dependency it should be fine.

trefis:
That's true at the opam level, but not the dune level.
If we want the tests to run again when `dot-merlin-reader` changes (but `merlin` itself is unchanged) or when it isn't built at all, then we need this.

trefis:
(or I think we do. But perhaps there's some dune magic I'm not aware of)

trefis:
I believe #1259 and #1260 fix this in a more satisfying way.
Thanks though!

PR #1254, opened 2021-02-02
Make canonical Windows paths more canonicals

voodoos:
These changes ensure that the `/ ` or `\` in the drive part of a Windows path (`c:\`) is correctly converted to the current platform separator.

It also always make the drive letter uppercase. This seems to be the standard (even if Windows understand both) and matches `Sys.getcwd ()` style.

This fixes communication issues between Merlin and Dune's `ocaml-merlin` configuration server on WIndows.

# Todo
- [ ] Propagate to 411
- [ ] Propagate to 412

### DIFF
--- a/src/utils/misc.ml
+++ b/src/utils/misc.ml
@@ -116,9 +116,19 @@ let remove_file filename =
   with Sys_error _msg -> ()
 
 let rec split_path path acc =
-  match Filename.dirname path, Filename.basename path with
-  | dir, _ when dir = path -> dir :: acc
-  | dir, base -> split_path dir (base :: acc)
+  match Filename.dirname path with
+  | dir when dir = path ->
+    let is_letter c = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') in
+    let dir =
+      if String.length dir > 2 && is_letter dir.[0] && dir.[1] = ':'
+      then
+        Printf.sprintf "%s%s"
+          (String.sub dir ~pos:0 ~len:2)
+          Filename.dir_sep
+      else dir
+    in
+    dir :: acc
+  | dir -> split_path dir (Filename.basename path :: acc)
 
 (* Deal with case insensitive FS *)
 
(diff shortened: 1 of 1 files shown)


nojb:
> It also always make the drive letter uppercase. This seems to be the standard (even if Windows understand both) and matches `Sys.getcwd ()` style.

Note that the case of `Sys.getcwd ()` is not well-defined. Recently we fixed a similar problem in `dune` and made the opposite choice (lowercase) https://github.com/ocaml/dune/pull/3966

src/ocaml/utils/misc.ml
```diff
@@ -111,9 +111,19 @@ let remove_file filename =
...
+  match Filename.dirname path with
+  | dir when dir = path ->
+    let is_letter c = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') in
+    let dir =
+      if String.length dir > 2 && is_letter dir.[0] && dir.[1] = ':'
+      then
+        Printf.sprintf "%s%s"
+          (String.sub dir ~pos:0 ~len:2 |> String.uppercase_ascii)
+          Filename.dir_sep
+      else dir
```
nojb, line 123: This only normalizes the backslash following the  drive letter. Are we sure this is enough? I would have thought that under Windows we should normalize all slashes. I will try it and report back.

voodoos: Other slashes were already normalized as expected: `split_path` returns a list of path segments which are concatenated back using `Filename.concat`.

voodoos:
> > It also always make the drive letter uppercase. This seems to be the standard (even if Windows understand both) and matches `Sys.getcwd ()` style.
> 
> Note that the case of `Sys.getcwd ()` is not well-defined. Recently we fixed a similar problem in `dune` and made the opposite choice (lowercase) [ocaml/dune#3966](https://github.com/ocaml/dune/pull/3966)

So it it is a very recent change, that is why in your example the detected workspace root is still capitalized.

voodoos:
@trefis while Dune 2.9 is not release it is not obvious to determine if the drive letter should be capitalized or not. I added a revolting hack c8bf493 to retry loading the configuration with an uppercased drive letter when failing.

Issue #1256, opened 2021-02-02
merlin + dune + windows hangs


nojb:
I'm trying `merlin` 3.4.2 with `dune` 2.8.2 under native Windows (msvc) on Cygwin (no `opam`). When launching `merlin` the first time (ie the server is not running) under `emacs` the whole process hangs waiting for the spawned `dune` process to finish. If I open a follow-up session then `merlin` seems to work as expected. If I stop the server, the hanging behaviour comes back.

Glad to provide more details if it helps debugging.

nojb:
(same problem when using `dot-merlin-reader` by the way)

nojb:
I did a bit of debugging, it seems this call
https://github.com/ocaml/merlin/blob/59965113df9f7af1bb68d9b86ab8abcd768227a4/src/frontend/ocamlmerlin/ocamlmerlin.c#L638
is blocking... but am not sure what that means yet...

voodoos:
Should be fixed with the latest release (4.2 / 3.5.0). Please reopen if the issue persists.

PR #1263, opened 2021-02-08
Dune helper: don't inherit stderr

let-def:
Dune helper used to cause Emacs to block under windows. This should no longer be the case.

It happened because Merlin's stderr file descriptor was inherited by Dune helper.
This is not a problem under Unix because replacing stderr causes the previous file descriptor to be closed. Under windows, the previous file handle stays open even if it is no longer stderr.

@voodoos @nojb

### DIFF
--- a/src/kernel/mconfig_dot.ml
+++ b/src/kernel/mconfig_dot.ml
@@ -139,7 +139,20 @@ module Configurator = struct
       let stderr_r, stderr_w = Unix.pipe () in
       Unix.chdir dir;
       Unix.set_close_on_exec stdin_w;
+      (* Set close on exec for stderr: most processes spawned by merlin
+         are supposed to inherit stderr to output their debug information.
+         This is fine because these processes are short-lived.
+         However the dune helper we are about to spawn is long-lived, which can
+         cause issues because its stderr is going to outlive the one of
+         merlin's client process.
+         Under windows, this makes emacs block, synchronously waiting for
+         stderr to be closed.
+      *)
+      Unix.set_close_on_exec Unix.stderr;
       let pid = Unix.create_process prog args stdin_r stdout_w stderr_w in
+      (* Clear close on exec after spawning, such that other processes still
+         inherit it. *)
+      Unix.clear_close_on_exec Unix.stderr;
       Unix.chdir cwd;
       Unix.close stdin_r;
       Unix.close stdout_w;


trefis (approved):
:heart_eyes:

nojb:
Thanks @let-def ! I tried this with the set up of #1256 , I get the following error (inside `emacs`) : 
`flag -filename: error, Unix.Unix_error(1, "set_close_on_exec", "")`. Any ideas?

nojb:
(I cherry-picked your commit to the 3.4.2 branch)

let-def:
@nojb Worrying, I didn't manage to reproduce yet, I will look again tomorrow.
However, can you try again after commenting the `Unix.clear_close_on_exec Unix.stderr;` line?

Looking at OCaml unix source, the error message is wrong, `"set_close_on_exec"` message is raised both by set and clear commands, and it will help to be sure where it fails. (The win32unix implementation is different, but since we are on cygwin, I guess it is the normal unix implementation that is used?)

nojb:
> @nojb Worrying, I didn't manage to reproduce yet, I will look again tomorrow.
> However, can you try again after commenting the `Unix.clear_close_on_exec Unix.stderr;` line?

No difference, but the failing call is `set_close_on_exec` (see below)

> Looking at OCaml unix source, the error message is wrong, `"set_close_on_exec"` message is raised both by set and clear commands, and it will help to be sure where it fails. (The win32unix implementation is different, but since we are on cygwin, I guess it is the normal unix implementation that is used?)

I am not using Cygwin-compiled binaries, but rather binaries compiled using the `msvc` compiler, so we really are using `win32unix`.

let-def:
Thanks for the explanation. Here it seems the bug is reliably fixed on Cygwin.
I am not sure how to investigate on MSVC, I will keep you posted.

let-def:
@nojb I think I have a good fix. I tested with Mingw and MSVC. (curiously, I observed differences in handle management).

nojb:
> @nojb I think I have a good fix. I tested with Mingw and MSVC. (curiously, I observed differences in handle management).

Thanks for investigating this tricky bug! I confirm that `emacs` no longer blocks with this commit.

let-def:
Thanks for testing. The change should be released soon.

nojb:
> Thanks for testing. The change should be released soon.

Backporting to the 3.4.2 branch would also be appreciated. Thanks!

mroch:
sorry for pinging such an old PR. can you explain the difference between the original version of this PR ([e904010](https://github.com/ocaml/merlin/commit/e904010404449ff3620d94f102995612dcf318d6)) using `Unix.set_close_on_exec` and the final version ([9a7e589](https://github.com/ocaml/merlin/commit/9a7e58996d580f3f6fd43fd16f1fd92311fddc78)) that directly calls `SetHandleInformation`?

it seems like [Unix.stderr](https://github.com/ocaml/ocaml/blob/000d15d4e229e895612177590df08d0364b1cb04/otherlibs/unix/unix_win32.ml#L259) calls `_get_osfhandle` ([here](https://github.com/ocaml/ocaml/blob/742a6d2ac02a68a6b5c9ba6e42206dc4cc201898/otherlibs/unix/channels_win32.c#L164)) and then `Unix.set_close_on_exec` calls `caml_win32_set_inherit` ([here](https://github.com/ocaml/ocaml/blob/000d15d4e229e895612177590df08d0364b1cb04/otherlibs/unix/close_on.c#L22)) which calls the same `SetHandleInformation` ([here](https://github.com/ocaml/ocaml/blob/663e8d219f566095e3a9497c5bae07b6a95cae39/otherlibs/unix/unixsupport_win32.c#L343-L345))

I ask because we're doing the same thing using `Unix.set_close_on_exec` in Flow [here](https://github.com/facebook/flow/blob/c03e28279d8c2f5979dc4ef9c3d4d25ebaba3169/src/monitor/flowServerMonitorDaemon.ml#L106-L129).

Issue #1191, opened 2020-10-25
"unknown flag '-package'" when using merlin-use since 3.3.7


eugecm:
Using ocaml `3.10.1` and merlin `4.10.1` I get the following error when running the `merlin-use` command from emacs:

> unknown flag "-package"

It looks like the `ocamlmerlin` command is being constructed with a `-package` option, as can be seen in this line https://github.com/ocaml/merlin/blob/e0a3061e2c13748928bc77b5dc20ce207ad1218d/emacs/merlin.el#L496

However, said flag is not referenced in the source code other than in the emacs file since 7a45a1f99ea88e68b35ad67aa67d4455caa50523 according to git-bisect

Works fine in merlin `3.3.6`

voodoos:
Thanks for your precise report. We indeed dropped support for the `-package` flag and thus the `merlin-use` feature when moving to external configuration reader. The recommended way to indicate packages it to use the `.merlin` file.

However, crashing is not a good option and an informative message should be displayed instead.

c-cube:
@voodoos is there a reason why `-package ` was dropped? It'd be very useful to me in some contexts where merlin isn't invoked via dune.

voodoos:
One of the reasons for using external configuration readers was to make the core of Merlin free of its ocamlfind dependency and this feature also depended on it.

However I heard that @let-def was planning to reintroduce that feature directly in the editor plugins

voodoos:
The last release of Merlin introduce a new editor-side "merlin use" commands by @def.

Can you confirm that these answer to your use cases ?

> Merlin used to support the -package flag to load an ocamlfind/findlib package. This was exposed by the :MerlinUse and merlin-use commands in vim and emacs.
> 
> The commands are still there but the support for -package was dropped when packaging logic was moved to dune (via the reader helper). So they are broken in the latest release...
> 
> This PR reimplements the command so they work entirely on editor side: ocamlfind is invoked to list packages and resolve package names to include path. Only theses paths are passed to merlin using the -I flag.
> 
> Tradeoffs:
> 
>  - ocamlfind is called from the editor environment... the answers might be out of date if the user changed to another opam switch. Workaround: maybe we should call opam exec ocamlfind instead of ocamlfind? (That's much slower... but maybe we shouldn't mind?)
> -  only the include path is considered... This won't work with ppx and other fancy features supported by ocamlfind. I don't mind these use cases, any thoughts?

PR #1265, opened 2021-02-09
fix same file documentation bug

### LINKED ISSUE #1261: Documentation missing for references to things defined in the same file
Given an input, `a.ml`:

```ocaml
(** f docs *)
let f = ()

(** g docs *)
let g = ()

let h = g
```

When the point is on the reference to `g`, all is well:

```
$ ocamlmerlin single document -verbosity 0 -position 7:9 < a.ml | jq .value
"g docs"
```

However, if we change `h` to be `let h = f`, the documentation for `f` is not returned:

```
$ ocamlmerlin single document -verbosity 0 -position 7:9 < a.ml | jq .value
"No documentation available"
```

If we remove the documentation comment for `g` though (note the changed line number to correspond), the documentation for `f` _is_ found:

```
$ ocamlmerlin single document -verbosity 1 -position 6:9 < a.ml | jq .value
"f docs"
```

It seems that, for lookups of things defined in the same file, merlin will not return documentation for any but the final documented definition.

(Originally reported in https://github.com/ocamllabs/vscode-ocaml-platform/issues/519, where @ulugbekna suggested the actual failure pattern.)

ulugbekna:
Hi! 

Fixes #1261 

tl;dr `Ocamldoc.associate_comment` needs a list of doc-comments from the start to the end of file (non-decreasing positions of comments), but `Locate.get_doc` returned the comments from the end to the start of the file (non-increasing positions of comments). The last doc-comment was available because the last elem of non-decr list = head of non-incr list.

More detailed: 

`Ocamldoc.associate_comment` takes a list of doc-comments (with their positions) from a file and compares positions of those comments with the position `p` of the cursor. If a doc-comment `d` with position `p1` has `p1 > p` (ie documentation occurs later in the file than `p`), it considered there the list doesn't contain documentation for position `p`. (see [code](https://github.com/ocaml/merlin/blob/2a0dd5c16178efcc4b8132e3c3b3c2a6cc7b13ea/src/analysis/ocamldoc.ml#L42-L46))

Since `Locate.get_doc` was returning `List.rev comments`, it returned doc-comments in a non-decreasing order, which meant that head of the list would have position "larger" than `p` causing the `associate_comment` return `None`.

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -832,7 +832,7 @@ let get_doc ~config ~env ~local_defs ~comments ~pos =
   | `Found (loc, None) ->
     let comments =
       match File_switching.where_am_i () with
-      | None -> List.rev comments
+      | None -> comments
       | Some cmt_path ->
         let {Cmt_cache. cmt_infos; _ } = Cmt_cache.read cmt_path in
         cmt_infos.Cmt_format.cmt_comments
--- a/tests/test-dirs/src-documentation/run.t
+++ b/tests/test-dirs/src-documentation/run.t
@@ -1,0 +1,38 @@
+documentation for the last defined value (in the same file) is shown
+  $ $MERLIN single document -position 7:10 <<EOF
+  > (** first function *)
+  > let f () = ()
+  > 
+  > (** second function *)
+  > let g () = ()
+  > 
+  > let () = g (f ())
+  > 
+  > let list_rev = List.rev
+  > EOF
+  {
+    "class": "return",
+    "value": "second function",
+    "notifications": []
+  }
+
+documentation for the non-last defined value (in the same file) is show
+(we care about "non-last" value because of issue #1261)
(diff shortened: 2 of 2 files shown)


voodoos:
Thanks, these changes look good to me.
I wonder why these `List.rev` was used in the first place, @trefis ?

It might just  be a mistake; or the comments list was previously in the wrong order ? Or `Ocamldoc.associate_comment` was too aggressively optimized at some point...

I am also curious about the role of `File_switching.where_am_i ()` ?

ulugbekna:
Thanks Ulysse!

In retrospective, I should've probably added tests to show that there are
no regressions. I have tested manually however if that counts :p

voodoos:
It's never to late to add tests 🙂

voodoos:
This release should fix #1288, #1300, #714, #1261, #1191, ocaml/dune#4479.

PR #1270, opened 2021-02-16
system command variant: do not open a window console when launching a…

let-def:
… ppx under windows

This is an alternative to #1101 and should fix the same problem.
This patch does not go through `cmd.exe` and that seems to improve the latency (in my informal experiments).

### DIFF
--- a/src/platform/platform_misc.c
+++ b/src/platform/platform_misc.c
@@ -1,6 +1,7 @@
 #include <caml/mlvalues.h>
 #include <caml/memory.h>
 #include <caml/alloc.h>
+#include <stdlib.h>
 
 /* FS case */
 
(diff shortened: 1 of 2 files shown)


MisterDA:
I confirm that this patch works in my setup, on top of v4.1-411. I vote in favor of inclusion.

src/platform/platform_misc.c
```diff
@@ -54,6 +55,37 @@ value ml_merlin_dont_inherit_stdio(value vstatus)
...
+    PROCESS_INFORMATION p_info;
+    STARTUPINFOW s_info;
+    DWORD ReturnValue;
+
+    memset(&s_info, 0, sizeof(s_info));
+    memset(&p_info, 0, sizeof(p_info));
+    s_info.cb = sizeof(s_info);
+
+    wchar_t utf16cmd[MAX_SYSTEM_PROGRAM] = {0};
+    MultiByteToWideChar(CP_UTF8, 0, cmd, -1, utf16cmd, MAX_SYSTEM_PROGRAM);
```
voodoos, line 73: I am curious about this conversion, and `MAX_SYSTEM_PROGRAM`.
I guess if comes from [this SO answer](https://stackoverflow.com/a/57798301) ?
Do you now why it is useful ?

it is not used in Microsoft's documentation : https://docs.microsoft.com/en-us/windows/win32/procthread/creating-processes

MisterDA: Regarding the conversion, `CreateProcessW` expects unicode (UTF-16) and OCaml uses a different encoding (most likely UTF-8).
https://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#s%3Ainterfacing-windows-unicode-apis

```c
#define CAML_NAME_SPACE
#define CAML_INTERNALS
#include <caml/misc.h>
#include <caml/osdeps.h>
// ...
char_os *utf16cmd;
utf16cmd = caml_stat_strdup_to_os(cmd);
// ...
caml_stat_free(utf16cmd);
```

The maximum length used here seems arbitrary. The limit to `cmd.exe` is [8191 characters](https://docs.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation).

voodoos: Oh I missed the `W`.

MisterDA: And I missed to say that the code snippet would be better than the call to `MultiByteToWideChar` and remove the length limitation.

MisterDA:
If `CreateProcessW` fails the function returns an uninitialised value `ReturnValue`. 
What version of C are you using? If C89, then you should move the `utf16cmd` declaration up to not mix code and variable decls.

let-def:
> If `CreateProcessW` fails the function returns an uninitialised value `ReturnValue`.
> What version of C are you using? If C89, then you should move the `utf16cmd` declaration up to not mix code and variable decls.

Thanks for catching the missing return value. C99 is fine :)

PR #1271, opened 2021-02-16
Use relative paths to communicate with `dune ocaml-merlin`

voodoos:
Starting with Dune 2.8.3 the `dune ocaml-merlin` configuration server will handle both paths that are absolute or relative to the starting directory of the server. (ocaml/dune#4221)

Relative paths will become the new preferred way to communicate with the configuration server as it is more local and flexible.

This PR will make Merlin attempt send relative path to `dune ocaml-merlin`. 
If that fails Merlin will fallback to sending an absolute path as before.


trefis (approved):
Fine

voodoos:
@trefis I added a test illustrating the fact Merlin always ask for a relative path and then an absolute path if it fails.

The relative path is correctly constructed when the `-filename` param is a relative path, an absolute path ignoring symlinks and an absolute path with symlink.

The absolute path might or might not contain the symlink. This is the problem this issue is solving by using relative paths.

trefis:
:+1:

PR #1272, opened 2021-02-23
vim & emacs : implement MerlinUse/merlin-use/-package in editor modes

let-def:
Merlin used to support the `-package` flag to load an ocamlfind/findlib package. This was exposed by the `:MerlinUse` and `merlin-use` commands in vim and emacs.

The commands are still there but the support for `-package` was dropped when packaging logic was moved to dune (via the reader helper). So they are broken in the latest release...

This PR reimplements the command so they work entirely on editor side: `ocamlfind` is invoked to list packages and resolve package names to include path. Only theses paths are passed to merlin using the `-I` flag.

Tradeoffs:
- ocamlfind is called from the editor environment... the answers might be out of date if the user changed to another opam switch. Workaround: maybe we should call `opam exec ocamlfind` instead of `ocamlfind`? (That's much slower... but maybe we shouldn't mind?)
- only the include path is considered... This won't work with ppx and other fancy features supported by ocamlfind. I don't mind these use cases, any thoughts?

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -225,6 +225,9 @@ The association list can contain the following optional keys:
 (defvar-local merlin-buffer-packages nil
    "List of packages loaded in the buffer")
 
+(defvar-local merlin-buffer-packages-path nil
+   "List of path of packages loaded in the buffer")
+
 (defvar-local merlin-buffer-extensions nil
    "List of syntax extensions active in the buffer")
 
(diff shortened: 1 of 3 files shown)


trefis (approved):

let-def:
Merging as it is. We can always add `opam exec` later if it proves necessary.

Issue #1268, opened 2021-02-16
flet is obsolete since emacs 24.3


bschommer:
During startup I get a warning from emacs:
````
‘flet’ is an obsolete macro (as of 24.3); use either ‘cl-flet’ or ‘cl-letf’
````

I'm not that versed in lisp to fix this but I think it can be done rather straightforward.

voodoos:
Should be fixed with the latest release (4.2 / 3.5.0).

PR #1247, opened 2021-01-27
Modernisation of elisp code and conformance with coding guidelines

purcell:
I set about tidying up the merlin elisp to match GNU and package conventions and best practices. By targeting a post-2015 Emacs, we also improve the out-of-the-box experience by automatically enabling the xref backend. Everything is split up into self-explanatory commits, and I'm happy to make adjustments if desired.

I've tested this locally, including its operation with the `merlin-eldoc` package. (I actually think the core eldoc behaviour should be included in `merlin` itself and I might make a follow-up PR to add that.)

PR #1274, opened 2021-02-26
Use cl-letf instead of flet. (#1268)

bschommer:
This is what emacs suggests doing in the deprecation warning. This should fix the deprecation warning at startup.


bschommer:
I think the ci failures are independent of the change.

bschommer:
Will be fixed in #1247

PR #1276, opened 2021-02-26
Completion: use type information

trefis:
Type recovery on record degraded over the past few years, which in turns degraded completion results when inside a record expression.
I reintroduced a naïve recovery for record, but it needs to be reviewed (in particular: I am not calling `Msupport.erroneous_type_register`, should I?).

The other commit extends constructor completion candidates with the constructor from the expected type which matches the given prefix.

### DIFF
--- a/src/analysis/completion.ml
+++ b/src/analysis/completion.ml
@@ -276,6 +276,17 @@ let fold_variant_constructors ~env ~init ~f =
   in
   aux init
 
+let fold_sumtype_constructors ~env ~init ~f t =
+  let t = Ctype.repr t in
+  match t.desc with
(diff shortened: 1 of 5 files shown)


let-def (approved):
The only purpose of erroneous type register is to silence the unused argument warning. Not using it seems fine for the current purpose.

(Worst case I can think of: `ty_expected` is `'_a`, typing the record fails and it is immediately passed to e.g `ignore` -- I am not 100% sure the conditions for the unused argument warning)

trefis:
> The only purpose of erroneous type register is to silence the unused argument warning. Not using it seems fine for the current purpose.

Ah, indeed!
Thanks.

PR #1277, opened 2021-02-26
Improve prefixing of Destruct by removing opened modules

voodoos:
(this PR targets branch 411)

This is an attempt to improve prefixing on Destruct and it will be reused by Construct.

We scan the environment to gather opened modules paths and then we use that information to drop prefixes in paths that are in the list of opened paths.

As suggested by @def, because over-shortening of the prefix can happen is some shadowing cases, we always check that the resulting lid can be found in the environment. If not we fallback on the non-shortened lid.

I made the function looking up in the environment a parameter of `Destruct.Path_utils.to_shortest_lid` in anticipation of Construct usages. For Destruct the lookup function if always `Env.find_constructor_by_name` but Construct will need more flexibility as it handles more diverse types of objects.

### DIFF
--- a/src/analysis/destruct.ml
+++ b/src/analysis/destruct.ml
@@ -45,8 +45,10 @@ let () =
   )
 
 module Path_utils : sig
-  val to_shortest_lid : env:Env.t -> Path.t -> Longident.t
-  val replace_name_by : name:string -> Longident.t -> Longident.t
+  val to_shortest_lid :
+    env:Env.t ->
+    ?name:string ->
+    env_check:(Longident.t -> Env.t -> 'a) -> Path.t -> Longident.t
 end = struct
   let opens env =
     let rec aux acc = function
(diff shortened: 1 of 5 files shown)


src/analysis/destruct.ml
```diff
@@ -44,6 +44,47 @@ let () =
...
+    ?name:string ->
+    env_check:(Longident.t -> Env.t -> 'a) -> Path.t -> Longident.t
+end = struct
+  let opens env =
+    let rec aux acc = function
+    | Env.Env_open (s, path) -> aux (path::acc) s
+    | s -> Option.(
+        map ~f:(aux acc) (Browse_misc.summary_prev s)
+        |> value ~default:acc
+      )
```
trefis, line 59: I'm not OK with this non indentation

trefis: Also, which `map` is that?

tests/test-dirs/destruct/dune
```diff
@@ -0,0 +1,3 @@
+(cram
+ (applies_to :whole_subtree)
+ (alias all-destruct-tests))
```
trefis, line 3: why do we need this here, but for not in any of the other test dirs?

tests/test-dirs/destruct/prefixing.t
```diff
@@ -0,0 +1,108 @@
...
+### FROM EXPR ###
+#################
+Test 1.1
+  $ $MERLIN single case-analysis -start 3:2 -end 3:3 -filename typ.ml <<EOF | \
+  > tr -d '\n' | jq '.value[1]'
+  > module A = struct type my_list = Atom | Elt of string * my_list end
+  > let f x : A.my_list =
+  >   x
+  > EOF
+  "match (x : A.my_list) with | A.Atom -> (??) | A.Elt (_, _) -> (??)"
```
trefis, line 11: how did the type annotation on `x` appear?

I know that's unrelated by the content of the PR, but I'm somewhat baffled (also, isn't it incorrect?)

src/analysis/destruct.ml
voodoos: It is Option's map of course 😅 
I will change that.

tests/test-dirs/destruct/dune
voodoos: We do not need it. It is just a convenient alias to run (and watch) all Destruct tests together.

src/analysis/destruct.ml
trefis: Ah! I had missed that local open. Yes, please, sorry for the verbosity/repetition.

tests/test-dirs/destruct/prefixing.t
voodoos: I have no idea, but it is valid.

tests/test-dirs/destruct/dune
trefis: And just to confirm, it won't conflict in some way with the similar stanza in `tests/dune`? (And why don't we need to specify the test deps here?)

tests/test-dirs/destruct/prefixing.t
trefis: Is it? Nothing in the source indicates that `x` has that type, that type is the return type of the function.

tests/test-dirs/destruct/dune
voodoos: In my understanding it adds with previous stanzas, not redefine them. But you make me doubt, I will ask Rudi.

tests/test-dirs/destruct/prefixing.t
voodoos: But the body of the function is the single expression `x` ?

voodoos: So the return type of the function is the same as the type of `x`.

trefis: Right, I guess that's fair.
Still surprised that an annotation is added, but that's a mystery for another time.

PR #1279, opened 2021-03-08
Constify argv parameter in _spawnvp

MisterDA:
This suppresses a warning in mingw-w64.


voodoos:
This looks fairly innocuous, however I guess it would be best if @let-def could check that change.

voodoos:
(CI should be green after rebasing)

PR #1283, opened 2021-03-15
CI fix attempt

voodoos:
Ocaml CI was running one 4.11 build because the constraint was `>4.11.1` and not `4.11.2`.
This PR sets the constraint to `>= 4.12.0`.


voodoos:
Great success. 4.11 build disappeared from the CI, and all failures are expected.

PR #1284, opened 2021-03-15
Fix workdir detection

voodoos:
The `workdir` configuration value used to be the path to the folder containing the closest `.merlin` file.

When moving to `dot-merlin-reader` we start the reader in the folder containing the closest `.merlin` file and the `workdir` is the starting directory of the reader. This mean that using `dot-merlin-reader` did not change the `workdir` value.

However the `dune ocaml-merlin` configuration reader is not started from the closest `dune` file (the equivalent of the closest `.merlin` file) but from the closest `dune-project` or `dune-workspace` file. And because we use that starting path as the value for `workdir` as with `dot-merlin-reader` the `workdir` might wrongly be set to a parent of the "real workdir".

This could broke some projects using preprocessing with errors such as:
```
I/O error: memoize_tests.ml: No such file or directory
```

This PR ensures the workdir is always pointing to the closest `dune` (or `.merlin`)  file. While scanning the folders for a `.merlin` or `dune-project` file we simply record the folder of the first `dune` file we found.


voodoos:
(@jeremiedimino who alerted about that issue tested this PR succesfuly)

trefis (approved):

akaday:
const fs = require('fs');
const path = require('path');

function findClosestConfigFile(startDir, filenames) {
  let currentDir = startDir;

  while (currentDir !== path.parse(currentDir).root) {
    for (const filename of filenames) {
      const filePath = path.join(currentDir, filename);
      if (fs.existsSync(filePath)) {
        return currentDir;
      }
    }
    currentDir = path.dirname(currentDir);
  }

  return null;
}

const startDir = process.cwd();
const configFiles = ['dune', '.merlin'];
const workdir = findClosestConfigFile(startDir, configFiles);

if (workdir) {
  console.log(`Workdir set to: ${workdir}`);
} else {
  console.error('No configuration file found.');
}

PR #1286, opened 2021-03-17
Signature help

rgrinberg:
This series of patches were applied in LSP to provide signature hinting when a user types a function application. It would be much more convenient for us if they were upstreamed.

cc @mnxn

### DIFF
--- a/src/analysis/signature_help.ml
+++ b/src/analysis/signature_help.ml
@@ -146,13 +146,13 @@ let application_signature ~prefix = function
       | Some _ as ap -> ap
       | None -> active_parameter_by_prefix ~prefix result.parameters
     in
-    `Application { result with active_param }
+    Some { result with active_param }
 
   (* provide signature information directly after an unapplied function-type
      value *)
   | (_, Expression ({ exp_type = { desc = Tarrow _; _ }; _ } as e)) :: _ ->
     let result = separate_function_signature e ~args:[] in
     let active_param = active_parameter_by_prefix ~prefix result.parameters in
-    `Application { result with active_param }
+    Some { result with active_param }
 
-  | _ -> `Unknown
+  | _ -> None
--- a/src/analysis/signature_help.ml
+++ b/src/analysis/signature_help.ml
@@ -1,0 +1,138 @@
+open Std
+
+open Browse_raw
+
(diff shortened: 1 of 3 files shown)


src/analysis/signature_help.ml
```diff
@@ -0,0 +1,158 @@
...
+
+type application_signature =
+  { function_name : string option
+  ; function_position : Msource.position
+  ; signature : string
+  ; parameters : parameter_info list
+  ; active_param : int option
+  }
+
+(* extract a properly properly parenthesized identifier
```
mnxn, line 22:
```suggestion
(* extract a properly parenthesized identifier
```
Typo. oops.

voodoos (commented):
I started reading the changes, and I am wondering why you chose to return both the function signature as a string and the arguments offsets in that string. Wouldn't it be more simple to simply return the list of the arguments and then on the editor side you just print each argument with the correct style ?

mnxn:
The function needs to build the string in order to get the parameter offsets.

There's some complexity in the printing (like unwrapping option types), that I wasn't sure was appropriate to do from the editor side.

voodoos:
@mnxn before thinking about how we could improve and merge that PR I would like it's expected behaviour to be illustrated in a series of tests. Could you add some to merlin test-suite ?

The first cases that I can think about are:
- a simple function with test for the first arg, second and last
- a function with labelled argument that is then called with varying arguments order
- a function with an optional argument
- function calls before or after an error in the buffer 

I noticed some weird behaviour while using vscode, I while try to take notes next time it happens to write more specific tests.

Thank you !

mnxn:
My understanding is that to write a test like that example I would first need to expose the signature help functionality to the command line interface through the [`all_commands` list](https://github.com/ocaml/merlin/blob/8091fad9953c6b9840a53aa53336fdebe51658b6/src/frontend/ocamlmerlin/new/new_commands.ml#L69).

Is that correct?

voodoos:
Yes it is correct, and that also requires for a new Merlin feature: we eventually want it to be available for every Merlin's client and that means adding it to the protocol.

voodoos:
Closing for now since this has been re-implemented downstream.

PR #1287, opened 2021-03-19
[deconstruct][vim] Add next/previous hole commands

voodoos:
This PR add two new commands to the vim plugin : `MerlinNextHole` et `MerlinPreviousHole`
It also automatically jump to the first hole when destruct is called. This should allow nice chaining with Construct.

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -321,6 +321,16 @@ def command_occurrences(pos):
     except MerlinExc as e:
         try_print_error(e)
 
+def command_holes():
+    try:
+        lst_or_err = command("holes")
+        if not isinstance(lst_or_err, list):
+            print(lst_or_err)
+        else:
+            return lst_or_err
+    except MerlinExc as e:
+        try_print_error(e)
+
 ######## VIM FRONTEND
 
 def vim_complete_prepare(str):
@@ -557,7 +567,7 @@ def vim_type_reset():
     enclosing_types = [] # reset
     current_enclosing = -1
 
-def replace_buffer_portion(start, end, txt):
+def replace_buffer_portion_and_jump_to_hole(start, end, txt):
     (encode,decode) = vim_codec()
 
     start_line = start['line'] - 1
@@ -581,7 +591,10 @@ def replace_buffer_portion(start, end, txt):
 
     # Properly reindent the modified lines
     vim.current.window.cursor = (start['line'], 0)
-    vim.command("call feedkeys('%d==', 'n')" % nb_lines)
+    vim.command('normal %d==' % nb_lines)
+
+    # We look for a hole to move the cursor to in the range we replaced
+    vim_next_hole(start_line, start_line + nb_lines)
 
 def vim_case_analysis():
     global enclosing_types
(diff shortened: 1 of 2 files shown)


vim/merlin/autoload/merlin.py
```diff
@@ -607,12 +622,36 @@ def vim_case_analysis():
...
+        break
+
+def vim_next_hole(min = 0, max = sys.maxint):
+    line, col = vim.current.window.cursor
+    lst = command_holes()
+    holes = map(lambda x: x['start'], lst)
+    for pos in holes:
+      hnum = pos['line']
+      hcol = pos['col']
+      if hnum >= min and (hnum, hcol) > (line, col) and hnum <= max:
```
voodoos, line 651: Because the list is sorted this could be optimized by breaking if the `max` has been passed. 
But I am not sure it is worth it.

vim/merlin/autoload/merlin.py
```diff
@@ -607,12 +622,36 @@ def vim_case_analysis():
...
+      if (hnum, hcol) < (line, col):
+        vim.current.window.cursor = (hnum, hcol)
+        break
+
+def vim_next_hole(min = 0, max = sys.maxint):
+    line, col = vim.current.window.cursor
+    lst = command_holes()
+    holes = map(lambda x: x['start'], lst)
+    for pos in holes:
+      hnum = pos['line']
```
trefis, line 649: perhaps name that `hline`?

vim/merlin/autoload/merlin.py
```diff
@@ -557,7 +567,7 @@ def vim_type_reset():
     enclosing_types = [] # reset
     current_enclosing = -1
 
-def replace_buffer_portion(start, end, txt):
+def replace_buffer_portion(start, end, txt, next_hole = False):
```
trefis, line 570: FTR. I think this is only ever called with `next_hole=true`

trefis (commented):
With the changes proposed here, a call to `:MerlinDestruct` now entails three different queries to merlin.
Perhaps we could update the protocol in a backwards compatible way (adding a new command, or optional flags to the existing one) to only make one call?

voodoos:
> With the changes proposed here, a call to `:MerlinDestruct` now entails three different queries to merlin.
> Perhaps we could update the protocol in a backwards compatible way (adding a new command, or optional flags to the existing one) to only make one call?

To clarify, the three commands are the following:
1. `destruct`
2. `holes` to get the list of holes
3. `type-enclosing` to get the type of the destination holes

IMO the combination that would make most sense would be to have a flag for `destruct` so that it also returns the list of holes in the generated code (and still call `type-enclosing` separately). It might be a bit delicate to compute the absolute location of these holes however.

On my (very) small examples I don't feel any latency however. Are scenarios with multiple successive calls to destruct frequent ?

trefis:
The call to `type-enclosing` I had in mind was the one prior to calling destructing, and I feel like that one might not be necessary, we just have to handle gracefully in destruct when the given location doesn't make sense.

> IMO the combination that would make most sense would be to have a flag for destruct so that it also returns the list of holes in the generated code (and still call type-enclosing separately). It might be a bit delicate to compute the absolute location of these holes however.

I had something like that in mind yes.
Perhaps destruct should just return relative positions?

voodoos:
I just rebased this PR to use the new type information from the `holes` command. (#1289)

PR #1293, opened 2021-03-29
Remove result dependency

rgrinberg:
* It's no longer necessary because merlin doesn't support 4.02.
* This requires us to drop (implicit_transitive_deps false) temporarily.
The reason is that Csexp is still using result.

### DIFF
--- a/dot-merlin-reader.opam
+++ b/dot-merlin-reader.opam
@@ -15,7 +15,6 @@ depends: [
   "yojson" {>= "1.6.0"}
   "ocamlfind" {>= "1.6.0"}
   "csexp" {>= "1.2.3"}
-  "result" {>= "1.5"}
 ]
 description:
   "Helper process: reads .merlin files and gives the normalized content to merlin"
--- a/dune-project
+++ b/dune-project
@@ -4,4 +4,3 @@
 
 (cram enable)
 (formatting disabled)
-(implicit_transitive_deps false)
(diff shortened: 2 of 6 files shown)


voodoos:
Maybe we should wait for a release of Csexp ?

rgrinberg:
> Maybe we should wait for a release of Csexp ?

Then we'll need to add a lower bound on Csexp. Is that ok with you?

trefis:
> The reason is that Csexp is still using result.

At first I thought "But isn't `Result.result` defined as equal to `result` on all the versions we care about?", and then I realized the answer was "yes, but you can't know that without looking at its cmi!".

I think I'd rather keep `(implicit_transitive_deps false)` and the dependency on `result` for now.
Is there any reason to get rid of it?

rgrinberg:
> I think I'd rather keep (implicit_transitive_deps false) and the dependency on result for now.
Is there any reason to get rid of it?

It makes `(implicit_transitive_deps false)` less usable for everyone downstream. For example, if I want to use `Csexp` and `(implicit_transitive_deps false)`, I need to be able to interpret the issue you pointed out and add `result` as a dependency. People find this non obvious.

We can keep `(implicit_transitive_deps false)` if we bump the lower bound on csexp.

trefis:
But, how does that concern *us*? We already have the right dependencies.

voodoos:
Result dependency was removed in #1441.

PR #1296, opened 2021-03-31
Clarify .merlin requirement in README

felipeochoa:
Per the recent announcement (https://tarides.com/blog/2021-01-26-recent-and-upcoming-changes-to-merlin), .merlin files are not needed when using Dune.


voodoos (approved):
Thank you !

PR #1298, opened 2021-04-02
Disable failing tests on Windows

emillon:
This ensures that `dune runtest` works on Windows: if Windows runners are added to ocaml-ci, this will not break the build, and failures can be fixed incrementally.

### DIFF
--- a/tests/test-dirs/config/check/dune
+++ b/tests/test-dirs/config/check/dune
@@ -1,0 +1,4 @@
+(cram
+ (applies_to check-config)
+ (enabled_if
+  (<> %{os_type} Win32)))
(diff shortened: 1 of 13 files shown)


emillon:
Remark about the PR itself: I included 3 solutions depending on how you prefer:

1. individual `(subdir)` stanzas in the test dune file - this can be useful if you want to generate these
2. nested `(subdir)` stanzas in the test dune file - a bit more DRY
3. tests disabled in their own file close to the test. I think it's better since filenames are less duplicated (files can be moved more easily) but that's more files

If you prefer 1 we could extend the dune support for `(applies_to)` so that `(cram (applies_to a/b))` is a shorthand for `(subdirs a (cram (applies_to b)))`.

voodoos:
I think I am more in favour of option 3.

PR #1301, opened 2021-04-05
Tentative fix for iedit.

Drup:
the iedit emacs mode is broken with the last version of iedit. This is a tentative fix. It seems to work like that but don't ask me why. :)


Drup:
Hmm, I think this is already fixed in #1247

voodoos:
> 
> 
> Hmm, I think this is already fixed in #1247

We just merged #1247. Can you confirm that it fixes iedit ?

voodoos:
@drup I am closing this. Please reopen if the issue persists.

PR #1304, opened 2021-04-06
Do not fail when destructing no-nodes or recovered nodes

### LINKED ISSUE #1300: Assertion failure when calling destruct inside a pattern
Calling `case-analysis` (destruct) on `x` causes an assertion failure 

```ocaml
type t =
  | A of int
  | B

let f = function
  | A x (* <<< here *)
  | B ->
    0
```

```
File "ocaml-lsp-server/vendor/merlin/src/analysis/destruct.ml", line 284, characters 49-55: Assertion failed (-32603).
[Error - 5:08:53 PM] Request textDocument/codeAction failed.
  Message: File "ocaml-lsp-server/vendor/merlin/src/analysis/destruct.ml", line 284, characters 49-55: Assertion failed
  Code: -32603 
```

I checked with `ocamlmerlin` request -- the same assertion failure is raised. 

I guess this isn't too bad given a pattern such as above is not valid OCaml, but I get it quite often these days working on a real codebase.

voodoos:
Fixes #1300 

This PR adds graceful error handling for two cases:
- Recovered AST (#1300)
- No nodes

Future work: #1300 could be solved in a more satisfying way by improving the recovery

### DIFF
--- a/src/frontend/query_commands.ml
+++ b/src/frontend/query_commands.ml
@@ -610,7 +610,7 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
     Destruct.log ~title:"nodes after" "%a"
       Logger.json (fun () -> `List (List.map nodes ~f:dump_node));
     begin match nodes with
-      | [] -> raise No_nodes
+      | [] -> raise Destruct.Nothing_to_do
       | (env,node) :: parents ->
         let source = Mpipeline.input_source pipeline in
         let config = Mpipeline.final_config pipeline in
--- a/tests/test-dirs/destruct/issue1300.t
+++ b/tests/test-dirs/destruct/issue1300.t
@@ -1,0 +1,52 @@
+From issue 1300:
+https://github.com/ocaml/merlin/issues/1300
+  $ $MERLIN single case-analysis -start 6:5 -end 6:5 -filename i1300.ml <<EOF
+  > type t =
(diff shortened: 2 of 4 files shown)


tests/test-dirs/destruct/issue1300.t
```diff
@@ -0,0 +1,52 @@
...
+  > | A of int
+  > | B of int
+  > 
+  > let f = function
+  > | A x (* <<< here *)
+  > | B -> 0
+  > EOF
+  {
+    "class": "error",
+    "value": "Destruct not allowed on recovered node",
```
trefis, line 14: This is definitely better than an assertion failure, but perhaps we could give a nicer error message?
Something like "Error: the node on which destruct was called is ill-typed"?

src/frontend/query_commands.ml
```diff
@@ -33,6 +33,10 @@ module Printtyp = Type_utils.Printtyp
 
 exception No_nodes
 
+let () = Location.register_error_of_exn (function
+  | No_nodes -> Some (Location.error "No nodes")
```
trefis, line 37: I should have noticed this in another PR, but perhaps instead of `No_nodes` we could reuse `Destruct.Nothing_to_do`?

trefis (approved):
Some cosmetic comments, looks good otherwise.

PR #1310, opened 2021-04-08
Update Emacs usage docs where applicable

purcell:
Follow-up to https://github.com/ocaml/merlin/pull/1247

### DIFF
--- a/emacs/merlin-ac.el
+++ b/emacs/merlin-ac.el
@@ -10,6 +10,14 @@
 ;; Package-Requires: ((emacs "25.1") (merlin "3") (auto-complete "1.5"))
 ;; URL: http://github.com/ocaml/merlin
 
+;;; Commentary:
+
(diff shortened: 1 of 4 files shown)


emacs/merlin-ac.el
```diff
@@ -153,7 +161,6 @@ wrong then recompute it."
     (add-to-list 'ac-sources 'merlin-ac-source)))
 
 (add-hook 'merlin-mode-hook #'merlin-ac--setup)
-(when merlin-mode (merlin-ac--setup))
```
voodoos, line 156: Non-expert here: Is that call "non-sensical" because the function will and should be called via the hook `(add-hook 'merlin-mode-hook #'merlin-ac--setup)` ?

voodoos (approved):
Look good to me, just naively curious about 5fecfaa

Issue #1311, opened 2021-04-08
refactor-open rewriting the whole record instead of a field label


ulugbekna:
`refactor-open` is rewriting the whole record instead of a record field label.

This 

```ocaml
module M = struct
  type r = { i : int }
end

open M

let r = { M.i = 1 }
```

is rewritten to 

```ocaml
module M = struct
  type r = { i : int }
end

open M

let r = i
```

This seems to happen because `Record_field` constructor in `node_paths` fn at `browse_raw.ml:864` contains location for the whole record.

PR #1312, opened 2021-04-08
refactor-open record field bug test

ulugbekna:
adds a test for #1311﻿


tests/test-dirs/refactor-open/record_field.t
```diff
@@ -0,0 +1,25 @@
+FIXME refactor open rewriting the whole record instead of a field label
+  $ $MERLIN single refactor-open -action unqualify -position 4:7 ./ocaml-lsp-server/src/t.ml <<EOF
```
trefis, line 2: Is the path here supposed to be `-filename <path>`?
If yes, let's just add the filename, and also, let's just make the path `test.ml`, because why not.

PR #1313, opened 2021-04-09
refactor-open qualify uses short-paths

ulugbekna:
`refactor-open qualify` uses a "full path" to qualify modules, e.g., 

```ocaml
open Import (* calling refactor open qualify on this open does the following qualification, which is undesirable *)

let path = Dune__exe.Import.Option.value_exn (Dune__exe.Import.Unix_env.get Dune__exe.Import.Unix_env.initial "PATH")
```

I think this is not what a user wants when they use this command. 

This PR makes the command to qualify paths up to the leftmost ident in the open expression, e.g., up to `Import` in the above example despite the full path being `Dune__exe.Import.Option.value_exn`. 

Notes: 
- based on #1312 branch
- added before- and after-PR tests
- the tests concern only refactor-open qualify because the changes concern only refactor-open qualify 
- made minimal changes to keep PR easy to review; otherwise, there were some bits that could be improved

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,13 @@ git version
     - add new module holes that can replace module expressions (#1333)
     - add a new command `construct` that builds a list of possible terms when
       called on a typed hole (#1318)
+    - `refactor-open qualify` improvements (#1313)
+      - do not make paths absolute, simply prefix with the identifier under the cursor
+        ```ocaml
+        open Foo (* calling refactor-open qualify on this open *)
+        let _ = Foo.bar (* previously could result in [Dune__exe.Foo.bar] *)
+        ```
+      - does not return identical (duplicate) edits
   + editor modes
     - vim: add a simple interface to the new `construct` command:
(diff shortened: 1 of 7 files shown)


trefis (approved):
LGTM.

Could you please rebase and add an entry to CHANGES.md? Then I'll merge.

ulugbekna:
@trefis Thanks for the review. I may have just found a bug with this implementation; let me test this slightly more, and I'll get back to it.

ulugbekna:
@trefis I fixed the bug I mentioned; it wasn't related to this PR, but was trivial enough to include here. See 2d5cc93 and 1cfa22d for the bug example and its fix. This should be ready. Thanks :-)

ulugbekna:
gentle ping @trefis if you have time

trefis (approved):
I'd rewrite the changelog entry to not mention short paths, but to say something like this instead:
> do not make paths absolute, simply prefix with the ident under the cursor

Otherwise this makes sense to me!

PR #1314, opened 2021-04-10
refactor-open: don't return useless edits

ulugbekna:
Before this PR `refactor-open` returns edits that when applied do not change the document, e.g, 

Calling `unqualify` on this:

```ocaml
open Unix 

let f = times () 
```

returned 

```ts
{
  "class": "return",
  "value": [
    {
      "start": {
        "line": 2,
        "col": 16
      },
      "end": {
        "line": 2,
        "col": 21
      },
      "content": "times"
    }
  ],
...
}
```

This is a hacky solution though, which doesn't work for multi-line identifiers (which are rare in OCaml, I guess). A proper solution would involve changing `Mbrowse.node_paths` to return `longident`s along with paths. I would appreciate input on which way to go. 

Notes:
- Based on #1313 
- I'm sending these PRs separately since their goals are not interconnected (the commits are but that can be fixed easily)

### DIFF
--- a/src/frontend/query_commands.ml
+++ b/src/frontend/query_commands.ml
@@ -815,24 +815,22 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
     let typer = Mpipeline.typer_result pipeline in
     let str = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
     let pos = Mpipeline.get_lexing_pos pipeline pos in
-    let tnode =
-      let should_ignore_tnode = function
+    let enclosing = Mbrowse.enclosing pos [str] in
+    let curr_node =
+      let is_wildcard_pat = function
         | Browse_raw.Pattern {pat_desc = Typedtree.Tpat_any; _} -> true
         | _ -> false
       in
-      let rec find = function
-        | [] -> Browse_tree.dummy
-        | (env, node)::rest ->
-          if should_ignore_tnode node
-          then find rest
-          else Browse_tree.of_node ~env node
-      in
-      find (Mbrowse.enclosing pos [str])
+      List.find_some enclosing ~f:(fun (_, node) -> 
+        (* it doesn't make sense to find occurrences of a wildcard pattern *)
+        not (is_wildcard_pat node))
+      |> Option.map ~f:(fun (env, node) -> Browse_tree.of_node ~env node)
+      |> Option.value ~default:Browse_tree.dummy
     in
     let str = Browse_tree.of_browse str in
     let get_loc {Location.txt = _; loc} = loc in
     let ident_occurrence () =
-      let paths = Browse_raw.node_paths tnode.Browse_tree.t_node in
+      let paths = Browse_raw.node_paths curr_node.Browse_tree.t_node in
       let under_cursor p = Location_aux.compare_pos pos (get_loc p) = 0 in
       Logger.log ~section:"occurrences" ~title:"Occurrences paths" "%a"
         Logger.json (fun () ->
@@ -855,13 +853,14 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
         let loc (_t,paths) = List.map ~f:get_loc paths in
         List.concat_map ~f:loc ts
 
-    and constructor_occurrence d =
-      let ts = Browse_tree.all_constructor_occurrences (tnode,d) str in
+    in
+    let constructor_occurrence d =
+      let ts = Browse_tree.all_constructor_occurrences (curr_node,d) str in
       List.map ~f:get_loc ts
 
     in
     let locs =
-      match Browse_raw.node_is_constructor tnode.Browse_tree.t_node with
+      match Browse_raw.node_is_constructor curr_node.Browse_tree.t_node with
       | Some d -> constructor_occurrence d.Location.txt
       | None -> ident_occurrence ()
     in
--- a/src/frontend/query_commands.ml
+++ b/src/frontend/query_commands.ml
@@ -504,11 +504,22 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
           in
           aux [] p |> String.concat ~sep:"."
         in
+        (* checks if the (un)qualified longident has a different length, i.e., has changed
+
+           XXX(Ulugbek): computes longident length using [loc_start] and [loc_end], hence
+           it doesn't work for multiline longidents because we can't compute their length *)
+        let same_longident new_lident { Location. loc_start; loc_end; _ } =
+          let old_longident_len = Lexing.column loc_end - Lexing.column loc_start in
+          loc_start.Lexing.pos_lnum = loc_end.Lexing.pos_lnum &&
+            String.length new_lident = old_longident_len
+        in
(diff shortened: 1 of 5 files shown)


rgrinberg:
Ping @trefis or @voodoos. We'd like these fixes before including the feature in lsp

trefis:
(I'll have a look at this next week)

src/frontend/query_commands.ml
```diff
@@ -481,30 +481,48 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+            | Pdot (path', s) ->
+              aux (s :: acc) path'
+            | _ -> raise Not_found
+          in
+          aux [] p |> String.concat ~sep:"."
+        in
+        (* checks if the (un)qualified longident has a different length, i.e., has changed
+
+           XXX(Ulugbek): computes longident length using [loc_start] and [loc_end], hence
+           it doesn't work for multiline longidents because we can't compute their length *)
```
trefis, line 510: Perhaps `qual_or_unqual_path` should just return the list, without concatenating the elements, and then you could just compare that to the result of `Longident.flatten`?

tests/test-dirs/refactor-open/qualify.t
```diff
@@ -0,0 +1,123 @@
...
+          "col": 9
+        },
+        "content": "M.N.u"
+      }
+    ],
+    "notifications": []
+  }
+
+Works for stdlib modules (stdlib modules differ from other in-file modules because their
+full path is different)
```
trefis, line 57: Not sure what you're trying to test, but `Unix` is not part of the stdlib.

tests/test-dirs/refactor-open/unqualify.t
```diff
@@ -0,0 +1,136 @@
...
+          "col": 13
+        },
+        "content": "u"
+      }
+    ],
+    "notifications": []
+  }
+
+Works for stdlib modules (stdlib modules differ from other in-file modules because their
+full path is different)
```
trefis, line 57: Not sure what you're trying to test, but Unix is not part of the stdlib.

(Repeated from above, because one of the two will probably disappear after a rebase?)

tests/test-dirs/refactor-open/unqualify.t
```diff
@@ -0,0 +1,136 @@
...
+          "line": 2,
+          "col": 22
+        },
+        "content": "times"
+      }
+    ],
+    "notifications": []
+  }
+
+FIXME shouldn't return anything, as nothing to unqualify
```
trefis, line 81: No need for the FIXME anymore.

tests/test-dirs/refactor-open/unqualify.t
```diff
@@ -0,0 +1,136 @@
...
+  > open Unix
+  > let times = times ()
+  > EOF
+  {
+    "class": "return",
+    "value": [],
+    "notifications": []
+  }
+
+FIXME shouldn't return anything, as nothing to unqualify for multiline paths
```
trefis, line 93: No need for the FIXME anymore.

src/frontend/query_commands.ml
ulugbekna: Let me describe how I see this: `Browse_tree.all_occurrences_of_prefix` returns `(Browse_raw.node * Path.t Location.loc list) list`. This means that we get a node `n` and the absolute path `p` associated with it, but we do not know the identifier `i` of the node `n`. However, we need to check whether `[qual_or_unqual_path] p = i` to be able to tell whether the identifier is updated and we should return that result. (AFAIU, `Ident.t` represents the identifier used in the source code, while `Path.t` is the absolute path, which can be different from that identifier, e.g., `foo` vs `Dune__exe.Import.foo`)

If I understand it correctly, we need to either get the ident either from node `n` or make sure that `Browse_tree.all_occurrences_of_prefix` returns also idents along with paths, which is quite an invasive change. Does this make sense? If yes, which solution would you prefer me to work on? 

---

The motivation behind this PR is to have smarter code actions (code change proposal based on the current context) 

For example, in the screenshot below, we get a code action `Remove module name from the identifiers` because there are meaningful changes possible to the code (remove `Unix` from `Unix.putenv`). Without this PR, this code action is proposed even if it doesn't change the code. 

image

trefis: Sorry, my message wasn't clear enough. But I think I understood what you're trying to achieve alright.
My point was just that, as you say, comparing string lengths is brittle. And you're going to run into trouble in more cases than just multi-line longidents. Anytime you have a blank space in the source is going to be problematic for you.

So what I was suggesting is to not count the raw/textual length, but the number of elements of the path.
Something like this:
```ocaml
let qual_or_unqual_path p =
  let rec aux acc (p : Path.t) =
    match p with
    | Pident ident ->
      Ident.name ident :: acc
    | Pdot (path', s) when mode = `Unqualify && Path.same path path' ->
      s :: acc
    | Pdot (path', s) when mode = `Qualify && s = leftmost_ident ->
      s :: acc
    | Pdot (path', s) ->
      aux (s :: acc) path'
    | _ -> raise Not_found
  in
  aux [] p
in
let same_path old_path new_path =
  let rec path_len = function
    | Pident _ -> 1
    | Pdot (p, _) -> 1 + path_len p
    | _ -> assert false
  in
  path_len old_path = List.length old_path
in
...
    match qual_or_unqual_path path with
    | new_path when same_path path new_path -> None
    | new_path -> Some (String.concat ~sep:"." new_path, loc)
    | exception Not_found -> None
...
```

What's wrong with that?

ulugbekna: Given ident = `foo`, old_path = `Import.foo`, we call unqualify `Import` on the old_path and get new_path = `foo`, right? `same_path old_path new_path` returns false and we return this as an edit, even though the user already has ident as `foo` and doesn't need this edit. Does this make sense? 

(Again, I'm under assumption that `Path.t` we get from `Browse_tree.all_occurrences_of_prefix` represents the absolute path for the node, which can be different from the ident used for that node)

trefis: Nevermind, I got confused. At first I thought you had the longident corresponding to the path, which would have been ideal. But you don't. Then I tried to adapt my solution to use the unedited path, but you're right that that one is always absolute, so that won't work either.

I think we can revisit this in the future, and that just playing with locations for now is fine.

trefis:
I'll let you rebase and update the changelog entry before giving it one last look (and merging).

ulugbekna:
@trefis this should be ready

tests/test-dirs/refactor-open/unqualify.t
trefis: Still not sure what you're trying to test. This test already behaves like that before the PR.

ulugbekna: I wanted to make sure unqualify works the same for the modules defined in the file where unqualify is called and for modules defined outside of that file. I put the test to show that the PR doesn't bring unintended changes 

If that doesn't make sense, I can remove the test.

trefis: Yes, I don't think that test makes sense.

PR #1318, opened 2021-04-16
[Construct] Part I 

voodoos:
This PR adds:

- a new command `construct` that given the position of a hole will return a list of possible expressions to fill this hole
- a handful of tests for that command
- the VIM plugin support for that feature:
  + `MerlinConstruct` on a hole `_` will open a completion dialog with the alternatives or replace the hole is only one construction is possible
  + One can ask to increase (or decrease) search depth with `c-i` (`c-u`) while choosing a completion
  + Once a completion is chosen it replaces the hole and the cursor jump to the next hole

Todo:
- [x] Polish the VIM plugin (for example we should only jump to holes in the constructed expression, not further)
- [x] Treat the various todo in `construct.ml`
- [x] Add documentation (and a changelog entry)
- [x] Maybe improve the test suite

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,14 @@ git version
     - fix location of module definitions done via functors (#1329, fixes #1199)
     - fix -cmt-path dirs mistakenly added to build path (#1330)
     - add new module holes that can replace module expressions (#1333)
+    - add a new command `construct` that builds a list of possible terms when
+      called on a typed hole (#1318)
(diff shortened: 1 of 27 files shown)


voodoos:
I think we can start the review, maybe ignoring the VIM plugin first. I will probable rewrite part of it has it became more and more messy while I added features.

trefis (approved):

PR #1319, opened 2021-04-16
Emacs: fix 2 erroneous calls to merlin-locate

jamesjer:
Emacs byte compile says:
```
In ac-merlin-locate:
merlin-ac.el:142:8:Warning: merlin-locate called with 1 argument, but accepts
    only 0

In merlin-company-backend:
merlin-company.el:112:23:Warning: merlin-locate called with 1 argument, but
    accepts only 0
```

I think these should be calls to `merlin-call-locate`; i.e., the former `merlin/locate`.

### DIFF
--- a/emacs/merlin-ac.el
+++ b/emacs/merlin-ac.el
@@ -139,7 +139,7 @@ wrong then recompute it."
     (when (popup-hidden-p ac-menu)
       (ac-show-menu))
     (let ((merlin-locate-in-new-window 'always))
-      (merlin-locate (ac-selected-candidate)))
+      (merlin-call-locate (ac-selected-candidate)))
     (ac-show-menu)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
--- a/emacs/merlin-company.el
+++ b/emacs/merlin-company.el
@@ -109,7 +109,7 @@
       (doc-buffer (merlin-company--doc-buffer arg))
       (location
        (ignore-errors
-         (let ((data (merlin-locate arg)))
+         (let ((data (merlin-call-locate arg)))
            (when (listp data)
              (let ((filename (merlin-lookup 'file data (buffer-file-name)))
                    (linum (cdr (assoc 'line (assoc 'pos data)))))


voodoos:
My Emacs setup is broken right now, so I cannot test, but I guess that means locate is broken on our latest releases ?

jamesjer:
Yes, the autocomplete and company integrations both try to complete the wrong symbol.

I should introduce myself.  I recently took over as maintainer of the merlin package for the Fedora Linux distribution.  I'm afraid that merlin (at least the Emacs part) is fairly broken in Fedora 33.  I managed to get most of the problems fixed for Fedora 34, which will be released next week, but a few problems remain.  I hope to get the remaining problems ironed out and have a very solid merlin presence in Fedora 35, due out in October/November.  I was an XEmacs developer for many years, so I've got a lot of experience with both the internals and the Lisp level of an Emacs-like system.

voodoos:
> I was an XEmacs developer for many years, so I've got a lot of experience with both the internals and the Lisp level of an Emacs-like system.

The editor plugins did not receive much love recently, your help is most welcome, thank you !

trefis (commented):
I'm not sure what calling locate from company/ac-mode is supposed to do, but with this PR it looks like it's just going to ignore the result of the call to merlin.

Don't you rather want to call `merlin-locate-ident`?

trefis:
OK, my last comment was apparently incorrect.
Or rather: you're just reproducing the previous behavior (before the recent breakage).

Well, I don't know what's going on, do whatever you guys think is best.
(Also, this feature is ridiculous anyway)

voodoos:
Also there is another call to `merlin/locate` that was wrongly replaced by `merlin-locate` in `merlin-xref.el` (line 25). Can you take care of this one too ?

voodoos:
> Yes, the autocomplete and company integrations both try to complete the wrong symbol.

Actually I don't think this is the case: completion should work but there appear to be a "feature" allowing the user, during the completion, to locate one of the suggested identifiers. And it is that feature which is broken (but arguably not very useful).

voodoos:
Closing in favor or #1349

PR #1324, opened 2021-04-20
Fix location of patterns in Occurrences

rgrinberg:
For a pattern like `(x)`, merlin will include the parens in the locations. This
PR fixes changes the location to only include the identifier.


rgrinberg:
This was originally reported in https://github.com/ocaml/ocaml-lsp/issues/375

rgrinberg:
This was originally reported in https://github.com/ocaml/ocaml-lsp/issues/375

trefis:
Thanks!

PR #1329, opened 2021-04-27
Attempt to fix #1199

### LINKED ISSUE #1199: Go-to-definition of a module defined using a functor leads to functor not module
Go-to-definition of a module defined using a functor shows the functor used to define the module instead of the place where the module was defined. 

Example:

```ocaml
1 module Make () = struct
2   let u = ()
3 end
4
5 module T = Make ();;
6 
7 let () = T.u
```

Go-to-definition on `T` in `T.u` in goes to the definition of `Make` (line 1) instead of `T` (on line 5). 

I think that's not the expected behavior.

Using the (latest) ocamllsp 1.2.0-5-g68a0237. (I think that ocaml-lsp relies on merlin for this feature, so reporting here)

rgrinberg:
Not confident this is correct, but it should be easy to evaluate since it's a tiny patch.

### DIFF
--- a/src/analysis/typedtrie.ml
+++ b/src/analysis/typedtrie.ml
@@ -677,15 +677,19 @@ let rec follow ~remember_loc ~state scopes ?before trie path =
               inspect_node state node
             end
           | Apply { funct; arg } ->
-            log ~title:"functor application" "";
-            let functor_argument = inspect_functor_arg (snd arg) in
-            let state =
-              { state with
-                functor_arguments = functor_argument :: state.functor_arguments
-              }
-            in
-            inspect_functor (Namespaced_path.peal_head_exn path) state
-              (snd funct)
+            let path = Namespaced_path.peal_head_exn path in
+            begin match Namespaced_path.head path with
+            | None -> Found (loc, doc)
+            | Some _ ->
+              log ~title:"functor application" "";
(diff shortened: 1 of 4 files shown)


src/analysis/typedtrie.ml
```diff
@@ -677,15 +677,19 @@ let rec follow ~remember_loc ~state scopes ?before trie path =
...
-              { state with
-                functor_arguments = functor_argument :: state.functor_arguments
-              }
-            in
-            inspect_functor (Namespaced_path.peal_head_exn path) state
-              (snd funct)
+            let path = Namespaced_path.peal_head_exn path in
+            if Namespaced_path.equal path Namespaced_path.empty then
+              Found (loc, doc)
+            else begin
```
trefis, line 683: Just for consistency with the cases above, I'd write this as:
```ocaml
let path = Namespaced_path.peal_head_exn path in
begin match Namespaced_path.head path with
| None -> Found (loc, doc)
| Some _ ->
```

trefis (approved):
That looks correct indeed!

src/analysis/typedtrie.ml
rgrinberg: done!

voodoos:
Thank you ! Can you add the changelog entry ?

rgrinberg:
Added!

PR #1331, opened 2021-04-28
Fix #1228

### LINKED ISSUE #1228: Go-to-def on an alias module leads to aliased module
Go-to-definition on an alias module `A` goes not to the definition of the alias module `A` but to the definition of the module being aliased `M`. Is this the expected behavior?

```ocaml
module  M = struct end
module A = M
module K = A (* go to def on `A` leads to `module M = ...` instead of `module A = ...` *)
```

Maybe related to #1199

rgrinberg:
Fix `locate` for module aliases


tests/test-dirs/locate/context-detection/cd-mod_constr.t/run.t
```diff
@@ -3,9 +3,9 @@
   {
     "class": "return",
     "value": {
-      "file": "lib/ocaml/string.ml",
+      "file": "lib/ocaml/stdlib.ml",
```
voodoos, line 6: This looks like a less useful result than before, escpecially since the user won't be able to jump to string.ml from stdlib.ml

ulugbekna: Interesting. Maybe we could also take into account doc comment `(** inline *)`.

rgrinberg: Yes, that's a shame. Another idea would be for merlin to transparently go through transparent aliases like `module Foo = Foo`. Tbh, I'm not too excited by this since people do want to jump to the alias.  Note that this isn't an issue for lsp. We have an implementation request that we could use to climb through the aliases.

voodoos:
With the upcoming shape-rewrite of locate (#1431) this issue is fixed (with correct behaviour for jumps to the stdlib too) but only for OCaml >= 4.14.

I think this PR would do more harm than good if we don't restore the expected behaviour when jumping to Stdlib's modules. It might also cause issue with Dune's generated wrappers that should be transparently traversed.

We could try to have special handling for these cases but I am not sure it is worth it since we already have a working fix for 4.14.

PR #1333, opened 2021-04-29
[construct] Module holes

voodoos:
This PR add support for module holes in the parser and the typer.
It also modifies the recovery to use module holes instead on empty structure by default.


src/analysis/typedtrie.ml
```diff
@@ -380,6 +384,7 @@ let rec build ~local_buffer ~trie browses : t =
           | `Sg  sg ->
             let sg = lazy (build ~local_buffer ~trie [of_signature sg]) in
             f (Included (Items sg))
+          | `Hole -> assert false
```
trefis, line 387: Is that right? I haven't really thought about it, but can't I `include _`? If indeed I can't, perhaps put a comment in there explaining why the assert is correct?

src/ocaml/preprocess/parser_raw.mly
```diff
@@ -1341,6 +1341,9 @@ module_expr [@recovery default_module_expr ()]:
     | (* An extension. *)
       ex = extension
         { Pmod_extension ex }
+    | (* A hole. *)
+      ex = UNDERSCORE
```
trefis, line 1345: `ex =` is perhaps excessive?

src/ocaml/typing/oprint.ml
```diff
@@ -541,6 +541,7 @@ and print_simple_out_module_type ppf =
   | Omty_alias id -> fprintf ppf "(module %a)" print_ident id
   | Omty_functor _ as non_simple ->
      fprintf ppf "(%a)" print_out_module_type non_simple
+  | Omty_hole -> fprintf ppf "'S"
```
trefis, line 544: Is that supposed to represent a signature variable?

src/ocaml/typing/typemod.ml
```diff
@@ -1931,7 +1934,7 @@ and package_constraints env loc mty constrs =
     match Mtype.scrape env mty with
     | Mty_signature sg ->
         Mty_signature (package_constraints_sig env loc sg constrs)
-    | Mty_functor _ | Mty_alias _ -> assert false
+    | Mty_functor _ | Mty_alias _ | Mty_for_hole -> assert false
```
trefis, line 1937: Can you justify why this case can't happen?

trefis (commented):
I have a couple of questions, but this mostly looks fine :)

src/ocaml/typing/oprint.ml
voodoos: Yes, the intention is to mean "Any signature". Do you have a better solution ? Maybe an explicit name like "Any_sig" ?
Note that `sig end` carries a different meaning. Maybe `sig _ end` ?

src/ocaml/typing/typemod.ml
voodoos: I cannot, but wasn't able to trigger this case.
However it is probably safer to just return `Mty_for_hole` in that case.

tests/test-dirs/type-enclosing/hole.t
```diff
@@ -33,3 +33,162 @@ Check that we can access the expected type of a hole:
...
+          "col": 25
+        },
+        "type": "(module Hole)",
+        "tail": "no"
+      }
+    ],
+    "notifications": []
+  }
+
+Type when no type avalable ?
```
trefis, line 161: available*

Issue #1335, opened 2021-04-30
Stack overflow when verbosity = 1


rgrinberg:
Consider this snippet:

```ocaml
type 'k t = 'k baz
and 'k baz =
  | Foo
  | Bar of 'k t
```

Now get the type enclosing of `'k t` on the 4th line. You will see a stackoverflow:

```
  Raised by primitive operation at Ctype.expand_abbrev in file \"src/ocaml/typing/ctype.ml\" (inlined), line 1655, characters 2-57
  Called from Ctype.try_expand_once in file \"src/ocaml/typing/ctype.ml\", line 1675, characters 22-44
  Called from Ctype.try_expand_safe in file \"src/ocaml/typing/ctype.ml\", line 1681, characters 6-28
  Called from Ctype.try_expand_head in file \"src/ocaml/typing/ctype.ml\", line 1687, characters 12-27
  Called from Ctype.try_expand_head in file \"src/ocaml/typing/ctype.ml\", line 1688, characters 6-38
  Called from Ctype.try_expand_head in file \"src/ocaml/typing/ctype.ml\", line 1688, characters 6-38
  Called from Ctype.try_expand_head in file \"src/ocaml/typing/ctype.ml\", line 1688, characters 6-38
```

rgrinberg:
PR #1336 reproduces

voodoos:
This look like a duplicate, or at least a similar issue than https://github.com/ocaml/merlin/issues/1216. 
@trefis did you look into it since ?

ulugbekna:
Btw, I also notice stackoverflows when hovering over recursive polymorphic variants in type definitions, which I think didn't happen before. I don't know whether it is better reported here or #1216. 

Example:

Hovering over `A in the screenshot below raises a stackoverflow.

![image](https://user-images.githubusercontent.com/16353531/116861695-fa19f900-ac1c-11eb-9cd8-ddf35d7b210b.png)

trefis:
> did you look into it since?

No.

I'd say here the main issue is that we're trying to get the definition of a type constructor from inside its definition. The result (on both examples) is nonsense with a verbosity of 0, and a stack overflow with a greater verbosity.
But these particular type declarations don't seem to be problematic as soon as the types are fully declared. So I suspect that detecting when we're looking at a type in the process of being declared and looking at his definition in an environment just after the declaration would make the problem disappear.

PR #1336, opened 2021-04-30
Reproduce issue #1335

rgrinberg:
type enclosing with verbosity=1


rgrinberg:
@voodoos I think this is fixed in master?

voodoos:
No it is not, but I would like to do something about it before the next release. As I remember it there are two sides of that issues:

- Merlin crashes  with a stack overflow when trying to print recursive type when `verbosity=1`. The best solution is to check if we are in a recursive type before entering the printing loop. The quick solution is to catch the stack overflow but that might not be reliable.

- LSP does not support increasing verbosity so it is underwhelming to have it at `0` when asking for the type of a module. There is [a PR](https://github.com/ocaml/merlin/pull/1374) from @ulugbekna to implement a "`smart`" verbosity mode.

voodoos:
I opened https://github.com/ocaml/merlin/pull/1525 which include this test and a naive fix (that simply catches the SO and retry with decreased verbosity).

PR #1339, opened 2021-05-07
Fixes random documentation fetched on hover

ulugbekna:
Fixes the issue reported in https://github.com/ocaml/ocaml-lsp/issues/344 (the link contains the reported issue and some investigation work 🕵️ ), where documentation for the last hovered or completed file is shown for the current file. 

The reported issue:

> On hover, I sometimes get documentation for a completely different thing than the identifier I hover over. The "different thing" is usually something that comes from stdlib such as `>` operator or as in the screenshot, where I hover over `err_pos_range` and get correct type but not doc.
>
> image

Re: fix

The fix removes code that looked for documentation in "last visited" file (a "visit" to a file is made by completion or type-enclosing request, for example). Maybe I'm missing where this behavior is necessary, but so far doc-on-hover behavior works well for me, and no tests are failing.

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -831,13 +831,6 @@ let get_doc ~config ~env ~local_defs ~comments ~pos =
   | `Found (_, Some doc) ->
     `Found doc
   | `Found (loc, None) ->
-    let comments =
-      match File_switching.where_am_i () with
-      | None -> comments
-      | Some cmt_path ->
-        let {Cmt_cache. cmt_infos; _ } = Cmt_cache.read cmt_path in
-        cmt_infos.Cmt_format.cmt_comments
-    in
     log ~title:"get_doc" "%a" Logger.fmt (fun fmt ->
         Format.fprintf fmt "looking around %a inside: [\n"
           Location.print_loc !last_location;
--- a/tests/test-dirs/random-doc-issue.t
+++ b/tests/test-dirs/random-doc-issue.t
@@ -1,0 +1,78 @@
+Reproduce [ocaml-lsp #344](https://github.com/ocaml/ocaml-lsp/issues/344): a symbol's
+documentation is some other random documentation from a library that was last pulled for
+completion. See the issue for details. The space is necessary for the position of
+documentation to be fetch-able.
+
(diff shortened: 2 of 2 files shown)


rgrinberg:
Firstly, thanks for tackling this. It's quite an annoying issue, and our users would greatly benefit if it is addressed.

As for the fix itself, I'm not convinced this fixes the issue in all cases. As far as I understand, the bug is caused by a heuristic returning a false positive. What this PR does is change the heuristic based lookup to look inside a different list of comments - the one provided to the function rather than the one picked up through the search. What is to stop this from picking up false positives in other circumstances?

ulugbekna:
> As far as I understand, the bug is caused by a heuristic returning a false positive. What this PR does is change the heuristic based lookup to look inside a different list of comments - the one provided to the function rather than the one picked up through the search. What is to stop this from picking up false positives in other circumstances?

Hm, let me explain how I see this all, and you can tell me if I'm missing something. 

The bug is that when we request documentation for position `P` in file `F` but the symbol at position `P` doesn't have documentation, merlin pulls in comments from the last visited file `A` of some library and considers those comments as if they are in `F`, which I think is not what we want, so I removed the piece of code that's doing that. 

- The list of comments passed to `get_doc` as an argument `comments` is the list of comments for the file from which we call the "find document" request.
- If [this](https://github.com/ocaml/merlin/blob/a69ef83f2894c7e019c9b487b9646f7a63456471/src/analysis/locate.ml#L818-L829) piece of code fails to find documentation, which is of course the case because the symbol at the given position `P` doesn't have documentation, the following code (that I removed) appends comments of the last visited file `A` to the comments of the current file `F` and tries to associate position `P` with the positions of the new list of comments (the association happens in `Ocamldoc.associate_comment`.)

To sum up, I think what's happening is when we fail to find documentation for position `P` in file `F` from other libs, we take the last visited file's docs and consider them as if they are in `F`.

Something I don't understand is why that code was there in the first place -- maybe there are cases when we do want that behavior.

ulugbekna:
@trefis, your input on this PR would also be appreciated :-) 

This is one of the most frequently reported issues in ocaml-lsp. The most recent report was by craigfe

PR #1340, opened 2021-05-16
(vim) Split the ocaml filetype into several filetypes

undu:
Now ocamlinterface ocamllex and menhir types must be taken into account as well.

The filetypes are introduced by https://github.com/ocaml/vim-ocaml/pull/61

This is a draft PR to attract comments and suggestions on how to adapt the code to the change, Merlin loads under the new filetypes and syntastic  works as well, ctrlp and neomake code is untested.


copy:
> ctrlp and neomake code is untested

Perhaps it should be included in another PR, so that users of those plugins can test it.

Besides that, this looks like a pretty safe change, and I would like to gently nudge for this to get merged. It's required so that progress on https://github.com/ocaml/vim-ocaml/pull/61 can be made.

voodoos:
@undu thanks for that contribution, and sorry about the long answer delay.

Would you have some time to test or remove the "untested ctrlp and neomake code" ?

undu:
Sure, I'll remove the code later today

vim/merlin/dune
```diff
@@ -13,9 +13,12 @@
         (ftdetect/merlin.vim as vim/ftdetect/merlin.vim)
         (ftplugin/merlin.vim as vim/ftplugin/merlin.vim)
         (ftplugin/ocaml.vim as vim/ftplugin/ocaml.vim)
+        (ftplugin/ocamlinterface.vim as vim/ftplugin/ocamlinterface.vim)
+        (ftplugin/ocamllex.vim as vim/ftplugin/ocamllex.vim)
```
voodoos, line 17: Is it expected that there is no reference to the `https://github.com/ocaml/merlin/pull/1340/files#diff-9be022c3a5969aeabb7003ebd6ab488ed56021a64cbac51be234b491e572d781` here ?

voodoos (approved):
Thanks @undu 

Could you add a change entry ?

vim/merlin/dune
undu: Yes, that would need explicit support for the filetype, which last time I checked was missing. Has it been added?

From the commit:
> Menhir filetype is not included as there seem to be no plans to include it and even then it would take a big effort.

voodoos: No, there's no such support, but then why do you need the ``vim/merlin/ftplugin/menhir.vim` file ?
(sorry, not a vim user :-) )

voodoos: Right, I think I understand. The filetype is described by the name of the file in `vim/merlin/ftplugin` and we want to load Merlin in menhir files even if support is not good.

undu: That's right, I think it might be better to just remove the file that enables merlin for menhir. It doesn't make any sense to have it, but not install it because merlin doesn't support it.

undu: I've kept the reference to the menhir filetype in the commit, but not on the changelog

voodoos:
Thank you @undu !

Issue #599, opened 2016-11-08
dot-opam not found: cygwin + emacs-w32 + fdopen's opam


ttamttam:
When opening a .ml file in emacs, tuareg and merlin modes are loaded as expected.
But merlin seems not to find my `.merlin` file:
* it complains about unbound module until I "use a package" from the menu
* when I "dot-merlin check" from the menu, the answer is "No .merlin loaded"

I'm not sure how to solve this problem?

Thank you for any help.

I'm describing the installation steps below:

I installed cygwin and opam from [https://fdopen.github.io/opam-repository-mingw/](https://fdopen.github.io/opam-repository-mingw/). I choosed 32 bits flavour.

I installed tuareg, utop and merlin with opam:
```
opam install tuareg utop merlin
```

Since I'm using emacs-w32 package from cygwin, I downloaded
https://www.emacswiki.org/emacs/download/cygwin-mount.el and
https://www.emacswiki.org/emacs/download/windows-path.el. I put them
in `~/.emacs.d/` and added in `~/.emacs`:
```lisp
(require 'cygwin-mount "~/.emacs.d/cygwin-mount.el")
(require 'windows-path "~/.emacs.d/windows-path.el")
(windows-path-activate)
```
Here is the last part of my emacs init file:
```
(
 let ((opam-share
       (ignore-errors (car (process-lines "opam" "config" "var" "share")))))
  (when (and opam-share (file-directory-p opam-share))

    (add-to-list 'load-path (expand-file-name "emacs/site-lisp" opam-share))

    (load "tuareg-site-file")

    (autoload 'utop "utop" "Toplevel for OCaml" t)
    (setq utop-command "utop -emacs")
    (autoload 'utop-minor-mode "utop" "Minor mode for utop" t)
    (add-hook 'tuareg-mode-hook 'utop-minor-mode)

    (load "merlin")
    (autoload 'merlin-mode "merlin" nil t nil)
    (add-hook 'tuareg-mode-hook 'merlin-mode t)
    (setq merlin-command  'opam)

    )
  )
```

let-def:
Could you share the directory setup? Is the `.merlin` is the same direcotry as ML files or a parent directory?

ttamttam:
It is in the same directory as the ML files.
Here is an example:
```bash
$ ls -al
total 6
drwxr-xr-x+ 1 vagrant        None  0 Nov  8 16:04 .
[...]

PR #1345, opened 2021-06-05
Replace spaces in socketname with '_'

ttamttam:
On windows, if the username retrieved by `GetUserName` contain spaces, they end up in the socketname, and emacs fails when trying to open it.

### DIFF
--- a/src/frontend/ocamlmerlin/ocamlmerlin.c
+++ b/src/frontend/ocamlmerlin/ocamlmerlin.c
@@ -537,6 +537,11 @@ static void compute_socketname(char socketname[PATHSZ], struct stat *st)
 
   if (!GetUserName(user, &dwBufSize))
     user[0] = '\0';
+   
+  for (int i = 0; user[i] != '\0' && i <= UNLEN; i++)
+     if (isspace(user[i]))
+        user[i] = '_';
+   
   // @@DRA Need to use Windows API functions to get meaningful values for st_dev and st_ino
   snprintf(eventname, PATHSZ,
       "ocamlmerlin_%s_%lx_%llx",
--- a/src/frontend/ocamlmerlin/ocamlmerlin.c
+++ b/src/frontend/ocamlmerlin/ocamlmerlin.c
@@ -9,6 +9,7 @@
 #include <windows.h>
 #include <Lmcons.h>
 #include <process.h>
+#include <sddl.h>       // ConvertSidToStringSid
 #ifndef STDIN_FILENO
 #define STDIN_FILENO 0
 #endif
@@ -521,35 +522,74 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
 }
 
 #ifdef _WIN32
+
+/* May return NULL */
+LPSTR retrieve_user_sid_string()
+{
+  LPSTR usidstr;
+  HANDLE process_token;
+  if ( ! OpenProcessToken( GetCurrentProcess(), TOKEN_QUERY, &process_token ) )
+    return NULL;
+
+  DWORD sid_buffer_size;
+  if ( ! GetTokenInformation(process_token, TokenUser, NULL, 0, &sid_buffer_size ) &&
+        ( GetLastError() != ERROR_INSUFFICIENT_BUFFER ) )
+  {
+    CloseHandle(process_token);
+    return NULL;
+  }
+
+  TOKEN_USER * token_user_ptr = (PTOKEN_USER) malloc(sid_buffer_size);
+  if ( ! token_user_ptr )
+  {
+    CloseHandle( process_token);
+    return NULL;
+  }
+
+  if ( ! GetTokenInformation(process_token, TokenUser, token_user_ptr,
+                             sid_buffer_size, &sid_buffer_size))
+  {
+    free(token_user_ptr);
+    CloseHandle(process_token);
+    return NULL;
+  }
+
+  if (! ConvertSidToStringSid(token_user_ptr->User.Sid, &usidstr))
+    usidstr = NULL;
+
+  free(token_user_ptr);
+  CloseHandle(process_token);
+
+  return usidstr;
+}
+
 static void compute_socketname(char socketname[PATHSZ], char eventname[PATHSZ], const char merlin_path[PATHSZ])
 #else
 static void compute_socketname(char socketname[PATHSZ], struct stat *st)
 #endif
 {
 #ifdef _WIN32
-  CHAR user[UNLEN + 1];
-  DWORD dwBufSize = UNLEN;
   BY_HANDLE_FILE_INFORMATION info;
+  LPSTR user_sid_string;
   HANDLE hFile = CreateFile(merlin_path, FILE_READ_ATTRIBUTES, FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE, NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
   if (hFile == INVALID_HANDLE_VALUE || !GetFileInformationByHandle(hFile, &info))
     failwith_perror("stat (cannot find ocamlmerlin binary)");
   CloseHandle(hFile);
 
-  if (!GetUserName(user, &dwBufSize))
-    user[0] = '\0';
-   
-  for (int i = 0; user[i] != '\0' && i <= UNLEN; i++)
-     if (isspace(user[i]))
-        user[i] = '_';
-   
+  user_sid_string = retrieve_user_sid_string() ;
+  if (! user_sid_string)
+    user_sid_string = LocalAlloc(LPTR, 1);
+
   // @@DRA Need to use Windows API functions to get meaningful values for st_dev and st_ino
   snprintf(eventname, PATHSZ,
       "ocamlmerlin_%s_%lx_%llx",
-      user,
+      user_sid_string,
       info.dwVolumeSerialNumber,
       ((__int64)info.nFileIndexHigh) << 32 | ((__int64)info.nFileIndexLow));
   snprintf(socketname, PATHSZ,
       "\\\\.\\pipe\\%s", eventname);
+
+  LocalFree(user_sid_string);
 #else
   snprintf(socketname, PATHSZ,
       "ocamlmerlin_%llu_%llu_%llu.socket",
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ git version
       - handle multi-line paths
     - Handle `Persistent_env.Error` in `Typemod.initial_env` (#1355)
     - locate: reset global state from all entry points (#1364)
+    - Windows: replace user names by their SID in socketnames (#1345, @ttamttam)
   + editor modes
     - vim: add a simple interface to the new `construct` command:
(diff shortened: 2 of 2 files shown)


let-def:
The patch look good thanks. I am surprised by the emacs bit: the socketname should not leak to the editor, right?

ttamttam:
> The patch look good thanks. I am surprised by the emacs bit: the socketname should not leak to the editor, right?

A more detailed explanation of how I found this solution is here:

https://github.com/ocaml/merlin/issues/599#issuecomment-846427952

Roughly (I did not check), I would think that at some time, ocamlmerlin has to say to the editor which socket it should connect to?

I can no more test it, because I renamed my windows account since then (which was not easy…).

Best regards

let-def:
Indeed, now I remember. Thanks for the PR. The socketpath does not leak in normal emacs mode, it is only provided in the debug info. Emacs does not connect directly to the socket, the merlin wrapper handles both side (starting a server if there is none, connecting to it if there is one).
So the problem has to do with some Windows IPC that I does not understand, but I am glad the `_` fixes it, thanks!

let-def:
(Maybe @dra27 can tell us more about that?)

let-def (approved):

ttamttam:
replacing the USER by it’s SID.

Instead of removing spaces, I tried to follow @dra27  original plan  (see https://github.com/ocaml/merlin/issues/599#issuecomment-856175679).

dra27 (commented):
Some comments on the implementation as given, but I think it's better to avoid `LookupAccountName` and do [`GetCurrentProcessId`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocessid) -> [`OpenProcessToken`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocesstoken) -> [`GetTokenInformation`](https://docs.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-gettokeninformation) with the `TokenUser` class which yields a [`TOKEN_USER`](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_user) struct containing the SID which finally goes on to the code you already have for `ConvertSidToStringSid`!

ttamttam:
Thanks @dra27 . I’ll have a look. Despite how painful is C for me, nowadays!

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
 }
 
 #ifdef _WIN32
+
+/* May return NULL */
+char * retrieve_user_sid_string()
+{
+  char * usidstr = NULL;
+
+  void * process_token = NULL;
```
dra27, line 531:
```suggestion
  HANDLE process_token;
```
(don't second-guess Windows API types)

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
+  if ( ! OpenProcessToken( GetCurrentProcess(), TOKEN_QUERY, &process_token ) )
+    return usidstr;
```
dra27, line 533: Personally, I would explicitly `NULL` here, rather than subtly via a variable
```suggestion
    return NULL;
```

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
+
+  unsigned long sid_buffer_size = 0;
```
dra27, line 535:
```suggestion
  DWORD sid_buffer_size;
```

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
+  if ( ! GetTokenInformation(process_token, TokenUser, NULL, 0, &sid_buffer_size ) &&
+        ( GetLastError() != ERROR_INSUFFICIENT_BUFFER ) )
+  {
+    CloseHandle( process_token );
+    process_token = NULL;
+    return usidstr;
```
dra27, line 541:
```suggestion
    return NULL;
```
(this isn't C++ - the assignment of `NULL` should be eliminated by the optimiser anyway 🙂)

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
+  }
+
+  TOKEN_USER * token_user_ptr = (PTOKEN_USER) malloc(sid_buffer_size);
+  if ( ! token_user_ptr)
+  {
+    CloseHandle( process_token);
+    process_token = NULL;
+    return usidstr;
```
dra27, line 549:
```suggestion
    return NULL;
```
(as before)

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
+  }
+
+  if ( ! GetTokenInformation(process_token, TokenUser, token_user_ptr,
+                             sid_buffer_size, &sid_buffer_size))
+  {
+    free(token_user_ptr);
+    token_user_ptr = NULL;
+    CloseHandle( process_token );
+    process_token = NULL;
+    return usidstr;
```
dra27, line 559:
```suggestion
    CloseHandle( process_token );
    return NULL;
```

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -9,6 +9,7 @@
 #include <windows.h>
 #include <Lmcons.h>
 #include <process.h>
+#include <Sddl.h>       // ConvertSidToStringSid
```
dra27, line 12:
```suggestion
#include <sddl.h>       // ConvertSidToStringSid
```

It's very niche, but if ocamlmerlin is cross-compiled from Linux, the mingw-w64 version of this header is `sddl.h` not `Sddl.h`!

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
+  }
+
+  ConvertSidToStringSid(token_user_ptr->User.Sid, &usidstr);
+
+  free(token_user_ptr);
+  token_user_ptr = NULL;
+  CloseHandle(process_token);
+  process_token = NULL;
+
+  return usidstr;
```
dra27, line 569:
```suggestion
  if (!ConvertSidToStringSid(token_user_ptr->User.Sid, &usidstr)
    usidstr = NULL;

  free(token_user_ptr);
  CloseHandle(process_token);

  return usidstr;
```

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
 }
 
 #ifdef _WIN32
+
+/* May return NULL */
+char * retrieve_user_sid_string()
+{
+  char * usidstr = NULL;
```
dra27, line 529:
```suggestion
  LPSTR usidstr;
```

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
 }
 
 #ifdef _WIN32
+
+/* May return NULL */
+char * retrieve_user_sid_string()
```
dra27, line 527:
```suggestion
LPSTR retrieve_user_sid_string()
```

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
       "ocamlmerlin_%s_%lx_%llx",
-      user,
+      user_sid_string,
       info.dwVolumeSerialNumber,
       ((__int64)info.nFileIndexHigh) << 32 | ((__int64)info.nFileIndexLow));
   snprintf(socketname, PATHSZ,
       "\\\\.\\pipe\\%s", eventname);
+
+  LocalFree(user_sid_string);
+  user_sid_string = NULL;
```
dra27, line 602:
```suggestion
```

src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -521,30 +522,84 @@ static void compute_merlinpath(char merlin_path[PATHSZ], const char *argv0, stru
...
-  if (!GetUserName(user, &dwBufSize))
-    user[0] = '\0';
+  char * user_sid_string = retrieve_user_sid_string();
+
+  if (! user_sid_string)
+  {
+    user_sid_string = LocalAlloc(LPTR, 1);
+    user_sid_string[0] = '\0';
+  }
+
```
dra27, line 591: Note that `LPTR` explicitly zeros the memory, so there's no need to assign `\0` afterwards.

I don't know how much it matters these days, but there are still supported versions of MSVC which require C90 variables at the start of blocks. I'd put `LPSTR user_sid_string` below `#ifdef _WIN32` above and then:
```suggestion
  /* user_sid_string is either the sid, or `""` and needs `LocalFree`ing */
  user_sid_string = retrieve_user_sid_string() || LocalAlloc(LPTR, 1);
```

dra27 (commented):
The only important change is the (un-)capitalisation of `Sddl.h` - this looks correct to me, thanks!

The rest is C "style" points, which are very much "take or leave":

- Assignments of `NULL` after `free` makes the flow harder to follow (for me), especially when there's a clear `return` statement following (i.e. there's visibly no chance of use-after-free)
- I'd _definitely_ use Windows API types
- Personally, I think `return NULL;` is a much clearer statement of intent than `return usidstr;` where `usidstr` was set to `NULL` many control blocks and lines earlier.

I haven't actually compiled this - happy to be pinged to do that as and when it's confirmed ready to merge!

voodoos:
@ttamttam : I just enabled Windows CI on master. It should run once you rebase your PR. The CI builds the binaries and run a subset of the test suite.

ttamttam:
> @ttamttam : I just enabled Windows CI on master. It should run once you rebase your PR. The CI builds the binaries and run a subset of the test suite.

:+1:

ttamttam:
@dra27  @voodoos 
Thanks for the detailed review.
It seems the CI workflows were triggered, but need to be approved.

voodoos:
Thank you for your contribution @ttamttam !

PR #1350, opened 2021-06-15
Allow using a custom/home made dot reader

vsiles:
At Facebook, we are using both dune and Buck to build ocaml projects.
With the recent changes in dune's behavior w.r.t. .merlin files (no
longer generated in the source tree), we need a way to tweak merlin's
logic to suits our dev environment.

Merlin's current behavior is to look into the directory of the target ml
file for a .merlin file. If it's there, call `dot-merlin-reader`. If it
is not, tries to detect dune files, and call dune.

This commit adds a simple logic to override merlin's behavior: merlin
will first look for a `.merlin.custom` file. If it is there, it will
call `dot-merlin-custom-reader` on it, and stops there.

This reader will be provided by us (or any third party that want some
custom logic).
The name of the reader can be overriden by an environment value
`DOT_MERLIN_CUSTOM_READER`

### DIFF
--- a/src/kernel/mconfig_dot.ml
+++ b/src/kernel/mconfig_dot.ml
@@ -138,7 +138,8 @@ module Configurator = struct
         match cfg with
         | Dot_merlin_custom ->
           let prog = dot_merlin_custom in
-          prog, [| prog |]
+          let prog_name = Filename.basename prog in
+          prog, [| prog_name |]
         | Dot_merlin ->
           let prog = "dot-merlin-reader" in
           prog, [| prog |]
--- a/src/kernel/mconfig_dot.ml
+++ b/src/kernel/mconfig_dot.ml
@@ -100,17 +100,26 @@ let prepend_config ~dir:cwd (directives : directive list) config =
 
 module Configurator = struct
   type t =
+    | Dot_merlin_custom
     | Dot_merlin
     | Dune
 
   let of_string_opt = function
+    | ".merlin.custom" ->
+      Some Dot_merlin_custom
     | ".merlin" ->
       Some Dot_merlin
     | "dune-project" | "dune-workspace" ->
       Some Dune
     | _ -> None
 
+  let dot_merlin_custom =
+    try
+      Unix.getenv "DOT_MERLIN_CUSTOM_READER"
+    with Not_found -> "dot-merlin-custom-reader"
+
   let to_string = function
+    | Dot_merlin_custom -> dot_merlin_custom
     | Dot_merlin -> "dot-merlin-reader"
     | Dune -> "dune"
 
@@ -127,6 +136,9 @@ module Configurator = struct
     let start ~dir cfg =
       let prog, args =
         match cfg with
+        | Dot_merlin_custom ->
+          let prog = dot_merlin_custom in
+          prog, [| prog |]
         | Dot_merlin ->
           let prog = "dot-merlin-reader" in
           prog, [| prog |]
@@ -284,6 +296,8 @@ let get_config { workdir; process_dir; configurator } path_abs =
         "A problem occured with merlin external configuration reader. %s If \
          the problem persists, please file an issue on Merlin's tracker."
         (match configurator with
+        | Dot_merlin_custom ->
+          Printf.sprintf "Check that `%s` is installed." Configurator.dot_merlin_custom
         | Dot_merlin -> "Check that `dot-merlin-reader` is installed."
         | Dune -> "Check that `dune` is installed and up-to-date.")
       in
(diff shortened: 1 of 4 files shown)


vsiles:
Things I'd like to discuss:
- name of the new file, `.merlin.custom`
- name of the new reader `dot-merlin-custom-reader`
- how to override it using env variables (the current code is probably not suitable for that :D) Maybe keep the name but allow specifying the location of the binary rather than using `PATH` (we'd like not to depend on `PATH`)

voodoos:
I am curious: why can't you just name your custom reader `dot-merlin-reader` and add it to your path so that Merlin use it unknowingly ? And the configuration files could keep the name `.merlin`. (these should not be removed by dune cleans)

vsiles:
Buck is using a regular .merlin file (that we versioned), and is our "main" building system.
The idea would be to only have the dune files and .merlin files versioned, so most devs work out of the box.
For dune users, we would add a .merlin.custom file that would override the .merlin scan and starts dune.

We have some ideas of more complex logic, but that's the simplest scenario we want to support atm.

vsiles:
Also we'd like people to avoid fiddling with their PATH (it doesn't mix well with VScode & the like) so being able to point merlin to the right binary is quite important

src/kernel/mconfig_dot.ml
```diff
@@ -284,6 +296,8 @@ let get_config { workdir; process_dir; configurator } path_abs =
         "A problem occured with merlin external configuration reader. %s If \
          the problem persists, please file an issue on Merlin's tracker."
         (match configurator with
+        | Dot_merlin_custom ->
+          Printf.sprintf "Check that `%s` is installed." Configurator.dot_merlin_custom
```
voodoos, line 300:
```suggestion
          Printf.sprintf "Check that `%s` is installed."
            Configurator.dot_merlin_custom
```

src/kernel/mconfig_dot.ml
```diff
@@ -320,15 +338,15 @@ let find_project_context start_dir =
     try
       Some (
         List.find_map [
-            ".merlin"; "dune-project"; "dune-workspace"
+          ".merlin.custom"; ".merlin"; "dune-project"; "dune-workspace"
```
voodoos, line 341: I think we should move this to the end of the list because it is the least-expected case.

```suggestion
          ".merlin"; "dune-project"; "dune-workspace"; ".merlin.custom"
```

src/ocaml/preprocess/menhirLib.ml
```diff
@@ -3785,5 +3792,5 @@ module MakeEngineTable (T : TableFormat.TABLES) = struct
 end
 end
 module StaticVersion = struct
-let require_20201216 = ()
+let require_20210419 = ()
```
voodoos, line 3795: There is no need to update `MenhirLib` and `parser_raw` with this PR.

src/kernel/mconfig_dot.ml
vsiles: But we want it to be picked up first: if you move it at the end, it will pick up .merlin first so we can't override the logic (or I am mistaken ?)

src/ocaml/preprocess/menhirLib.ml
vsiles: Ok, I added all the modified files. I'll revert them.

src/kernel/mconfig_dot.ml
vsiles: Do you have an `ocamlformat` config file or something to enforce your style ?

src/kernel/mconfig_dot.ml
```diff
@@ -100,17 +100,26 @@ let prepend_config ~dir:cwd (directives : directive list) config =
...
   let of_string_opt = function
+    | ".merlin.custom" ->
+      Some Dot_merlin_custom
     | ".merlin" ->
       Some Dot_merlin
     | "dune-project" | "dune-workspace" ->
       Some Dune
     | _ -> None
 
+  let dot_merlin_custom =
```
vsiles, line 116: I think this part needs improvement. I'm not sure what's the best way:
- fix the binary name to `dot-merlin-custom-reader` and allow overriding the path where it is located
- allow overriding the binary name, but do stuff here to correctly pass argv[0]

src/kernel/mconfig_dot.ml
voodoos: You are right, let's keep this order.

src/kernel/mconfig_dot.ml
voodoos: We do not use ocamlformat (yet).
We don't have strict style guidelines but we do wrap lines at 80 chars.

src/kernel/mconfig_dot.ml
vsiles: I decided on using the full path, and correctly extracting argv[0] from it (see line 139)

vsiles:
Closing PR after discussing with @trefis and @voodoos . We found a better solution that doesn't involve changing merlin's behavior (based on a wrapper around merlin itself).

PR #1351, opened 2021-06-15
CI tweaks, do not merge

voodoos:


voodoos:
@let-def 

I fixed most of the trivial lint errors, but there is a few left:
```
merlin-ac.el:135:0: error: "ac-merlin-locate" doesn't start with package's prefix "merlin-ac".
merlin-cap.el:74:0: error: Aliases should start with the package's prefix "merlin-cap".
merlin-imenu.el:89:0: error: "merlin-use-merlin-imenu" doesn't start with package's prefix "merlin-imenu".
```

If I am not mistaken these concern functions which are publicly available, so I am not sure we can do something about that ?

I guess a simple option is to not fail when linting these specific files.

voodoos:
> @let-def
> 
> I fixed most of the trivial lint errors, but there is a few left:
> 
> ```
> merlin-ac.el:135:0: error: "ac-merlin-locate" doesn't start with package's prefix "merlin-ac".
> merlin-cap.el:74:0: error: Aliases should start with the package's prefix "merlin-cap".
> merlin-imenu.el:89:0: error: "merlin-use-merlin-imenu" doesn't start with package's prefix "merlin-imenu".
> ```
> 
> If I am not mistaken these concern functions which are publicly available, so I am not sure we can do something about that ?
> 
> I guess a simple option is to not fail when linting these specific files.

I eventually used `define-obsolete-function-alias` to mark old-named function obsolete and make the linter happy. But I am not sure we really want that...

PR #1352, opened 2021-06-15
[construct] Add basic emacs support

voodoos:
New command: `merlin-construct` 
If multiple choices the basic completion buffer is used. 
After substitution the cursor jumps to the first hole in the constructed expression.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1308,6 +1308,18 @@ strictly within, or nil if there is no such element."
             (merlin--goto-point start)
             (message "%s" typ)))))))
 
+(defun merlin--first-hole-between (pmin pmax)
+  "Jump to the first hole in the given range and prints its type"
+  (let* ((hole (merlin--first-hole (merlin--holes) pmin '>)))
+    (when hole
+      (let* ((start (merlin-lookup 'start hole))
+             (typ (merlin-lookup 'type hole))
+             (hole-point (merlin-make-point start)))
+        (if (<= hole-point pmax)
+          (progn
+            (merlin--goto-point start)
+            (message "%s" typ)))))))
+
 (defun merlin-next-hole ()
   "Jump to the next hole and print its type"
   (interactive)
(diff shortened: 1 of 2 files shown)


emacs/merlin.el
```diff
@@ -1358,6 +1370,45 @@ strictly within, or nil if there is no such element."
...
+    (if (= (length results) 1)
+      (progn
+        (completion--replace start stop (car results))
+        (merlin--first-hole-between start (+ start (length (car results)))))
+      (cl-labels ((replace-c (_b _e newtext)
+            (advice-remove 'completion--replace #'replace-c)
+            (completion--replace start stop newtext)
+            (merlin--first-hole-between start (+ start (length newtext)))))
+        (progn  
+          (advice-add 'completion--replace :override #'replace-c)
```
voodoos, line 1390: @gasche : @trefis told me that you were an Emacs user, it would be great if you could find some time to have a look at this function.

The idea is that we want to call Merlin's new `construct` command and then show the result list in a "completion at point" buffer. If the user clicks one of the possible construct the hole in the code is replaced by this term and the cursor jumps to the first hole in this term.

To do that I used an `advice` to override the standard `completion--replace` function with the new function `replace-c`.

However I remove this advice only if a completion has been chosen (line 1386), and not if the user cancel the action (because I do not know how to do it). 

We would expect that in that second case the standard completion at point of Merlin would be broken because of the left-over overridden function, but it is not ! This might be due to some sort of scoping of the advice that we are unaware of.

The question is: is it really safe to not remove the advice ? If not, how can I hook a "completion cancel" event to remove it always ?

gasche: I'm an Emacs user, but my knowledge of Emacs Lisp is *very* limited and I don't think I can easily answer your question. One person who may is @cpitclaudel?

ulugbekna:
@voodoos I wonder how you intend to handle the depth parameter for
construct. Do you have it fixed (if yes, at what value?) or can iteratively
increase depth, say, with a completion entry `more depth`?

voodoos:
In a first version I don't plan to use it. However the VIM plugin does: when a list on constructions is proposed to the user it is possible to ask for more/less depth with <c-i> <c-i>
Using one of these replaces the content of the completion list by new results of the required depth.

emacs/merlin.el
voodoos: Thanks to @mattiasdrp I know understand why my code "works". It is actually very simple:
If the user does not select a completion the advice is not magically  removed. In fact when asking for a normal Merlin completion it will ran... and have no effect because there is no hole in normal Merlin completions.

cpitclaudel: I haven't read the code in detail, but wouldn't setting `completion-list-insert-choice-function` work instead of using advice?

cpitclaudel: > However I really wonder how I could have find completion-list-insert-choice-function by myself ?

In this particular instance `M-x apropos-value completion--replace` would work; but more generally it's a bit hard.  I looked for `completion--replace` in the sources.

PR #1355, opened 2021-06-17
Handle `Persistent_env.Error` in `Typemod.initial_env`

antalsz:
We have some users who are nondeterministically getting the error

```
error in process sentinel: merlin-mode failure: Uncaught exception Persistent_env.Error in initial_env.open_module: Persistent_env.Error(_)
```

We believe this to be due to a race with writing out `.cmi` files during the build process, and think it can be fixed by handling this exception in `Typemod.initial_env` in the same way other relevant errors are handled.

This change should probably be added to the "Backport to 3.4" and "Backport to 411" projects, but as I'm not a member of this repository I don't seem to be able to do that.


trefis:
Thanks!

PR #1358, opened 2021-07-01
Use locks for server related tests :tada: (require Dune 2.9)

voodoos:
This should fix random failures of the CI.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,4 @@
-(lang dune 2.7)
+(lang dune 2.9)
 (name merlin)
 (using menhir 2.0)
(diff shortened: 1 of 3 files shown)


trefis (approved):
:partying_face:

voodoos:
Tests ran successfully twice in a row so either I am very lucky or this is fixed.
There is a failure in OcamlCI linting, probably due to a cache which has not yet been updated to Dune 2.9.

Also, do we mind enforcing a dependency to Dune 2.9 or not ? 
I guess not since you approved the PR...

PR #1359, opened 2021-07-02
`merlin-locate-type` editor support

voodoos:


tests/test-dirs/locate-type.t/run.t
```diff
@@ -1,3 +1,8 @@
+  $ ocamlc b.ml a.ml -bin-annot -o a.cmi
+  $ cat >.merlin <<EOF
+  > S .
+  > B .
+  > EOF
```
trefis, line 5: `.` is added to the source and build dirs by default. So this `.merlin` is useless (or could remain empty).

tests/test-dirs/locate-type.t/run.t
```diff
@@ -1,3 +1,8 @@
+  $ ocamlc b.ml a.ml -bin-annot -o a.cmi
```
trefis, line 1: That `-o a.cmi` looks wrong:
```
$ ocamlc b.ml a.ml -bin-annot -o a.cmi
$ file b.cmi
b.cmi: OCaml interface file (.cmi) (Version 029)
$ file a.cmi
a.cmi: a ...opam/4.12.0/bin/ocamlrun script executable (binary data)
```
Why don't you just replace it by `-c`?

trefis: And actually, looking at the rest of the test, you only need to compile `b.ml`, not `a.ml`.

trefis (approved):
LGTM modulo some minor changes with the added test.

PR #1365, opened 2021-07-13
[Construct] fix suggested values

voodoos:
Thank you @ulugbekna for trying construct and noticing these issues.

Even if editor modes for emacs and vim don't use that feature yet, construct is able to look in the local environment for values whose type or return type is compatible with the current hole type.

Well, at least that is what the head of marketing said, but in practice this was mostly broken, and this PR attempts to fix it.

@ulugbekna could you test is with your shiny ocaml-lsp setup ? Even if I am still not convinced that we should enable this by default in the current form (showing all matching values) I am very interested by your feedback !

### DIFF
--- a/doc/dev/PROTOCOL.md
+++ b/doc/dev/PROTOCOL.md
@@ -226,13 +226,14 @@ Entries is the list of possible completion. Each entry is made of:
 ### `construct -position <position> [ -with-values <none|local> -depth <int> ]`
 
     -position <position>      Position where construct should happen
-    -with-values <none|local> Use values from the environment (defaults to none)
+    -with-values <none|local> Use values from the environment 
+                              (experimental, defaults to none)
     -depth <int>              Depth of the search (defaults to 1)
 
 When the position determined by `-position` is a hole (`_`), this command
   returns a list of possible terms that could replace it given its type.
 When `-with-values` is set to local, values in the current environment will be
-  used in the constructed terms.
+  used in the constructed terms. This feature is still under development.
 
 ### `document -position <position> [ -identifier <string> ]`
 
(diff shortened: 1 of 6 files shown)


ulugbekna:
Hi. It seems to have gotten much nicer! Most of the completions have the desired type and are relevant.

image

Side note: I wonder whether it is possible to avoid having `Lsp__types` kind of prefixes, since `Range` is available in the environment (via `open Import`).

---

Here I remove some of the existing code and recreate it with `construct` easily: 

https://user-images.githubusercontent.com/16353531/125615001-f2c32fb9-97fb-4f5b-b08b-45bb617d67c5.mov

voodoos:
> Most of the completions have the desired type

Do you mean that some don't ?

> I wonder whether it is possible to avoid having Lsp__types kind of prefixes

If you can craft a small repro I will try to improve that.

> I wonder if the code below should account for prefix Stdlib__...

It is not clear to me when these prefixes are added. It's Dune wrapping the lib ?

voodoos:
> sorting completions by their type with lower priority for polymorphic return type.
That's a great idea

> I'll try to. I assume I need to create a dune library and and executable that uses that library to reproduce this in the tests.
My guess is that this is not related to dune's prefixing and a simple local sub module would show the same behaviour.

voodoos:
I will merge that PR but more work need to be done if we want to have a great user experience.

@ulugbekna if you still decide to use `-with-values local` in OCaml-LSP please state explicitely that this is experimental and suggestions are subject to multiple changes in the future !

ulugbekna:
@ulysse thanks for the heads up! We don't use local values for now, but I'm
thinking of adding a configuration option with which a user can turn on
that mode. Would such an option be helpful for you to test things, for
example?

voodoos:
Yes that would be very useful ! (and the option could mention the experimental status of that feature)

PR #1366, opened 2021-07-15
Refactor open fixes

trefis:
Ping @ulugbekna


trefis:
I also added some tests to show functor applications are sometimes not rewritten.
But fixing that is going to require a non trivial patch, so I'm leaving it as future work.

(PS: the ping was just "FYI").

PR #1367, opened 2021-07-15
Minor Emacs fixes

mattiase:
* remove an Emacs 28 deprecation warning
* better error display
* avoid no-op buffer changes in refactoring
*

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1535,8 +1535,8 @@ loading"
 (defun merlin-locate-type ()
   "Locate the type of the expression under point."
   (interactive)
-  (let ((result (merlin/call "locate-type"
-                             "-position" (merlin/unmake-point (point)))))
+  (let ((result (merlin-call "locate-type"
+                             "-position" (merlin-unmake-point (point)))))
     (unless result
       (error "Not found. (Check *Messages* for potential errors)"))
     (unless (listp result)
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -70,7 +70,8 @@
   "The function takes four arguments:
    - the path to the merlin binary
    - the name of the command
-   - the total time spent in the server (or -1 if that information is not available)
+   - the total time spent in the server (or -1 if that information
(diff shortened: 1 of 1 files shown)


mattiase:
More fixes added. Tell me if you prefer them all squashed into a single commit; I'm not sure what the project convention is. The Emacs 28 compilation and loading now pass without warnings.

emacs/merlin.el
```diff
@@ -1493,6 +1493,8 @@ loading"
                              "-look-for" merlin-locate-preference)))
     (unless result
       (error "Not found. (Check *Messages* for potential errors)"))
+    (when (stringp result)
+      (user-error "%s" result))
     (unless (listp result)
       (error "%S" result))
```
voodoos, line 1519: I am not sure why you added a case instead of simply replacing `(error ...` occurences by `(user-error ...` ?

emacs/merlin.el
```diff
@@ -1949,7 +1949,7 @@ Runs a merlin process in the background and perform queries on it.
 
 Short cuts:
 \\{merlin-mode-map}"
-  nil
```
voodoos, line 1952: Will this change remain compatible with older versions of Emacs ?
(sorry, I am not myself an Emacs user)

emacs/merlin.el
voodoos: I think it should be sufficient, it look like such "hard errors" are already thrown inside `merlin-call`.

emacs/merlin.el
mattiase: Yes. `:init-value` was introduced in Emacs 21, along with the other keywords that merlin.el already uses.
Positional arguments to `define-minor-mode` have been deprecated for years but it's only now in Emacs 28 that we get a warning about them.

emacs/merlin.el
mattiase: Right, patch now updated accordingly.

voodoos (approved):
Thank you !

PR #1368, opened 2021-07-16
A new implementation of locate-declaration using uids

voodoos:
All tests pass but the suite is probably lacking some cases...

#### Todo
- [x] There are one case disjunction that is still incomplete. [Here](https://github.com/voodoos/merlin/blob/b15704d013f1b7f9e8d24ab3418572b522c9db64/src/analysis/locate_with_uids.ml#L45).

#### However this PR appears to already fix some issues (involving functors):
In the test `locate/functors/f-generative.t` and the last test of `locate/decl-uid.t` locate was wrongly pointing to the implementation instead of the interface file. It is not the case anymore.

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -28,6 +28,14 @@
 
 open Std
 
+exception Internal_uid
+
+let () =
+  Location.register_error_of_exn (function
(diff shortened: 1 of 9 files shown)


voodoos:
Closing, shapes superseed this

PR #1372, opened 2021-07-23
Fix unqualify shouldnt qualify

ulugbekna:
`unqualify` previously could qualify, for example:

```
module M = struct
   module N = struct
     type t = Foo | Bar
   end
end

open M (* unqualify [M] previously qualified [Foo] below into [N.Foo], which isn't what user wants I think *)

let v : N.t = Foo
```

The PR also uses a faster way to check if the new lident is different from the old one. 

Should be easiest to review commit-by-commit.

### DIFF
--- a/src/analysis/refactor_open.ml
+++ b/src/analysis/refactor_open.ml
@@ -7,6 +7,8 @@ open Std
 
       `Qualify - returns [node_path] with its prefix equal to [open_lident]
 
+    Returns [None] if [node_lid] doesn't need changes.
+
     Note: by "prefix" we mean the leftmost consecutive part of a longident or a path. *)
 let qual_or_unqual_path mode ~open_lident ~open_path node_path node_lid =
   let open_lid_head = Longident.head open_lident in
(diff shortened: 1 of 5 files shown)


src/analysis/refactor_open.ml
```diff
@@ -1,36 +1,52 @@
 open Std
 
-let qual_or_unqual_path mode leftmost_ident path p =
+(** [leftmost_lident lid] returns the leftmost part of [lid], e.g., 
+    given [String.Map.empty], [String] is returned *)
+let rec leftmost_lident (lid : Longident.t) = 
+  match lid with
+  | Lident s -> s
+  | Ldot(lid', _) -> leftmost_lident lid'
+  | Lapply(_, _) -> raise @@ Invalid_argument "leftmost_lident: don't support Lapply"
```
trefis, line 9: I think this could be
- named `head` (it's the equivalent of `Path.head`)
- moved to the `Longident` module (and we'd upstream it later)
- `raise @@ Invalid_argument ...` is `invalid_arg ...`. 
  But here I think we can just `assert false` (like `Path.head`).

src/analysis/refactor_open.ml
```diff
@@ -30,16 +30,16 @@ let qual_or_unqual_path mode ~open_lident ~open_path node_path node_lid =
...
     | _ -> raise Not_found
   in
-  aux [] node_path
-
-let same_longident new_lident old_lident =
-  List.length new_lident = List.length (Longident.flatten old_lident)
+  let new_node_lid = make_new_node_lid [] node_path in
+  if String.equal node_leftmost_lident (List.hd new_node_lid) (* if new lid = old lid *)
+  then None
+  else Some (String.concat ~sep:"." new_node_lid)
```
trefis, line 42: Since we're now returning an option, could we rewrite this to
```suggestion
match make_new_node_lid [] node_path with
| new_node_lid when not (String.equal node_leftmost_lident (List.hd new_node_lid)) ->
  Some (String.concat ~sep:"." new_node_lid)
| _ | exception Not_found -> None
```
So this function doesn't leak an exception (and we can simplify its caller).

trefis (commented):
Two inline comments, plus some meta-comments:
- the test update should be in the same commit as the code update, not in the one updating the changelog
- I'm not a fan of "[foo]" in commit messages, I'd rather have "foo:" :grin:

ulugbekna:
> the test update should be in the same commit as the code update, not in the one updating the changelog

yes, this is definitely not intended

trefis:
I think the history is fine like this, I'm just going to merge :)

voodoos:
Don't forget the projects^^

PR #1374, opened 2021-07-27
Support verbosity=smart

ulugbekna:
Implements support for `verbosity=smart`.

Includes a couple of clean ups, I can remove them from this PR if necessary.

Also includes a fix for: The format for MERLIN_LOG env var is different in two places in merlin `new_merlin.ml` and `ocamlmerlin_server.ml` -- `MERLIN_LOG=`  `filename{,section}*` or `filename`


src/analysis/completion.ml
```diff
@@ -185,7 +185,9 @@ let make_candidate ~get_doc ~attrs ~exact ~prefix_path name ?loc ?path ty =
     | `Mod m   ->
       begin try
           if not exact then raise Exit;
-          let verbosity = !Type_utils.verbosity in
+          let verbosity = 
+            Mconfig.Verbosity.to_int !Type_utils.verbosity ~for_smart:1
```
ulugbekna, line 189: In most cases, I tried to keep in mind ocaml-lsp's use case for `verbosity=smart`, but I'm not sure what to put here.

ulugbekna:
Can I jump on the current reviews train? :D

voodoos:
> Can I jump on the current reviews train? :D

Sorry, it will have to be after the imminent release

ulugbekna:
> I am curious: is that already in-use in OCaml-lsp ?

Unfortunately not, but it should be

rgrinberg:
We can experiment with this patch in lsp first if you'd like, but it needs to be a lot smaller. For starters, get rid of all nonessential changes to ease the maintenance.

ulugbekna:
Indeed that’s the goal.

voodoos:
@ulugbekna could you move the clean-ups/refactoring parts of the PR to another one as you suggested ?

ulugbekna:
I hope to do so this week :-)

src/analysis/type_utils.ml
```diff
@@ -78,7 +81,7 @@ module Printtyp = struct
               Btype.iter_type_expr (iter (pred d)) ty0
           end
       in
-      iter !verbosity ty;
+      iter verbosity ty;
```
voodoos, line 84: Lets be conservative since we did not do the "remove the reference" refactoring.
```suggestion
      iter !verbosity ty;
```

src/analysis/type_utils.ml
```diff
@@ -169,7 +188,11 @@ let print_short_modtype verbosity env ppf md  =
     Printtyp.modtype env ppf md
 
 let print_type_with_decl ~verbosity env ppf typ =
-  if verbosity > 0 then
+  match verbosity with
+  | Verbosity.Smart | Lvl 0 -> 
+    Printtyp.type_scheme env ppf typ
+
+  | Lvl _ -> begin
```
voodoos, line 201:
```suggestion
  | Verbosity.Smart | Lvl 0 ->  Printtyp.type_scheme env ppf typ
  | Lvl _ -> begin
```

src/kernel/mconfig.ml
```diff
@@ -341,13 +366,14 @@ let merlin_flags = [
...
         let verbosity =
-          try int_of_string verbosity
-          with _ -> invalid_arg "argument should be an integer"
+          Verbosity.of_string verbosity        
         in
         {query with verbosity}),
-    "<integer> Verbosity determines the number of expansions of aliases in answers"
+    "\"smart\" | <integer> Verbosity determines the number of \
+      expansions of aliases in answers. \"smart\" automatically \
+      determines this number itself."
```
voodoos, line 376: In fact, `smart` has a precise behavior: it expands Module types and nothing else. The spec should say that instead of a blurry statement.

src/analysis/type_utils.ml
```diff
@@ -120,7 +120,13 @@ module Printtyp = struct
...
   let modtype env ppf mty =
     let smart ppf = function 
-      | Types.Mty_ident _ | Mty_alias _ -> verbose_modtype env ppf mty
+      | Types.Mty_ident _ | Mty_alias _ -> 
+        Format.pp_print_string ppf "(* Module type: *)\n";
+        modtype ppf mty;
+        Format.pp_print_newline ppf ();
+        Format.pp_print_newline ppf ();
+        Format.pp_print_string ppf "(* Inside module type: *)\n";
+        verbose_modtype env ppf mty
```
voodoos, line 129: I really dislike that, when Merlin is asked for a type, the `type` field of the answer should be  a type, not some formatted text with types in it.

An equation would be better: `(module M) = sig ...end` but even that seems a bit ankward.

The users asked for the type of a module, she doesn't need it to be repeated in the answer. IMO we should just drop that commit.

voodoos (changes requested):
Thanks a lot @ulugbekna the changes mostly look good to me :-)

ulugbekna:
Thanks for a prompt review! Addressed all comments

voodoos:
We also need to update the protocol documentation:
https://github.com/ocaml/merlin/blob/master/doc/dev/PROTOCOL.md#type-enclosing--position-position---expression-string----cursor-int----index-int-

ulugbekna:
but that document doesn't mention `-verbosity` at all?

ulugbekna:
In another PR perhaps?

PR #1375, opened 2021-07-31
Destruct record field

trefis:
Record field projections were a bit broken, so adding a bunch of tests, and some fixes.


voodoos:
#le-changelog-c-est-maintenant

PR #1383, opened 2021-09-02
Vim plugin: fix filename style when under cygwin

nchataing:
When under cygwin, the vim plugin tries to send a request to the merlin server with filename of the unix form `/path/to/file.ml` whereas it needs a filename of the form : `C:/cygwin/path/to/file.ml`. This PR fixes that.

The only method I stumbled upon to get the (windows) cygwin root path was to use `cygpath -m /`, but I don't know if it is a good way. Let me know if you know a better way to get it.

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -88,6 +88,9 @@ def concat_map(f, args):
 
 def current_context():
     filename = vim.eval("expand('%:p')")
+    if platform == "cygwin":
+        cygpath = os.popen('cygpath -m /').read().strip()
+        filename = cygpath + filename
     content = "\n".join(vim.current.buffer) + "\n"
     return (filename, content)
 


vim/merlin/autoload/merlin.py
```diff
@@ -88,6 +88,9 @@ def concat_map(f, args):
 
 def current_context():
     filename = vim.eval("expand('%:p')")
+    if platform == "cygwin":
+        cygpath = os.popen('cygpath -m /').read().strip()
+        filename = cygpath + filename
```
nojb, line 93: If you are calling `cygpath` you can pass it the full filename instead of concatenating: `os.popen('cygpath -m ' + filename)` should be enough.

vim/merlin/autoload/merlin.py
```diff
@@ -88,6 +88,9 @@ def concat_map(f, args):
 
 def current_context():
     filename = vim.eval("expand('%:p')")
+    if platform == "cygwin":
```
nojb, line 91: Are you sure this is the right check? The code could be running on Cygwin and being used with a native Windows port of `merlin` (your case) or using the Cygwin-compiled version of `merlin` (which would need to use usual Unix paths). At the moment I am not sure how best to differentiate between these two cases... One possibility could be to use `ldd` and check if the merlin executable depends on `cygwin1.dll`, but that rather ugly...

nchataing:
There are still other problem when using merlin in this setting, I am closing this PR and I will reopen it when I’ll have other bugs fixed...

PR #1396, opened 2021-10-28
[WIP] A new website for Merlin documentation

voodoos:
This PR targets the `gh-pages` branch and aim to provide a modern website with resources for new users of Merlin.
The website can be previewed here: https://voodoos.github.io/merlin

It does not pretend to be an exhaustive source of information. In my opinion it should stay focussed on common workflows with easy setup and a simple reference for the most useful editor functions.

These items could then be removed from the github wiki, but the website will still redirect users eager for more in-depth configuration details to this wiki.

### Todo
- [x] First pass on the VIM reference
- [ ] Decide what will be the final form of the website
- [ ] Change video encoding: chromium does not support h264...
- [ ] Allow scan from search engines via the robots.txt file 
- [x] Make sure the fine bactrian caml's hat is free of use

### Final form of the website
Multiple organisations can be considered:
1. The current layout with one page for each editor
2. Having one page by feature by editor
3. Having one page by feature with both editor commands in it
4. The current layout + an editor-agnostic overview of the nicest features
5. An ∞ of other options...

Personally I am in favour of 4 because:
- It is already ~~50%~~  80% done
- It stays simple as in "not a lot of pages", but navigation is still good with the fixed left menu
- In the future I would like to add a VSCode page. Feature support is very different from other editors so it may not fit well in a common "feature menu"
- A and 4 are incompatible with screen cast exhaustivity because of visual pollution. But I am not convinced that a lot of them are needed.

But of course all this is open to discussion


voodoos:
@trefis, @def, any input on this ?
I feel like it is already mergeable in the current state...

I am not sure about the emacs installation: what is the relation between `tuareg` and `caml-mode` ? Are they complementary or is `ocaml-mode` a replacement of tuareg ? And if so what is the prefered one today ?

PR #1401, opened 2021-11-04
Add support for `-alert` compiler flag

### LINKED ISSUE #1400: `-alert` not understood by Merlin
It seems Merlin does not understand the `-alert` flag to the compiler.

image

nojb:
Fixes #1400

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ unreleased
   + merlin binary
     - enable `occurences` to work when looking for locally abstract types
       (#1382)
+    - handle `-alert` compiler flag
 
 merlin 4.3.1
 ============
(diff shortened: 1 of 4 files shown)


nojb:
Friendly ping.

trefis (approved):
LGTM!

Would you mind adding a test to make sure we don't break this in the future (however unlikely).
It'd be even nicer if you could add the test as a preliminary commit, and updated its output in the later commits. But I understand that might be a bit tedious.

nojb:
> LGTM!
> 
> Would you mind adding a test to make sure we don't break this in the future (however unlikely). It'd be even nicer if you could add the test as a preliminary commit, and updated its output in the later commits. But I understand that might be a bit tedious.

Sure, no problem. I'll do that later today or tomorrow. Thanks for the review!

nojb:
> Would you mind adding a test to make sure we don't break this in the future (however unlikely).
> It'd be even nicer if you could add the test as a preliminary commit, and updated its output in the later commits.

This is now done.

PR #1409, opened 2021-11-30
test: flag `-alert` with `=` is not accepted by Merlin

voodoos:
#1401 introduced support for the `-alert` option of the compiler however it seems that the compiler accept the syntax `-alert=-deprecated` while Merlin doesn't. This PR add a test illustrating that.

The issue arised in dune: ocaml/dune#5245

@nojb Looking at #1401 I am not sure of what the issue is since it mostly reuses compiler's functions. But maybe Merlin in general rejects the `<option>=<arg>` syntax for all command line arguments (@trefis ?)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ git version
     - fix handlink of ppx's under Windows (#1413)
     - locate: look for original source files before looking for preprocessed
       files (#1219 by @ddickstein, fixes #894)
+    - handle `=` syntax in compiler flags (#1409)
 
 merlin 4.4
 ==========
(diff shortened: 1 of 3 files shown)


nojb:
> But maybe Merlin in general rejects the `<option>=<arg>` syntax for all command line arguments (@trefis ?)

My understanding is that indeed `merlin` does not know how to handle arguments separated by `=`.

voodoos:
@trefis I proposed a change to handle `name=value` flags.

voodoos:
ping @trefis ?

trefis (approved):

PR #1412, opened 2021-12-08
Windows: fix ppx command output redirection

MisterDA:
The code introduced in #1270 switched from using `system()` to
`CreateProcess()` in order to not show a window for each execution of a
ppx. However, it kept `1>&2` which is shell syntax, leading to errors.

With this patch, we revert to `Sys.command` and appropriate quoting for
Unix systems, but on Windows we wrap `CreateProcess` and use the same
file descriptor for stdout and stderr to have the same redirection as
with the `system()` call.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ git version
 
   + merlin binary
     - filter dups in source paths (#1218)
+    - Windows: fix ppx command output redirection (#????, follow-up to #1270)
 
 merlin 4.4
 ==========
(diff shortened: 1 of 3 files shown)


nojb (commented):
Maybe I'm missing something, but to me it looks like there is something wrong in the patch. The argument `fn_in` and `fn_out` represent the input and output files passed to the PPX on the command-line (temporary files containing AST data). The redirection of `CreateProcess` on the other hand should just be to redirect `stdout` to `stderr` to avoid messing up `merlin`'s protocol. `fn_in` and `fn_out` should not be used for the redirection.

MisterDA:
Oh right! My patch is completely wrong. I made a mix of everything in my
head...
I'll make a revision shortly! sorry about that.

PR #1413, opened 2021-12-09
Windows: correctly redirect stderr to stdout of called ppx 

MisterDA:
There's no need for quoting the files as we're not calling the ppx
through a shell.

Redirect the stderr of the child process to which of the parent (as
posix `system(3)`), and the stdout of the child to the stderr of the
child (as merlin's `ppx_commandline`).

Setting up the redirection in the parent process has the downside that
we must set the `bInheritHandles` flag of `CreateProcess`, which means
that if merlin leaks a handle (not setting it non-inheritable /
cloexec), the ppx process will have access to it.

cc @let-def @nojb

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ git version
   + merlin binary
     - filter dups in source paths (#1218)
     - improve load path performance (#1323)
+    - fix handlink of ppx's under Windows (#1413)
 
 merlin 4.4
 ==========
--- a/src/platform/platform_misc.c
+++ b/src/platform/platform_misc.c
@@ -3,6 +3,7 @@
 #define CAML_INTERNALS
 #include <caml/misc.h>
 #include <caml/osdeps.h>
+#include <caml/unixsupport.h>
 #endif
 
 #include <caml/mlvalues.h>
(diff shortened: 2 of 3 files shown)


src/platform/platform_misc.c
```diff
@@ -68,29 +69,51 @@ static int windows_system(const char *cmd)
...
+    p_stdout = GetStdHandle(STD_OUTPUT_HANDLE);
+    if (p_stdout == INVALID_HANDLE_VALUE) {
+        err = GetLastError(); goto ret1;
+    }
+    hp = GetCurrentProcess();
+    if (! DuplicateHandle(hp, p_stdout, hp, &(s_info.hStdOutput),
+                          0, TRUE, DUPLICATE_SAME_ACCESS)) {
+        err = GetLastError(); goto ret1;
+    }
+    /* Redirect stderr to stdout */
```
nojb, line 90: Don't we want to redirect stdout to stderr instead ?

MisterDA: Oh yes, I read `1>&2` incorrectly.

src/platform/platform_misc.c
```diff
@@ -68,29 +69,51 @@ static int windows_system(const char *cmd)
...
+    /* Duplicate stderr to make sure it is inheritable */
+    p_stderr = GetStdHandle(STD_ERROR_HANDLE);
+    if (p_stderr == INVALID_HANDLE_VALUE) {
+        err = GetLastError(); goto ret1;
+    }
+    hp = GetCurrentProcess();
+    if (! DuplicateHandle(hp, p_stderr, hp, &(s_info.hStdError),
+                          0, TRUE, DUPLICATE_SAME_ACCESS)) {
+        err = GetLastError(); goto ret1;
+    }
```
nojb, line 89: I don't understand why all this is necessary. I'm not very familiar with all this "inheritable" intrincacies, but if we are setting stderr to stdout, shouldn't we make sure that `stdout` is inheritable (which we already do in the call to `Duplicate Handle below)?

src/platform/platform_misc.c
```diff
@@ -68,29 +69,61 @@ static int windows_system(const char *cmd)
...
+    if (p_stderr == INVALID_HANDLE_VALUE) {
+        err = GetLastError(); goto ret1;
+    }
+    if (! GetHandleInformation(p_stderr, &handleInfo)) {
+        err = GetLastError(); goto ret1;
+    }
+    if (! (handleInfo & HANDLE_FLAG_INHERIT)) {
+        hp = GetCurrentProcess();
+        if (! DuplicateHandle(hp, p_stderr, hp, &(s_info.hStdError),
+                              0, TRUE, DUPLICATE_SAME_ACCESS)) {
```
nojb, line 92: Is this call to `DuplicateHandle` necessary, given that the following call to `DuplicateHandle` below also makes the `stderr` handle inheritable?

MisterDA: We're redirecting stdout to stderr so I'm considering stderr first. We need to make sure stderr of the parent (`p_stderr`) is inheritable (keep-on-exec), so if it's not, I duplicate it to mark in inheritable. If it is already inheritable, no need to duplicate it.
I think you're right that a `DuplicateHandle` is superfluous, but I'd say it's the second. It may be possible to set `s_info.hStdOutput = s_info.hStdError` instead of duplicating the handle again.

MisterDA:
I think with the latest push the code should be right: we still need to quote the arguments given to the program, we add `1>&2` when we're not running on Windows. For the process creation, we disable stdin of the child. We get the stderr of the parent, check whether it is inheritable (keep-on-exec) (it should be). If not, we duplicate it and mark it inheritable. We set the child's stderr and stdout to that handle.

src/platform/platform_misc.c
```diff
@@ -68,29 +69,61 @@ static int windows_system(const char *cmd)
...
+
+    /* If needed, duplicate stderr to make sure it is inheritable */
+    p_stderr = GetStdHandle(STD_ERROR_HANDLE);
+    if (p_stderr == INVALID_HANDLE_VALUE) {
+        err = GetLastError(); goto ret1;
+    }
+    if (! GetHandleInformation(p_stderr, &handleInfo)) {
+        err = GetLastError(); goto ret1;
+    }
+    hp = GetCurrentProcess();
```
nojb, line 90:
```suggestion
```
(the definition of `hp` is repeated below)

src/platform/platform_misc.c
```diff
@@ -68,29 +69,61 @@ static int windows_system(const char *cmd)
...
+        }
+    } else {
+        s_info.hStdError = p_stderr;
+    }
+
+    /* Redirect stdout to stderr */
+    s_info.hStdOutput = s_info.hStdError;
+
+    flags = CREATE_NO_WINDOW | CREATE_UNICODE_ENVIRONMENT;
+    char_os *utf16cmd = caml_stat_strdup_to_os(cmd);
```
nojb, line 105:
```suggestion
    WCHAR *utf16cmd = caml_stat_strdup_to_utf16(cmd);
```
(it is equivalent, but a bit clearer since this is Windows-specific code)

src/platform/platform_misc.c
```diff
@@ -68,29 +69,61 @@ static int windows_system(const char *cmd)
...
+    if (! CreateProcessW(NULL, utf16cmd, NULL, NULL,
+                         TRUE, flags, NULL, NULL, &s_info, &p_info)) {
+        err = GetLastError();
+    }
+    caml_stat_free(utf16cmd);
+    CloseHandle(s_info.hStdOutput);
```
nojb, line 111:
```suggestion
```
Since `s_info.hStdOutput` is the same as `s_info.hStdError`, the same handle is being closed twice below.

nojb (commented):
Looks good, but I think there are a couple of points to correct (and a stylistic suggestion).

MisterDA:
Thanks for the thorough review! A lot of omissions on my end...

nojb (approved):
LGTM. I can give this a try on my setup a bit later today and confirm that it works.

let-def:
Thanks @MisterDA and @nojb.

@MisterDA: can you update the changelog (`CHANGES.md`)?
And I will merge.

MisterDA:
> @MisterDA: can you update the changelog (`CHANGES.md`)? And I will merge.

Done.

CHANGES.md
```diff
@@ -4,6 +4,7 @@ git version
   + merlin binary
     - filter dups in source paths (#1218)
     - improve load path performance (#1323)
+    - correctly redirect stdout to stderr of ppx on Windows (#1413)
```
nojb, line 7: I would write something a bit more general so that people will notice it if looking for the specific bug:
```suggestion
    - fix handling of ppx's under Windows (#1413)
```

MisterDA: I see, thanks :) applied!

let-def:
Thank you ;).

voodoos:
@MisterDA should we backport this patch to Merlin for OCaml 411 and/or 412 in your opinion ?

MisterDA:
Yes. Merlin is still usable, but without the patch is broken when using ppx's. It cannot do anything when the code is in extension points (so that breaks for instance lwt_ppx or ppx_expect) or cannot show the type of functions generated by derivers (for instance ppx_sexp or yojson). I ignored errors on derivers, but on extension points it's more annoying not to be able to use Merlin on bigger chunks of code.
I don't think the patch would be difficult to backport.

PR #1414, opened 2021-12-13
Adding a buffer to display errors in emacs mode

panglesd:
In emacs mode, there is specific buffer for displaying types, which is useful when displaying large types, or keeping the type visible after other messages are sent to the echo area.

However, there is no buffer for displaying errors. This PR adds a `*merlin-errors*` buffer which records displayed errors on point.

Notable points:
- Contrary to the `*merlin-types*` buffer, it is not automatically opened when the error is long. This is because displaying errors is not run by command/shortcut but by putting the cursor on an error, which happen often without the user wanting to have the error buffer opened. 
- No major ocaml mode is loaded, as they do not fontify errors well.

image


voodoos:
Thank you !

emacs/merlin.el
```diff
@@ -629,6 +632,53 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
                                           (merlin-switch-list-by-ext '(".mli" ".ml")))))
   (merlin-switch-to name '(".mli" ".ml")))
 
+;;;;;;;;;;;;;;;;;;
+;; ERROR BUFFER ;;
+;;;;;;;;;;;;;;;;;;
+
+(defun merlin--error-is-short (text)
```
trefis, line 639: I don't see that function used anywhere.

panglesd: Yes, you are right. It is a remainder of when the buffer was automatically displayed when the error was too big to fit in the minibuffer. But that turned out to be annoying, so I removed it.

This function can be safely removed.

Sorry about that oversight! I'm too used to OCaml static analysis...

voodoos: Done in 5e25357d894a8b9935d47bb64283cf9583c6818d

PR #1418, opened 2021-12-20
Always generate LF files on Windows

MisterDA:
Setting the output to binary mode prevents the conversion of LF `\n` to CRLF `\r\n` on Windows, which simplifies the diff when the repo is cloned in LF line ending. This reverses the problem if the repo is cloned in CRLF be default, but come on, who's doing that.
Files copied from menhir or generated by menhir may still suffer from the problem:

```
        modified:   src/ocaml/preprocess/menhirLib.ml
        modified:   src/ocaml/preprocess/menhirLib.mli
        modified:   src/ocaml/preprocess/parser_raw.ml
        modified:   src/ocaml/preprocess/parser_raw.mli
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,11 @@ configure eol=lf
 *.sh eol=lf
 *.patch eol=lf
 tests/merlin-wrapper eol=lf
+
+src/ocaml/preprocess/menhirLib.ml text eol=lf
+src/ocaml/preprocess/menhirLib.mli text eol=lf
+src/ocaml/preprocess/parser_explain.ml text eol=lf
+src/ocaml/preprocess/parser_printer.ml text eol=lf
+src/ocaml/preprocess/parser_raw.ml text eol=lf
+src/ocaml/preprocess/parser_raw.mli text eol=lf
+src/ocaml/preprocess/parser_recover.ml text eol=lf
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,6 @@ git version
   + merlin binary
     - filter dups in source paths (#1218)
     - improve load path performance (#1323)
-    - output generated parser files in LF line endings on Windows (#1418)
 
 merlin 4.4
 ==========
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ git version
   + merlin binary
     - filter dups in source paths (#1218)
     - improve load path performance (#1323)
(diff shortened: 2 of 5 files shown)


nojb (commented):
Instead of doing this (ugly, IMO) trick, another approach is to add
```
<path> text eol=lf
```
to `.gitattributes`, one for each generated file `<path>`. This will inform `git` that we intend to check in only LF newlines for these files and will compute diffs appropriately.

MisterDA:
> Instead of doing this (ugly, IMO) trick, another approach is to add
> 
> ```
> <path> text eol=lf
> ```
> 
> to `.gitattributes`, one for each generated file `<path>`. This will inform `git` that we intend to check in only LF newlines for these files and will compute diffs appropriately.

It works, but the generated files still show up in the `git status`. There's also a warning when running `git diff`, that can be disabled with

    git config --global core.safecrlf false

nojb:
> It works, but the generated files still show up in the `git status`. There's also a warning when running `git diff`, that can be disabled with

Indeed, but I consider both of these as a feature, since it explains what is going on while still showing an empty `git diff`.

MisterDA:
My intent was to make it easier to rebase, build, and move commits around. If there are generated files with changes or conflicts, I have to stop and `git restore` the files with changes. I find that mildly annoying, and although going with gitattributes is definitely an improvement over the current situation, I think it'd be better not to generate the crlf in the first place.

nojb:
> My intent was to make it easier to rebase, build, and move commits around.

The `.gitattributes` approach also helps here because it guarantees that these files are always committed with LF end of lines.

dra27:
> This reverses the problem if the repo is cloned in CRLF be default, but come on, who's doing that.

All Windows users by default! (me included...)

dra27:
Ideally `.gitattributes` should always begin:

```gitattributes
# Default behaviour, for if core.autocrlf isn't set
* text=auto
```

voodoos:
@MisterDA @nojb what is the status on this ?

nojb:
> `.gitattributes`, one for each generated file `<path>`. This will inform `git` that we intend to check in only LF newlines for these files and will compute diffs appropriatel

This looks OK to merge to me. @dra27 suggested a slightly more general fix (setting `* text=auto`) which enforces LF end of lines on checkin for *all* text files instead of just the explicit list in this commit. That would also be fine.

Issue #1388, opened 2021-09-15
'+' in include path with dune


vzaliva:
I have the following dune file which compiles fine from command line but not under merlin:

```
(executable
 (name main)
 ; workaround for https://github.com/ocaml/dune/issues/3636
 (flags (:standard -w -27 -I +../cerberus/frontend))) 
```

In merling-log buffer I see:

```
"ocaml": {
    "include_dirs": [
      "/home/lord/src/cheri-c-semantics/ubfinder/+../cerberus/frontend"
    ],
```
If I change it to full path:

```
(executable
 (name main)
 ; workaround for https://github.com/ocaml/dune/issues/3636
 (flags (:standard -w -27 -I /home/lord/.opam/cerberus/lib/cerberus/frontend))) 
```

it shows in log:

```
  "ocaml": {
    "include_dirs": [ "/home/lord/.opam/cerberus/lib/cerberus/frontend" ],
```

and compiles fine.

It looks like an issue with `+` handling with `-I`

PR #1419, opened 2021-12-21
Fix issue 1388

voodoos:
Fixes issue #1388.

This necessitate an update of `dot-merlin-reader`.
I did not test with `dune ocaml-merlin`.

@vzaliva could you try [this branch](https://github.com/voodoos/merlin/tree/fix-issue-1388) and confirm whether it solves your issue ? Thank you !

### DIFF
--- a/tests/test-dirs/config/issue1388.t
+++ b/tests/test-dirs/config/issue1388.t
@@ -5,11 +5,10 @@
   > B +../cerberus/build
   > EOF
 
-FIXME include_dir should also replace the `+` by the lib path
   $ echo "" | $MERLIN single dump-configuration -filename test.ml | \
   > grep -A2 'include_dirs\|source_path\|build_path'
         "include_dirs": [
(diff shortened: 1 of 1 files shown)


vzaliva:
I've tested and it does not work. In merlin-log buffer: 

```
"include_dirs": [],
```

The corresponding flags stanza:

```
(flags (:standard -w -27 -I +../cerberus/frontend))
```

It compiles fine with `dune`.

voodoos:
Are you sure your editor plugin uses the Merlin from this branch ?

One way to do that is to use an opam pin:
`opam pin merlin git@github.com:voodoos/merlin.git#fix-issue-1388`

vzaliva:
yes, this is how I tested.

voodoos:
Ok. Unfortunately I am not able to reproduce a similar issue on my test project any more. 
Could you provide a small project and a few reproduction steps ?

PR #1425, opened 2022-01-10
Allow the use of dune-base configuration reader in tests

voodoos:
So far it was ankward to use the new dune-based configuration reader in the testsuite. The reason was that Merlin completely resets the environment and thus the `INSIDE_DUNE` variable was not forwarded to the spawned `.

This PR fixes that.
It might help to have a better testing story in #1219

### DIFF
--- a/src/frontend/ocamlmerlin/new/new_merlin.ml
+++ b/src/frontend/ocamlmerlin/new/new_merlin.ml
@@ -147,9 +147,12 @@ let run = function
         prerr_endline ("Exception: " ^ Printexc.to_string exn);
         1
 
-let run env wd args =
-  Os_ipc.merlin_set_environ env;
-  Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()));
+let run ~new_env wd args =
+  begin match new_env with
+  | Some env ->
+    Os_ipc.merlin_set_environ env;
+    Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()))
+  | None -> () end;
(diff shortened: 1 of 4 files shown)


src/frontend/ocamlmerlin/new/new_merlin.ml
```diff
@@ -147,9 +147,12 @@ let run = function
...
 
-let run env wd args =
-  Os_ipc.merlin_set_environ env;
-  Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()));
+let run ~new_env wd args =
+  begin match new_env with
+  | Some env ->
+    Os_ipc.merlin_set_environ env;
+    Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()))
+  | None -> () end;
```
let-def, line 155: I think we should do `     Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()))` in both branches.
This is the most reliable way for a process to know that it was spawned by Merlin.

voodoos: https://github.com/ocaml/merlin/pull/1425/files#diff-937b49ae01eae7a401f67f85552d50422bf71ef99f8c2a9f9f62eeb25bc3d93eR72-R74

let-def: Looks good. My bad :)

let-def (approved):

let-def:
Thanks :fireworks:

PR #1426, opened 2022-01-10
Add prefix command to `merlin-locate` to open a new window.

panglesd:
Currently, `merlin-locate` opens a new window if and only if the location is in another file.

However:
- When using `merlin-locate` several times in a row to find the definition we want, it can be annoying to have new windows erasing where we come from.
- Sometimes we want to keep the location we come from visible even if the definition is in the same file.

This PR gives the possibility to force a behavior, using the `prefix argument` mechanism:
- `C-u C-c C-l` force to stay in the same window, even if the cursor is moved to another file,
- `C-u C-u C-c C-l` force open another window, even if the cursor is moved the same file.

Other uses of the prefix defaults to the regular `C-c C-l`.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1527,10 +1527,16 @@ loading"
   (interactive "s> ")
   (merlin--locate-result (merlin-call-locate ident)))
 
-(defun merlin-locate ()
+(defun merlin-locate (prefix)
   "Locate the identifier under point"
-  (interactive)
-  (merlin--locate-result (merlin-call-locate)))
+  (interactive "P")
+  (cl-letf ((merlin-locate-in-new-window
+    (cond
+     ((equal prefix '(4)) 'never)
+     ((equal prefix '(16)) 'always)
+     (t 'merlin-locate-in-new-window)
+     )))
+    (merlin--locate-result (merlin-call-locate))))
 
 (defun merlin-locate-type ()
   "Locate the type of the expression under point."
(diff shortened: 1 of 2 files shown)


trefis:
We already have some variables for controlling the behavior wrt new windows.
Can't you override those locally instead of introducing new variables with the same meaning?

emacs/merlin.el
```diff
@@ -1527,10 +1527,16 @@ loading"
...
   "Locate the identifier under point"
-  (interactive)
-  (merlin--locate-result (merlin-call-locate)))
+  (interactive "P")
+  (cl-letf ((merlin-locate-in-new-window
+    (cond
+     ((equal prefix '(4)) 'never)
+     ((equal prefix '(16)) 'always)
+     (t 'merlin-locate-in-new-window)
+     )))
```
voodoos, line 1538:
```suggestion
     (t 'merlin-locate-in-new-window))))
```

voodoos:
For reference: https://www.emacswiki.org/emacs/PrefixArgument

Note to self: If I understand correctly `cl-letf` mask the global value of the variable so that using a prefix argument will change the behaviour only for that call to locate.

Looks good to me, I don't know if @trefis wants another look ? @panglesd could you add a changes entry and maybe clean the history ?

panglesd:
@voodoos Done!

mattiase:
> Note to self: If I understand correctly cl-letf masks the global value of the variable so that using a prefix argument will change the behaviour only for that call to locate.

Actually a plain `let` will do just as well and is more idiomatic elisp.

PR #1428, opened 2022-01-12
Add special exception for extension failure

let-def:
As requested in https://github.com/ocamllabs/vscode-ocaml-platform/issues/818.
@rgrinberg The exception constructor has path `Extend_main.Handshake.Error`, is it ok? Anything else I can do to help vscode-ocaml-platform code?

### DIFF
--- a/src/extend/extend_main.ml
+++ b/src/extend/extend_main.ml
@@ -104,6 +104,15 @@ module Handshake = struct
       prerr_endline "Unexpected value after handshake.";
       exit 1
 
+  exception Error of string
+
+  let () =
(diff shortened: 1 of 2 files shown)


rgrinberg:
This looks good. On the lsp side, I will simply turn this exception into a non fatal error and display it the message to the user. Thanks!

rgrinberg:
@let-def is this ready to merge?

let-def:
Done.

anmonteiro:
Should this be ported to https://github.com/let-def/merlin-extend/ too?

Issue #1421, opened 2022-01-03
Associate entry with toplevel comment and module comment


vch9:
Hello, this is more likely a request for help than an actual issue.

I am working with @rgrinberg fork vendored (similarly to `ocaml-lsp`), and I'm trying to work with
documented/undocumented entries in a OCaml program.

Using Merlin I can for now detect entries in a program (types, values, modules etc.) and with [odoc](https://github.com/ocaml/merlin/blob/master/src/analysis/ocamldoc.ml) determine whether they are documented or not. It is currently working fine except for two specific cases:

1. Toplevel documentation with a license header:
```ocaml
(*******************************)
(* This is a wonderful license *)
(*******************************)

(** Toplevel documentation *)

type t = ...
...
```

2. Module documentation inside module:
```ocaml
module Foo = struct
  (** Foo documentation *)

end
```

In the first Merlin does not seem to be capable to find the toplevel documentation. However, in the second case `merlin-document` on `Foo` succeeds to find the documentation. But, I was not able to manually reproduce `merlin-document` using [odoc](https://github.com/ocaml/merlin/blob/master/src/analysis/ocamldoc.ml).

I created a "minimal" [repository](https://github.com/vch9/merlin-analysis-module) with my actual use of the `merlin.analysis`.

There are the two only corner-cases I've found so far. Would you have any ideas on why they can not find the associated comments ?

voodoos:
Hello @vch9, what your are trying to achieve is not entirely clear to me. Are you using odoc for comparison with Merlin results or do you plan to mix the two ?

Also I think there is indeed an issue with Merlin but it is not linked to the licence header: can you try removing the space between the documentation string and the type declaration ?
```
(** Toplevel documentation *)
type t = ...
```

In my attempts Merlin find the doc only when there is no space, can you confirm that ?

I am not (yet) aware of the exact mechanics of Merlin's document command, @trefis , is that a bug or a feature ?

voodoos:
In fact this behaviour respects the [`ocamldoc` specification](https://ocaml.org/manual/ocamldoc.html#s%3Aocamldoc-comments):

> ## Comments in mli files:
> A special comment before an element is associated to this element if ‍:
> 
> - There is no blank line or another special comment between the special comment and the element. However, a regular comment can occur between the special comment and the element.
> - The special comment is not already associated to the previous element.
> - The special comment is not the first one of a toplevel module. 
> 
[...]

PR #1415, opened 2021-12-13
Upgrade Merlin to OCaml 4.14

voodoos:
## Roadmap

- [x] first draft which compiles succesfully
- [x] fix testsuite 
- [ ] merge new locate code when ready

@kit-ty-kate do you know if it is already possible to have some CI (Github or Ocaml CI) working here ?

PR #1431, opened 2022-01-14
[Shapes] OCaml 4.14 + Shapes

voodoos:
Another Pr for early testing.
- Based on #1415, an up-to-date upgrade to OCaml 4.14
- With the latest version of @gasche's [strong call-by-need PR](https://github.com/ocaml/ocaml/pull/10825)
- With the locate code that uses shapes similar to others recent testing PRs

- [x] Most tests are improved, one is not (jumping to declaration instead of a functor argument).
- [x] Documentation tests are not passing at all yet.
- [x] Add tests illustrating module aliases issue
- ~[ ] fix by patching the compiler~
- [x] fix documentation and dot-pp-dot-ml-dune tests 

The corresponding compiler (4.14 + @gasche's cbn) to be used with this merlin can be found here:
https://github.com/voodoos/ocaml/tree/shape-strong-call-by-need


voodoos:
@trefis all tests are now passing, and jumping to the argument of a functor works again. This PR is ready for review !

To restore that behaviour I had to store the functor arguments uid in the `uid_to_loc` table of the local environment. See c5e6d07.
I suppose it is not necessary to upstream this as it only occurs in local buffers. But we might want to to reduce the divergence between Merlin and the compiler... What do you think ?

voodoos:
@trefis : this now use uids to answer document queries when the target is in another comp_unit. It fallsback on the old heuristic when the target is in the current buffer or the previous algorithm found nothing (necessary for test `unattached-comment`).

I also reverted the changes for functor arguments. We now jump to the correct place in the argument's signature.

voodoos:
I rebased both the 414 branch and this PR.
This led to the discovery of an issue with module aliases.
I will add more tests to illustrate the issue.

- [ ] Add tests illustrating module aliases issue
- [ ] fix by patching the compiler
- [ ] fix documentation and dot-pp-dot-ml-dune tests

voodoos:
@trefis all test fixed. This is ready for review. My original intent was to merge this into the 4.14 update PR but maybe we can first merge the update PR onto master and this one second ?

- `get_doc` now use uids to get docstrings from external compilation units. I took that opportunity to fix https://github.com/ocaml/merlin/issues/1421.

tests/test-dirs/locate/locate-stdlib.t
```diff
@@ -0,0 +1,45 @@
...
+  > EOF
+
+  $ cat >dune <<EOF
+  > (executable (name s))
+  > EOF
+
+  $ dune build ./s.exe
+
+Should jump to set.ml:
+  $ $MERLIN single locate -look-for ml -position 5:13 -filename ./s.ml < ./s.ml
```
trefis, line 20: That's the kind of test whose output is likely going to be unstable: `set.ml` might change with each new release of ocaml, moving a bit the position returned by locate, creating noise in the diff we'll have to check.

What are you trying to test here and can we test it without adding "external dependencies"?

tests/test-dirs/locate/partial-cmt.t/run.t
```diff
@@ -48,9 +48,10 @@ That is: if the file is a.mli then the test is broken:
...
       "pos": {
         "line": 1,
-        "col": 0
+        "col": 9
       }
     },
     "notifications": []
   }
 
+TODO With shape it is extected that this test will fail.
```
trefis, line 57: expected*.

Also: let's just remove the test if we do not expect things to work with partial cmts.

voodoos: I kept it as an incentive :-)

PR #1432, opened 2022-01-20
Fix superfluous break in error reporting

voodoos:
Following a report by @panglesd this PR fix an issue with the formating of errors.

The following node:
` [%%ocaml.error "test with several words"]`

is reported by the compiler as:
`Error: test with several words`

but by merlin as:
```
test with several
words
```

The difference is that Merlin uses `Location.print_main` to produce the string while the compiler uses `Location.print_report`. The latter function wrap the printing of the `main` part of the report in a pretty-printing box while the former doesn't.
This PR wraps the output of `Location.print_main` in a pp box.

### DIFF
--- a/src/frontend/ocamlmerlin/query_json.ml
+++ b/src/frontend/ocamlmerlin/query_json.ml
@@ -255,8 +255,7 @@ let json_of_error (error : Location.error) =
   in
   let loc = Location.loc_of_report error in
   let msg =
-    Location.print_main Format.str_formatter error;
-    String.trim (Format.flush_str_formatter ())
+    Format.asprintf "@[%a@]" Location.print_main error |> String.trim
   in
   let typ =
     match error.source with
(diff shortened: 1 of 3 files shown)


src/frontend/ocamlmerlin/query_json.ml
```diff
@@ -255,8 +255,7 @@ let json_of_error (error : Location.error) =
   in
   let loc = Location.loc_of_report error in
   let msg =
-    Location.print_main Format.str_formatter error;
-    String.trim (Format.flush_str_formatter ())
+    Format.asprintf "@[%a@]" Location.print_main error |> String.trim
```
voodoos, line 258: @trefis maybe the use of a box makes the trim superfluous ?

voodoos: I have my answer: the testsuite disagrees strongly.

voodoos:
ping @trefis ?

trefis:
I'm a bit confused regarding how this is helping.
If the text fits on a line, why is a newline inserted in the absence of a box? Is it just that the formatter state is "dirty", leading to bad decisions when printing our string?

In any case, this seems "fine".

Issue #1221, opened 2020-12-09
Neovim bug: Cursor line jumps on :MerlinTypeOf


ddickstein:
@let-def's fix for issue #553 (7efcb22) introduces a bug in Neovim where the current line jumps to the top of the window on repeated calls to `MerlinTypeOf`. I do not yet understand why; only that I can reproduce it reliably.

1. Open Neovim to a single OCaml file. It must be the only window in the tab (not sure why).
2. Scroll down a bit so the top of your window is not at line 1 (again, not sure why).
3. Put your cursor over some OCaml value and run `:MerlinTypeOf`. Observe that works correctly.
4. Now run `:MerlinTypeOf` again on any value in *the same line*.
5. Observe that the window scrolls s.t. the current line is now at the top of the window.

![merlin-jump-2020-12-08_20 13 03](https://user-images.githubusercontent.com/802511/101561634-ae2b9680-3993-11eb-986b-feeefc75e9b7.gif)

I wonder whether 7efcb22 is still needed. When I tried undoing those changes the new issue disappears and I was unable to reproduce issue #553 (my type history buffer seems to work fine). Maybe a change to neovim and/or merlin in the last 4 years has obviated this change. Still, I don't really understand why this new bug happens -- it's pretty bizarre.

PR #1433, opened 2022-01-26
Revert "fix #553"

### LINKED ISSUE #553: :MerlinTypeHistory doesn't work with neovim
The type history buffer is constantly blank and all type queries are always sent to vim's pager rather the type buffer no matter how big are the results.

ddickstein:
This reverts commit 7efcb2265ed6525bf330666713f70cc83b32ce66.

The problem reported in #553 is not reproducible when this commit is reverted,
so it's likely that the underlying Neovim bug that caused it has been resolved.
The fix introduces a problem where the window is scrolled (#1221), so this
commit reverts it.


let-def:
Thank you!

PR #1436, opened 2022-02-08
Fix: destructing record field in pattern

ulugbekna:
First commit adds a test showing the problem; the second one fixes it. 

I hope I'm not leaving out some edge cases, went with the seemingly easiest fix.


trefis:
Doesn't your code break a call to destruct on `x` on the following example:
```ocaml
type r = { x : int }
type r_out = { r : r; y : string }

let f ({ r = x ; y } : r_out) = ()
```
by producing (in the end) `{ r = r = { x }; y }`?

Generally to detect punning you'll want to have a look at the location of the nodes.

Also, I'm surprised that the behavior depends on the _content_  of the field. :shrug:

PR #1437, opened 2022-02-11
Expose the Ill_typed exception for pkgs using Merlin as a library.

voodoos:
This is required by OCaml-LSP to improve error handling. 
See https://github.com/ocaml/ocaml-lsp/issues/617#issuecomment-1035508467

cc @ulugbekna


rgrinberg:
I think we can also just rebase? Or did the 4.14 made it in already?

Rudi.

voodoos:
> Or did the 4.14 made it in already? 
No, master is still on 4.13

rgrinberg (approved):
LGTM

PR #1439, opened 2022-02-18
Enable type-on-hover functionality in Vim

nilsbecker:
This PR refactors some functions in merlin.py to allow passing the mouse cursor position as input. In this way vim's balloonexpr functionality can be used to show the type at the mouse cursor. Hopefully this will be handy for browsing a codebase, where one does not necessarily wish to navigate the cursor to various places by keyboard commands.

Beware: the actualy popup window displayed by balloon-eval seems to be a bit flaky. I tested on two macos machines with identical MacVim version and identical .vimrc. The popup shows up only on one of the two. That is not a problem of merlin, as it fails also for the ALE plugin. The version with type display in the status line (or actually, below the status line? what is the bottom line of the screen called?) works in any case, so that should be a workable solution for everyone.

I'm happy for corrections, in particular also for the doc text.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@ git version
     - add module, module type, and class imenu items for emacs (#1244, @ivg)
     - add prefix argument to force or prevent opening in a new buffer in locate
       command (#1426, @panglesd)
-		- add type-on-hover functionality for vim (#1439)
+    - add type-on-hover functionality for vim (#1439, @nilsbecker)
   + test suite
     - cover locate calls on module aliases with and without dune
     - Add a test expliciting the interaction between locate and Dune's generated
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ git version
     - add module, module type, and class imenu items for emacs (#1244, @ivg)
     - add prefix argument to force or prevent opening in a new buffer in locate
       command (#1426, @panglesd)
+		- add type-on-hover functionality for vim (#1439)
   + test suite
     - cover locate calls on module aliases with and without dune
     - Add a test expliciting the interaction between locate and Dune's generated
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -644,15 +644,14 @@ def vim_case_analysis():
 
     vim_type_reset()
 
-def vim_type_enclosing():
+def type_enclosing_at_pos(to_line, to_col):
     global enclosing_types
     global current_enclosing
     vim_type_reset()
     try:
-        to_line, to_col = vim.current.window.cursor
         enclosing_types = command2(
                 ["type-enclosing",
-                 "-position", fmtpos((to_line,to_col)),
+                 "-position", fmtpos((to_line, to_col)),
                  "-index", "0"
                 ],
                 track_verbosity=True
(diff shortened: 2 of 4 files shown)


vim/merlin/autoload/merlin.vim
```diff
@@ -270,6 +270,22 @@ function! merlin#TypeOfSel()
   call merlin#TypeOf(s:get_visual_selection())
 endfunction
 
+function! s:type_at_balloon()
+	MerlinPy vim.command("let l:type = " + merlin.vim_type_enclosing_at_pos(
+				\ vim.vvars['beval_bufnr'], vim.vvars['beval_lnum'], vim.vvars['beval_col']))
```
trefis, line 275: I think I'd rather see the following function in `merlin.py`:
```python
def vim_type_enclosing_at_mouse():
  bufnr = vim.vvars['beval_bufnr']
  if bufnr != vim.current.buffer.number:
        return '{}'
  line = vim.vvars['beval_lnum']
  col = vim.vvars['beval_col']
  return type_enclosing_at_pos(line, col)
```

And then this line can be changed to
```
MerlinPy vim.command("let l:type = " + merlin.vim_type_enclosing_at_mouse())
```

Would there be an issue with that?

vim/merlin/autoload/merlin.vim
```diff
@@ -270,6 +270,22 @@ function! merlin#TypeOfSel()
...
+	return l:type  " a Dictionary...
+endfunction
+
+function! merlin#TypeAtBalloon()
```
trefis, line 279: Can't this function be merged with the previous one?

vim/merlin/autoload/merlin.py
```diff
@@ -644,15 +644,16 @@ def vim_case_analysis():
 
     vim_type_reset()
 
-def vim_type_enclosing():
+def vim_type_enclosing_at_pos(bufnr, to_line, to_col):
     global enclosing_types
     global current_enclosing
+    if bufnr != vim.current.buffer.number:
+        return '{}'
```
trefis, line 651: This can disappear (and the function can be renamed to lose the `vim_` prefix) if the other change I proposed is implemented.

trefis (changes requested):
Thanks for the PR!

I have a couple of suggestions, let me know if they make sense or not!

vim/merlin/autoload/merlin.vim
nilsbecker: That seems better, i'll try it.

trefis (approved):
LGTM.

trefis:
(The CI failure seems unrelated, but I'll wait for @voodoos' confirmation before merging).

voodoos:
> (The CI failure seems unrelated, but I'll wait for @voodoos' confirmation before merging).

Yes it was unrelated. @nilsbecker can you rebase your PR and add a changelog entry ?

nilsbecker:
@voodoos does it look ok now?

CHANGES.md
```diff
@@ -12,6 +12,7 @@ git version
     - fix superfluous break in error reporting (#1432)
   + editor modes
     - add module, module type, and class imenu items for emacs (#1244, @ivg)
+		- add type-on-hover functionality for vim (#1439)
```
voodoos, line 15:
```suggestion
    - add module, module type, and class imenu items for emacs (#1244, @ivg)
    - add type-on-hover functionality for vim (#1439)
```

voodoos:
Thanks for the changes. It would be great if you could rebase to clean the history a bit by squashing the fixes and simplifications.

voodoos:
Thank you !

PR #1440, opened 2022-03-05
add test: show locate failing in presence of same name definition in record punning

ulugbekna:


voodoos:
I feel like the current behaviour is actually correct:

When calling jump-to-definition at that (`|`) point:
```ocaml
module M = struct
  type t = { foo : int }
  
  let foo = 42
end

let foo = 43

let m = { M.fo|o }
```
I do expect the cursor to jump to the definition of  `M.foo` not the field in the type `M.t`.

However the behaviour when jumping explicitely to the label is broken so I added a test to illustrate it.

ulugbekna:
> I do expect the cursor to jump to the definition of M.foo not the field foo in the type M.t.

Oh, I meant that the cursor should jump to `let foo = 43` on line 7 because that will be the value of `m.foo`. Someone who wants to jump to the definition of `foo` in `{ M.foo }` likely wants to understand what's going to be the value of the field `foo` in the new record being created. Wdyt?

voodoos:
Yes that makes sense.
I will update the test before merging.

PR #1441, opened 2022-03-11
Remove unused module and dependency

kit-ty-kate:
As far as I can see, the `result` compatibility package is not required anymore

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ git version
     - expose all destruct exceptions in the api (#1437)
     - fix superfluous break in error reporting (#1432)
     - recognise binding operators in locate and occurrences (#1398, @mattiase)
+    - remove dependency on Result (#1441, @kit-ty-kate)
   + editor modes
     - fix an issue in Neovim where the current line jumps to the top of the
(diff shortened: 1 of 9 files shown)


voodoos:
There was a discussion about that change in https://github.com/ocaml/merlin/pull/1293 
What is the current status ? The other PR was also removing `(implicit_transitive_deps false)` from the `dune_project`.

Notably the discussion says that we should bump the version of `csexp` which depends on `result` until its latest version `1.5.1`. Is that still the case ?

voodoos:
@kit-ty-kate what motivated the removal of the dependency ?
Does the fact that we rely on Result causes issues with other packages ?

I guess that since we don't target old OCaml versions any-more it is fairly safe to get rid of the compat package ?

kit-ty-kate:
> I guess that since we don't target old OCaml versions any-more it is fairly safe to get rid of the compat package ?

yes, as well as being an annoying compatibility layer which isn’t compatible with what it’s trying to be compatible with (e.g. Result.result) and being the main cause of problems when using the `(implicit_transitive_deps false)` feature.

voodoos:
Note: we should probably bump csexp version to 1.5.1

PR #1445, opened 2022-03-18
Locate infix operators

### LINKED ISSUE #949: merlin-locate does not work on +.
`merlin-locate` does not work when the point in on a dotted infix operator such as `+.` (in Emacs).

Example:
```ocaml
module A = struct
  let ( +. ) a b = a +. b
end

let f x = A.(x +. 1.)
``` 
(put the cursor on the latest `+.` and `M-x merlin-locate`).

voodoos:
*~Two~ Three years later*

This fixes #949 

In #949 the issue was identified as `Longident.parse` returning wrong lids for infix operators. A PR was merged in the compiler to expose the true longident parser.

In fact this PR relies on `Longident.unflatten`.

~~This PR uses that parser to parse longidents produced by `reconstruct_identifier`.
However for usages of an infix operator the reconstructed identifier does not have surrounding parenthesis and the parser fails on these.

I chose to simply rerun the parser with added parenthesis if it fails, and fallback on `Longident.parse` if it fails again.
We probably could have a more deterministic style here, but I am not sure if that is worth the effort... @trefis ?~~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ unreleased
 
   + merlin binary
     - Update internal typer to match OCaml 4.14.1 release. (#1557)
+    - fix locate calls on infix operators (#1445, fixes #949)
 
 merlin 4.7
 ==========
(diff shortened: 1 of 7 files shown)


bbatsov:
> ~Two~ Three years later

Better late than never. ;-)

voodoos:
:-)
There are a still few things I want to change before considering reviewing and merging that PR.

voodoos:
Closing in favor of #1612

Issue #1446, opened 2022-03-22
Run CI more selectively


mattiase:
Running the full CI for every change is a waste of computrons. For example, a change in Emacs files only shouldn't necessitate a full OCaml CI run and vice versa, and changing CHANGES.md (for example) shouldn't trigger either.

mattiase:
See PR #1447.

mattiase:
Thanks. Not sure how to gate the "ocaml-ci" action; presumably it should be inhibited for non-OCaml changes.

PR #1447, opened 2022-03-22
Filter files to trigger github CI (#1446)

mattiase:
Maybe this would do; my own knowledge in these matters is sketchy at best.


voodoos:
That's a good idea, triggering the CI for every documentation/plug-in changes was quite wasteful.

Thank you !

PR #1452, opened 2022-03-30
Reformat elisp doc strings to silence compiler warnings

mattiase:
Since Emacs 28, the elisp compiler warns on doc strings with lines
wider than 80 characters.


voodoos (approved):
Thank you !

PR #1172, opened 2020-09-10
Add public_name to merlin libraries (upstream from `ocaml-lsp`)

voodoos:
ping @rgrinberg

PR #1455, opened 2022-04-11
Make merlin libraries public and wrapped

tmattio:
Alternative to https://github.com/ocaml/merlin/pull/1448 and https://github.com/ocaml/merlin/pull/1172 to make the libraries public and wrap them under a single top-level module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,19 @@ External contributors have implemented modes for more editors:
 - [nuclide for Atom](https://nuclide.io/) includes Merlin support
 - [Sublime Text 3](https://github.com/cynddl/sublime-text-merlin)
 
+
+Merlin as a library
+===================
+
+Merlin can also be used a library. Exemplary projects that use Merlin's libraries are:
+
+- [OCaml LSP](https://github.com/ocaml/ocaml-lsp) - The official OCaml's Language Server Protocol implementation
+
+If you're building editor tools, you might also want to use Merlin as a library!
+
+Note, however, that Merlin's public API is not stable and we don't garentee backward-compatibility between releases.
+If you're a Merlin user and depend on a public API, we recommend that you contact us or open an issue.
+
 Next steps
 ==========
 
--- a/README.md
+++ b/README.md
@@ -158,14 +158,14 @@ External contributors have implemented modes for more editors:
 Merlin as a library
 ===================
 
-Merlin can also be used a library. Exemplary projects that use Merlin's libraries are:
+Merlin can also be used as a library. Some projects already rely on this:
 
 - [OCaml LSP](https://github.com/ocaml/ocaml-lsp) - The official OCaml's Language Server Protocol implementation
 
 If you're building editor tools, you might also want to use Merlin as a library!
 
-Note, however, that Merlin's public API is not stable and we don't garentee backward-compatibility between releases.
-If you're a Merlin user and depend on a public API, we recommend that you contact us or open an issue.
+Note, however, that Merlin's public API is not stable and we don't guarantee backward-compatibility between releases.
+If you're a Merlin user and depend on our public API, we recommend that you contact us or open an issue.
 
 Next steps
 ==========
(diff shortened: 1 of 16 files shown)


voodoos:
This PR is still missing a few that are used by at least `ocaml-lsp`: 
- `query_protocol`
- `query_commands`

@rgrinberg do you see anything else missing ?

@trefis any thoughts about this ?

voodoos:
Also I think we should document somewhere that we don't provide any guarantee on the stability of the publicly available libraries. Maybe somewhere in the readme ?

tmattio:
> how this PR is better than my original PR

It's probably not, actually looking a the diff, this looks very similar 🙂 

It differs from your PR by:

- Making the libraries public and wrapping the libraries at the same time
- Changing the public names to `merlin.ocaml_*` for everything that is copied from OCaml (e.g. `merlin.ocaml_utils`)
- Rebasing on top of master
- Fixing some dependency cycles in the test suite to make the CI pass

rgrinberg:
The other changes are unrelated to wrapping so I think they should be done on top of my changes.

Rudi.

voodoos:
@rgrinberg the main thing is that finally PR introduces `public_name`s which is mostly an upstream of https://github.com/rgrinberg/merlin/commit/e077c9b03dcbf8656f609600360e8d160c42740c

Making Merlin libraries public has been in the backlog for quite some time now.. simply adding public names is not as satisfactory as having separate packages but we decided that it will do for now. 

@tmattio proposed this PR because he had it already done for another project that vendors Merlin. It does overlaps yours but we can still merge them successively if you prefer.

tmattio:
> It does overlaps yours but we can still merge them successively if you prefer.

I've added @rgrinberg as a co-author of the PR. I don't see much value in merging the other PR first since it's also implemented in the present PR, which is already rebased on master and fixes the CI, but I'll go with whatever you prefer 🙂

rgrinberg:
It's more of an issue because I'm already using commits from my branch in LSP so it would create extra surgery work.

tmattio:
> It's more of an issue because I'm already using commits from my branch in LSP so it would create extra surgery work.

Gotcha! Seems like @voodoos merged your branch, so I'll rebase this PR 🙂

README.md
```diff
@@ -154,6 +154,19 @@ External contributors have implemented modes for more editors:
...
+===================
+
+Merlin can also be used a library. Exemplary projects that use Merlin's libraries are:
+
+- [OCaml LSP](https://github.com/ocaml/ocaml-lsp) - The official OCaml's Language Server Protocol implementation
+
+If you're building editor tools, you might also want to use Merlin as a library!
+
+Note, however, that Merlin's public API is not stable and we don't garentee backward-compatibility between releases.
+If you're a Merlin user and depend on a public API, we recommend that you contact us or open an issue.
```
voodoos, line 168:
```suggestion
Merlin can also be used as a library. Some projects already use Merlin's libraries:

- [OCaml LSP](https://github.com/ocaml/ocaml-lsp) - The official OCaml's Language Server Protocol implementation

If you're building editor tools, you might also want to use Merlin as a library!

Note, however, that Merlin's public API is not stable and we don't guarantee backward-compatibility between releases.
If you're a Merlin user and depend on our public API, we recommend that you contact us or open an issue.
```

voodoos, line 168:
```suggestion
Merlin can also be used as a library. Some projects already use Merlin's libraries:

- [OCaml LSP](https://github.com/ocaml/ocaml-lsp) - The official OCaml's Language Server Protocol implementation

If you're building editor tools, you might also want to use Merlin as a library!

Note, however, that Merlin's public API is not stable and we don't guarantee backward-compatibility between releases.
If you're a Merlin user and depend on our public API, we recommend that you contact us or open an issue.
```

voodoos (approved):
LGTM, thank you @tmattio!

PR #1457, opened 2022-04-12
Split Merlin in two (three) packages

voodoos:
Most of the libraries are now publicly avalaible in the `merlin-lib` package.
This include the protocol: `query_commands` and `query_protocol` which are useful to lib users.
We might add `Query_json` too.

The `merlin` package contains the frontend.

All version of the packages are synced, but maybe this is not mandatory for `do-merlin-reader` ?


.github/workflows/main.yml
```diff
@@ -74,7 +74,9 @@ jobs:
           opam install . --deps-only --with-test
 
       - name: Build and test in release mode
-        run: opam exec -- dune runtest -p merlin,dot-merlin-reader
+        run: |
+          opam pin . --with-version dev --no-action
```
kit-ty-kate, line 78:
```suggestion
          opam pin . --with-version dev --no-action --cli 2.1
```

kit-ty-kate (approved):

merlin-lib.opam
```diff
@@ -0,0 +1,26 @@
...
+dev-repo:     "git+https://github.com/ocaml/merlin.git"
+license:      "MIT"
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
+  "ocaml" {>= "4.14" & < "4.15"}
+  "dune" {>= "2.9.0"}
+  "yojson" {>= "1.6.0"}
```
rgrinberg, line 15: Why does the backend need json encoding/decoding? Would it be possible to move this dependency to the frontend?

voodoos: There are dumping functions in `Query_commands` (and `Mconfig` and `Warnings`) that produce Json.
The `Dump` query is the only one that produces json before going through `Query_json.of_response`.

@let-def is there a reason for that ? It looks like it could be refactored but it is probably a bit tedious to do.

let-def: OCaml code is annoying to debug and having a common format for printing internals helped in the past.
However backend and frontend only needs to agree on this format.
To remove the dependency on Yojson, we can just copy the polymorphic variant definition of json in the backend.

voodoos:
I removed the dependency on yojson but it came with a cost: the logger does pretty print json in some cases, so I had to add the pretty printing code from yojson to `merlin-lib`'s std.

rgrinberg:
> That seems silly to me.

Indeed. Another alternative is to leave a global reference that contains a `Json.t -> string` (or formatter) that users of the library can use to add their own encoding. You would think that a library would allow callers to customize logging anyway.

> Let's just depend on yojson? The "plateform" people will eventually vendor everything IIUC, no?

In lsp, I would like to experiment with a different json encoding/decoding layer than yojson for efficiency. It's not something I need right away, but i do plan to drop yojson eventually.

voodoos:
> Indeed. Another alternative is to leave a global reference that contains a `Json.t -> string` (or formatter) that users of the library can use to add their own encoding. You would think that a library would allow callers to customize logging anyway.

I tried that in [0b25a95](https://github.com/ocaml/merlin/pull/1457/commits/0b25a951697bf3ae4b68cb563abbacd6f6e60050). It is not the cleanest of the api but it does feels a lot better than the previous solution. What do you think ?

rgrinberg:
> > Indeed. Another alternative is to leave a global reference that contains a `Json.t -> string` (or formatter) that users of the library can use to add their own encoding. You would think that a library would allow callers to customize logging anyway.
> 
> 
> 
> I tried that in [0b25a95](https://github.com/ocaml/merlin/pull/1457/commits/0b25a951697bf3ae4b68cb563abbacd6f6e60050). It is not the cleanest of the apis but it does feels a lot better than the previous solution. What do you think ?
> 
> 

IMO it's preferable to copy-pasting or doing nothing.

src/utils/std.ml
```diff
@@ -40,19 +49,10 @@ module Json = struct
   let list f x =
     `List (List.map f x)
 
-  let print f () x =
-    pretty_to_string (f x)
+  let pretty_to_string : (t -> string) option ref =
+    ref None
```
trefis, line 53: Instead of the option, you could just set it to your default printer as defined above:
```ocaml
fun _ ->
  Printf.sprintf
    "Logger error: `Std.Json.pretty_to_string` \
     is not set. You should initialize that reference with the \
     pretifier of your choice to enable json logging. \
     A common one is `Yojson.Basic.pretty_to_string`."
```

PR #1461, opened 2022-04-25
use unquoted merlin-locate-in-new-window with merlin-locate

### LINKED ISSUE #1460: merlin-locate-in-new-window is ignored after #1426
I’ve been using `merlin-locate-in-new-window` set to `'never` for a while to prevent Merlin to open a new window.

Recently, it stopped working altogether, and it as far as I can tell, it is related to https://github.com/ocaml/merlin/pull/1426.

I am not familiar enough with prefix argument, but the fact that this change breaks an existing configuration was a bit… annoying :sweat_smile:. Fortunately, I’ve been able to recover my previous setup by defining my own helper around `merlin-locate`:

```elisp
(defun +merlin-locate ()
    (interactive)
    ;; force `merlin-locate-in-new-window' to be 'never
    (merlin-locate '(4)))
```

I have a fairly customized Emacs configuration, but just calling `M-x merlin-locate` ignores `merlin-locate-in-new-window`, wihch looks like a bad UX to me.

emturner:
close #1460 

when calling `merlin-locate`, if there was no prefix set, then `merlin-locate-in-new-window` was set to `'merlin-locate-in-new-window` rather than the actual value

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1582,7 +1582,7 @@ loading"
     (cond
      ((equal prefix '(4)) 'never)
      ((equal prefix '(16)) 'always)
-     (t 'merlin-locate-in-new-window))))
+     (t merlin-locate-in-new-window))))
     (merlin--locate-result (merlin-call-locate))))
 
 (defun merlin-locate-type ()


lthms:
As far as I can tell, it “works on my machine.” (:

jberdine:
FWIW, I have tested this branch and am using it without issues.

Issue #453, opened 2015-10-18
source files have the same name -- but one of them does not exist


nilsbecker:
i am getting this error when using locate on CCHashtbl from `containers`:

     Several source files in your path have the same name: 

and then the path is _exactly_ the same, in my opam install location, but the two file names are `CCHashtbl.mli` and `cCHashtbl.mli`. the second one does not exist at that location.

def-lkb:
I assume you are using OSX? Then it is likely both exists :).

OSX has a weird fs semantics, sometimes described as "case insensitive but case aware", which confuses Merlin. This will be fixed...

nilsbecker:
yes, it's os x. fwiw, the directory `~/.opam/4.02.3/lib/containers` contains really only one file matching `??Hashtbl.mli`.

dbuenzli:
Related to #425

nilsbecker:
wait, now after installing from `#prerelease` i cannot reproduce it anymore. it may be already fixed.

nilsbecker:
ok, now invoking `MerlinLocate` on `CCHashtbl` does indeed work. what still does not work is invoking `MerlinLocate` on `CCHashtbl.Make` (ie on `Make`). the error message is that the ML file for CCHashtbl cannot be found. but it would be ok to just look up the MLI file just as for the module? this may be unrelated

Issue #425, opened 2015-08-21
String STring macosx


dbuenzli:
file system. @def-lkb knows what I mean.

PR #1462, opened 2022-04-28
Attempt at finding the 'real' capitalization of files on windows

mlasson:
# Description 

When merlin is locating module files, it first tries the capitalized filename (eg. "Module.ml") before the uncapitalized one ("module.ml"). On windows, filenames are case insensitive by default so the first filename matches even if the "real" case is uncapitalized. 

This is not a big problem (because... well it is case insensitive) but it still may confuse text editors that are interacting with merlin (eg. vscode through ocaml-lsp). Even if it is mostly cosmetics (you end up having tabs name that are capitalized when opened first after "Go To Definition", or "recently opened files" appears twice...). 

# What does this PR do ? 

It uses "FindFirstFile" that acts as a kind of "dir/ls" that returns the first filename that matches the inputs query. Here the query is an absolute path without wildcard; so there is at most one match which return as a string option. 

Let's say I have a document "Document.txt" in my home directory, then FindFirstFile("C:\UsErS\MlAsSon\dOcUmEnT.txt") would return "Document.txt".  

This "exact case basename" is then returned to the caml code which compares it with the input basename to check if they are the same (not the same in my example, since "Document.txt" is not "dOcUmEnT.txt"). May be an alternative implementation could have been to to this comparison in the C code. 🤷

Note: this method does not "fix" the case for "whole path" only for the filename (contrary to what has been done 
in commit bf79586fd3b3101f1f7f8e9a34da2f54340b48b0 for macos). 
 
I guess it is related with issues #425 ( and #453) (I guess these "Fixed need test" issues could be closed since it has been tested for 7 years now :)).


nojb:
For info, I also observed this issue in `emacs`.

nojb:
cc @MisterDA who has worked on previous Windows PRs

voodoos:
Anyone wants to review this ? @MisterDA, @nojb, @let-def ?

MisterDA:
Without testing the Windows specific code, one could also test that Merlin supports well [case-insensitive filesystems on Linux](https://www.collabora.com/news-and-blog/blog/2020/08/27/using-the-linux-kernel-case-insensitive-feature-in-ext4/). Maybe the issue can be reproduced there for non-Windows devs.

MisterDA:
I wonder if it wouldn't be better to open the file anyway, and get the "real" filename from the handle. Also, if we just compare the filenames, there might be problems with Unicode normalization.
https://docs.microsoft.com/en-us/windows/win32/memory/obtaining-a-file-name-from-a-file-handle

MisterDA:
Current master:

```
$ file=src/frontend/ocamlmerlin/ocamlmerlin_server.ml
$ dune exec -- ocamlmerlin single locate -position '95:11' -filename "$file" < "$file"
{"class":"return","value":{"file":"C:\\Users\\antonin\\Tarides\\merlin\\src\\utils\\Std.ml","pos":{"line":52,"col":6}},"notifications":[],"timing":{"clock":47,"cpu":31,"query":0,"pp":0,"reader":0,"ppx":0,"typer":31,"error":0}}
```

With this patch applied:

```
$ file=src/frontend/ocamlmerlin/ocamlmerlin_server.ml
$ dune exec -- ocamlmerlin single locate -position '95:11' -filename "$file" < "$file"
{"class":"return","value":{"file":"C:\\Users\\antonin\\Tarides\\merlin\\src\\utils\\std.ml","pos":{"line":52,"col":6}},"notifications":[],"timing":{"clock":31,"cpu":31,"query":0,"pp":0,"reader":0,"ppx":0,"typer":31,"error":0}}
```

So it seems to be working as expected (the `std.ml` file is lower-case in the returned json).

nojb (approved):
LGTM

nojb:
> I wonder if it wouldn't be better to open the file anyway, and get the "real" filename from the handle.

Yes, this is another possibility but then you need to parse the returned "real" path to extract the basename. I feel the current approach is simpler.

> Also, if we just compare the filenames, there might be problems with Unicode normalization.

In theory this may also happen already under macOS that also has a special logic in this function, right? But in practice I wonder if it will ever happen that the basename contains non-ASCII characters. (And these would not be valid OCaml module names anyway.)

voodoos:
Thank you @mlasson, could you add a changelog entry and maybe squash the irrelevant commits?

And thanks @MisterDA and @nojb for the review and testing!

mlasson:
Done !

PR #1463, opened 2022-05-02
Fix `merlin-locate-in-new-window` being ignored

panglesd:
Fixes the bug reported in #1460 and introduced in #1426, where the variable `merlin-locate-in-new-window` is ignored.


panglesd:
Closing as duplicate of #1461

PR #1469, opened 2022-05-18
Add test about upcoming occurences (WIP on ocaml-lsp issue 147)

Kakadu:
See also
https://github.com/ocaml/ocaml-lsp/issues/147

@voodoos ^^


trefis:
Thanks, but we'll add tests when implementing the feature.
Also, I'd rather not:
- have extra dependencies for our testsuite
- add screenshots to the repository

voodoos:
The title is wrong, the test is more about: "enclosing highlighting is broken when a PPX is used".
Thanks for the test @Kakadu, I will use it for reproduction when investigating the issue :-)

Kakadu:
@voodoos Should I create an  issue to make  PR not being forgotten?

Kakadu:
Please, ping me when you will have something to test

------- Original Message -------

voodoos:
Hi @Kakadu, we recently merged our first test relying on `ppxlib` so we now officially have it as a test dependency. 
Could you remove the dependency on `base` in your test ?
I will review it afterwards.

Kakadu:
> Hi @Kakadu, we recently merged our first test relying on `ppxlib` so we now officially have it as a test dependency.
> Could you remove the dependency on `base` in your test ?
> I will review it afterwards.

done

Kakadu:
@voodoos ^^

voodoos:
I finally found the time to look at the issue. As usual with this kind of things the problem is that the ppx break the location in a way that confuse Merlin's heuristic.

It is possible to fix the issue by removing line `48` of the `ppx_fresh.ml` file (that is the `let loc = efresh.pexp_loc in` definition). This way `loc = e.pexp_loc` and it fixes the problem.

We should think of some guidelines to help ppx-writer maintain the compatibility with Merlin and have them easily findable in both ppxlib and merlin documentations.

Kakadu:
It's very sad that it works) Basically, it means that my original issue is not related to occurrences, but to something else. 
As before when I put a blinking cursor over `a`, it highlights whole right hand side in the first line. Without macro it highlight occurrences of that `a` (on the second line). @voodoos, could you try get my screen shot in vscode + ocaml-lsp (or what do you usually use)? I'm again not sure where should I look for the source of issue...

Screenshot_20221014_232424

voodoos:
I see, in fact the problem is not "fixed" as I said in my previous comment: it does make `occurrences` return some results but these are wrong:

```
  $ ocamlmerlin single occurrences -identifier-at 4:28 -filename ./lib1.ml < ./lib1.ml | jq 'del(.timing)'
  {
    "class": "return",
    "value": [
      {
        "start": {
          "line": 4,
          "col": 12
        },
        "end": {
          "line": 4,
          "col": 29
        }
      },
      {
        "start": {
          "line": 4,
          "col": 27
        },
        "end": {
          "line": 4,
          "col": 28
        }
      }
    ],
    "notifications": []
  }
  ```
  
  Notice the first one, it is a selection too big to be an occurence of `a` and so it overlaps the over one in vscode.
  Again, this is probably not an issue with Merlin, but a problem with locations in the PPX.

voodoos:
And I think I found it. Could you try:
```diff
         let pat =
-          Pat.var
-            ~loc:var_loc
-              (* we declare a pattern with location of original identifier *)
-            (Ast_builder.Default.Located.mk var_name ~loc)
+          let loc = var_loc in
+          (* we declare a pattern with location of original identifier *)
+          Pat.var ~loc (Ast_builder.Default.Located.mk var_name ~loc)
         in
         ```
         
         Notice that in the original code the second loc was not `var_loc`.

voodoos:
@Kakadu could you confirm that the changes I proposed in my previous post fix the issue ?

Kakadu:
> @Kakadu could you confirm that the changes I proposed in my previous post fix your issue ?

Not yet, @voodoos. I tried to apply fixes you described but failed (still don't get any occurrences). Could you add a commit with right fixes?

voodoos:
It is because you commented line 26 instead of line 48 as described in a previous comment: https://github.com/ocaml/merlin/pull/1469#issuecomment-1279093221

Also you should try it in your real project, so you can see it live in your editor after a rebuild.

Changes (from your original file) are:

```diff
         let pat =
-          Pat.var
-            ~loc:var_loc
-              (* we declare a pattern with location of original identifier *)
-            (Ast_builder.Default.Located.mk var_name ~loc)
+          let loc = var_loc in
+          (* we declare a pattern with location of original identifier *)
+          Pat.var ~loc (Ast_builder.Default.Located.mk var_name ~loc)
         in
-        let loc = efresh.pexp_loc in
+       (* let loc = efresh.pexp_loc in *)
         [%expr call_fresh (fun [%p pat] -> [%e new_body])]
      | _ -> super#expression e
 ```

Kakadu:
I managed to apply ideas behind these changes to my real code. Now it is much better. Many thanks, @voodoos

PR #1475, opened 2022-06-02
Use the Yojson 2.0 `Seq` API

Leonidas-from-XIV:
Yojson 2.0 is dropping☨ the `Stream` API for compatibility with OCaml 5, thus it added an `Seq` based API. This PR does the small changes that are required to use it instead of `Stream`.

Which also comes with the added benefit of removing usage of `Stream` from Merlin itself.

☨Yojson 2.0 will drop the API when it is released. As such this PR requires the `master` version of Yojson but the idea is to prepare most Yojson consumers in advantage, so when Yojson 2.0 ends up on OPAM the migration is as smooth as possible.

### DIFF
--- a/src/frontend/ocamlmerlin/old/old_IO.ml
+++ b/src/frontend/ocamlmerlin/old/old_IO.ml
@@ -314,8 +314,7 @@ let make_json ?(on_read=ignore) ~input ~output () =
       read buf len
   in
   let lexbuf  = Lexing.from_function read in
-  let input   = Yojson.Basic.(stream_from_lexbuf (init_lexer ()) lexbuf) in
-  let input () = try Some (Stream.next input) with Stream.Failure -> None in
+  let input   = Seq.to_dispenser (Yojson.Basic.(seq_from_lexbuf (init_lexer ()) lexbuf)) in
   let output  = Unix.out_channel_of_descr output in
   let output' = Yojson.Basic.to_channel output in
   let output json =
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@ unreleased
       @kit-ty-kate)
     - Attempt at finding the 'real' capitalization of files on windows (#1462 by
       @mlasson)
+    - Use newer `Seq`-based API of Yojson 2.0, avoiding the need for the
+      deprecated `Stream` module (#1475 by @Leonidas-from-XIV)
 
 merlin 4.5
 ==========
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ features.
 Compilation
 -----------
 
-Dependencies: ocamlfind, yojson >= 1.6.0, dune >= 2.7.
+Dependencies: ocamlfind, yojson >= 2.0.0, dune >= 2.7.
 
 ```shell
 dune build -p dot-merlin-reader,merlin
--- a/merlin.opam
+++ b/merlin.opam
@@ -15,7 +15,7 @@ depends: [
   "dune" {>= "2.9.0"}
   "merlin-lib" {= version}
   "dot-merlin-reader" {= version}
-  "yojson" {>= "1.6.0"}
+  "yojson" {>= "2.0.0"}
   "conf-jq" {with-test}
 ]
 conflicts: [


src/frontend/ocamlmerlin/old/old_IO.ml
```diff
@@ -314,8 +314,13 @@ let make_json ?(on_read=ignore) ~input ~output () =
...
   let lexbuf  = Lexing.from_function read in
-  let input   = Yojson.Basic.(stream_from_lexbuf (init_lexer ()) lexbuf) in
-  let input () = try Some (Stream.next input) with Stream.Failure -> None in
+  let input   = ref (Yojson.Basic.(seq_from_lexbuf (init_lexer ()) lexbuf)) in
+  let input () = match Seq.uncons !input with
+    | Some (hd, tail) ->
+        input := tail;
+        Some hd
+    | None -> None
+  in
```
nojb, line 323: Can use `Seq.to_dispenser` instead.

Leonidas-from-XIV: Yes, that's true, it actually obsoletes the entire function. Fixed.

src/frontend/ocamlmerlin/old/old_IO.ml
```diff
@@ -314,8 +314,7 @@ let make_json ?(on_read=ignore) ~input ~output () =
       read buf len
   in
   let lexbuf  = Lexing.from_function read in
-  let input   = Yojson.Basic.(stream_from_lexbuf (init_lexer ()) lexbuf) in
-  let input () = try Some (Stream.next input) with Stream.Failure -> None in
+  let input   = Seq.to_dispenser (Yojson.Basic.(seq_from_lexbuf (init_lexer ()) lexbuf)) in
```
nojb, line 317:
```suggestion
  let input   = Seq.to_dispenser Yojson.Basic.(seq_from_lexbuf (init_lexer ()) lexbuf) in
```

rand00:
I get this error when trying it out:
```bash
> opam pin add https://github.com/Leonidas-from-XIV/merlin.git#yojson-2-compat
This will pin the following packages: dot-merlin-reader, merlin-lib, merlin. Continue? [Y/n] y
[NOTE] Package dot-merlin-reader is already pinned to git+https://github.com/Leonidas-from-XIV/merlin.git#yojson-2-compat (version 4.2).
dot-merlin-reader is now pinned to git+https://github.com/Leonidas-from-XIV/merlin.git#yojson-2-compat (version 4.2)
[NOTE] Package merlin-lib is already pinned to git+https://github.com/Leonidas-from-XIV/merlin.git#yojson-2-compat (version ~dev).
merlin-lib is now pinned to git+https://github.com/Leonidas-from-XIV/merlin.git#yojson-2-compat (version ~dev)
[NOTE] Package merlin is already pinned to git+https://github.com/Leonidas-from-XIV/merlin.git#yojson-2-compat (version 4.5-414).
merlin is now pinned to git+https://github.com/Leonidas-from-XIV/merlin.git#yojson-2-compat (version 4.5-414)

[ERROR] Package conflict!
  * Missing dependency:
    - merlin-lib >= 4.2
    no matching version

[NOTE] Pinning command successful, but your installed packages may be out of sync.
```

voodoos (commented):
Thank you @Leonidas-from-XIV !

voodoos:
@Leonidas-from-XIV the windows CI failed because Yojson 2.0.0 appears to be unavailable. Is that a known issue ?

voodoos:
The fdopen repository was a mirror of opam repository with some patched package specifically for windows.
Now that it is deprecated, I don't know what is the official way to go. Maybe you know, @nojb or @dra27 ?

Leonidas-from-XIV:
According to the blog post

> To ease the transition, I will create a special branch in the near future, which will only contain the compiler packages of the old Windows repository, and can be used in addition to the normal upstream repository (opam repo add ...). However, this branch will not be maintained.

So I think the best would be to use upstream opam-repository + this branch. Unfortunately I was not able to find that branch, it is possible that it never came into existence. In any case, this is more an issue with setup-ocaml, so I opened a ticket here: https://github.com/ocaml/setup-ocaml/issues/547

In theory I assume the Merlin Windows CI run could manually specify to use opam-repository also on Windows, but I don't know if that'll lead to issues with missing Windows packages. It would certainly be nicer to solve it upstream so it works for everyone without local adjustments.

nojb:
> The fdopen repository was a mirror of opam repository with some patched packages specifically for windows. Now that it is deprecated, I don't know what is the official way to go. Maybe you know, @nojb or @dra27 ?

I don't use OPAM much, but https://github.com/fdopen/opam-repository-mingw seems to be maintained (last commit was two days ago).

voodoos:
And it looks like Yojson 2.0 was part of the batch: https://github.com/fdopen/opam-repository-mingw/tree/opam2/packages/yojson/yojson.2.0.0 :-)

Leonidas-from-XIV:
I've squashed the commits to clean up and trigger CI again, so we can see it (hopefully) succeed on the Windows CI.

PR #1476, opened 2022-06-06
Make the prefix argument to `merlin-locate` optional

antalsz:
The `merlin-locate` command had a prefix argument added; it's elisp convention for prefix arguments to be optional, and this is especially helpful as the `merlin-locate` command is user-facing and this provides for backwards compatibility.

This change can also be backported to the 412 and 411 branches.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1575,13 +1575,21 @@ loading"
   (interactive "s> ")
   (merlin--locate-result (merlin-call-locate ident)))
 
-(defun merlin-locate (prefix)
-  "Locate the identifier under point"
+(defun merlin-locate (&optional in-new-window)
+  "Locate the identifier at point.
+
+Whether the result appears in a new window is controlled by
+`merlin-locate-in-new-window', but can be overridden with a
+prefix argument (IN-NEW-WINDOW): if prefixed once with
+\\[universal-argument], the result appears in the current window;
+if prefixed twice with \\[universal-argument], the result appears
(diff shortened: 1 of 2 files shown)


bbatsov:
@antalsz This might also be a good time to document what the 2 prefix arguments actually do. :-) Also - it's Emacs convention to use more descriptive names than `prefix` and to document all arguments in the docstrings.

antalsz:
@voodoos: Changelog entry added, and no need for 411 – I just didn't realize you'd dropped support, so I was looking back anywhere it seemed relevant.

@bbatsov: Thanks for the reminder to make a comprehensive PR rather than a minimal PR, it's appreciated :-)

emacs/merlin.el
```diff
@@ -1575,16 +1575,26 @@ loading"
...
   (cl-letf ((merlin-locate-in-new-window
     (cond
-     ((equal prefix '(4)) 'never)
-     ((equal prefix '(16)) 'always)
+     ((equal in-new-window '(4)) 'never)
+     ((equal in-new-window '(16)) 'always)
      (t merlin-locate-in-new-window))))
     (merlin--locate-result (merlin-call-locate))))
 
+(provide 'merlin)
```
bbatsov, line 1596: This seems like something that was added accidentally.

antalsz: Correct, deleted – thanks for the catch!

voodoos (approved):

PR #1481, opened 2022-06-20
Add imenu items for class methods

mndrix:
Consider the following `test.ml` file:

```ocaml
class foo =
  object
    val x = 42
    method get_x = x
  end
```

The imenu currently includes these items:

- Class: foo
- Value: foo.x

This pull request adds this item as well:

- Method: foo.get_x


mndrix:
I modeled this pull request after the changes made in #1244

voodoos (approved):
Looks good to me, thank you !

Issue #1468, opened 2022-05-17
Locate performance regression for v4.5-414


Zineb-Ada:
We ran this command line `ocamlmerlin single locate -look-for ml -position 171:13 -filename src/irmin-pack/mem/irmin_pack_mem.ml < src/irmin-pack/mem/irmin_pack_mem.ml | jq.` and we noticed performance regression between the old and the new version of Merlin.
As you can see below, oddly the “cpu” of version v4.4-413 is faster than that of v4.5-414. We can see that this lag is especially due to the “query” field which has increased significantly.

`Ocaml=4.14.0; Merlin= v4.5-414`
```json
{
  "class": "return",
  "value": {
    "file": "/Users/tarides/Desktop/Tarides-Ocaml/irmin/src/irmin/irmin.ml",
    "pos": {
      "line": 1,
      "col": 0
    }
  },
  "notifications": [],
  "timing": {
    "clock": 600,
    "cpu": 426,
    "query": 234,
    "pp": 0,
    "reader": 2,
    "ppx": 31,
    "typer": 160,
    "error": 0
  }
}
```

`Ocaml=4.13.1 Merlin= v4.4-413`
```json
{
  "class": "return",
  "value": {
    "file": "/Users/tarides/Desktop/Tarides-Ocaml/irmin/src/irmin/irmin.ml",
    "pos": {
      "line": 1,
      "col": 0
    }
  },
  "notifications": [],
  "timing": {
    "clock": 189,
    "cpu": 166,
    "query": 32,
    "pp": 0,
    "reader": 4,
    "ppx": 15,
    "typer": 114,
    "error": 0
  }
}
```

Zineb-Ada:
Here is an another example of performance regression between ocaml 4.14.0 and 4.12.1 with the same command `ocamlmerlin single locate -look-for ml -position 20:16 -filename ./examples/trees.ml < ./examples/trees .ml| jq .` on the same repository. The CPU difference, as you can see below, is very high.
when we do the command "Go To The Definition" on Irmin_unix on line : 20;
``` ocaml 
module Store = Irmin_unix.Git.FS.KV (Irmin.Contents.String)
```

`Ocaml=4.14.0`

```json 
{
  "class": "return",
  "value": {
    "file": "/Users/tarides/Desktop/Tarides-Ocaml/irmin/src/irmin-unix/irmin_unix.ml",
    "pos": {
      "line": 1,
      "col": 0
    }
  },
  "notifications": [],
  "timing": {
[...]

PR #1485, opened 2022-06-23
Add benchmarks for merlin locate on irmin files

Zineb-Ada:
The purpose of this PR is to benchmark merlin locate using the command `ocamlmerlin single locate -look-for ml -position 23:22 -filename ./examples/irmin_git_store.ml < ./examples/irmin_git_store.ml | jq.` On the  `Irmin repository`.
To do this we added `bench.Dockerfile` and `./tests/bench.sh`.
The `bench.Dockerfile` file contains all the environment to clone `Irmin repository` and launch the necessary installations for `Merlin` and `Irmin`. The `./tests/bench.sh` file allows us to run the benchmark on cases where the `merlin locate` command takes the longest time to run. We were able to identify these cases by running `merlin locate` on each word of each Irmin file that we tested on two different versions of ocaml (4.14.0 and 4.12.1). 
This comparison allowed us to spot a performance regression of merlin that we have reported on `https://github.com/ocaml/merlin/issues/1468`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,4 +18,7 @@ preprocess:
 promote:
 	dune promote
 
-.PHONY: all build dev clean test promote
+bench:
+	cd ../irmin && ../merlin/tests/bench.sh
+
+.PHONY: all build dev clean test promote bench
--- a/bench.Dockerfile
+++ b/bench.Dockerfile
@@ -1,0 +1,18 @@
+# The purpose of this file is to set up the environment necessary to launch benchmarks on current-bench.
+# The goal here is to launch ./tests/bench.sh which launches benchmarks on irmin files
+
+FROM ocaml/opam:debian-10-ocaml-4.14
+RUN sudo apt-get install -y gnuplot-x11 libgmp-dev pkg-config libffi-dev jq
+RUN sudo rm /usr/bin/opam && sudo ln -s /usr/bin/opam-2.1 /usr/bin/opam
+RUN git clone https://github.com/mirage/irmin.git \
+    && cd irmin \
+    && git checkout d23b43e238c7fd417d2bbcab8c9766386c1b0a09 \
+    && opam install --deps-only --with-test . \
+    && opam exec -- dune build @check
+RUN mkdir merlin
+WORKDIR merlin
+COPY --chown=opam:opam ./*.opam ./
+RUN opam pin add -ny --with-version=dev . \
+    && opam install --deps-only .
+COPY --chown=opam:opam . ./
+RUN opam install .
--- a/tests/bench.sh
+++ b/tests/bench.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# the purpose of this file is to run the ocamlmerlin locate, type-enclosing and occurrences commands
+# on the cases that the commands take the most time on all Irmin repository
+# and make sure to match the json of merlin to that of current-bench, in order to launch benchmarks.
+
+function prbench {
+     metricname="$2:$3 $4"
+     jq '{"results": [{"name": "'$1'", "metrics": [
+               {"name": "'"$metricname"'/clock", "value": .timing.clock, "units": "ms"},
+               {"name": "'"$metricname"'/typer", "value": .timing.typer, "units": "ms"},
+               {"name": "'"$metricname"'/ppx", "value": .timing.ppx, "units": "ms"},
+               {"name": "'"$metricname"'/query", "value": .timing.query, "units": "ms"},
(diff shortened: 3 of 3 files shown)


bench.Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+    && git checkout d23b43e238c7fd417d2bbcab8c9766386c1b0a09 \
+    && opam install --deps-only --with-test . \
+    && opam exec -- dune build @check
+RUN mkdir merlin
+WORKDIR merlin
+COPY --chown=opam:opam ./*.opam ./
+RUN opam pin add -ny --with-version=dev . \
+    && opam install --deps-only .
+COPY --chown=opam:opam . ./
+RUN opam install .
```
voodoos, line 15: Missing new line

tests/bench.sh
```diff
@@ -0,0 +1,14 @@
...
+     | jq '{"results": [{"name": "locate", "metrics": [{"name": "'$1':'$2' '$3'", "value": .timing.query, "units": "ms"}]}]}'
+}
+
+locate 23 16 ./examples/irmin_git_store.ml
+locate 118 21 ./examples/process.ml
+locate 49 20 ./examples/push.ml
+locate 34 16 ./examples/deploy.ml
+locate 111 9 ./examples/process.ml
+locate 43 26 ./examples/push.ml
+locate 70 22 ./test/irmin-graphql/common.ml
```
voodoos, line 14: Missing new line

Makefile
```diff
@@ -18,4 +18,7 @@ preprocess:
 promote:
 	dune promote
 
-.PHONY: all build dev clean test promote
+bench:
+	cd ../irmin && ../merlin/tests/bench.sh
+
+.PHONY: all build dev clean test promote bench
```
voodoos, line 24: Missing newline

tests/bench.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+function locate {
+     sed -n "$1"p "$3"
```
voodoos, line 3: I guess this prints the line on which Merlin is called ?
Where will this printing be visible ? In some logs ?

voodoos (commented):
I like the idea to add some benchmarks to the repository. But I have some questions:

It is a bit wasteful to have these run for every PR and commit on master. Do you have an ETA on some feature to prevent the benchs from running when it is not needed (for example with github actions we only run the Emacs lint CI when the emacs files change) ? Can we be of any help to implement this ? (@art-w)

Also, what should we do to allow ocaml-bench to run on this repository ? and does @trefis have enough rights on the ocaml organisation to do it ? If we do activate the benchmarks will they trigger right away for this PR ?

tests/bench.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
voodoos, line 1: Maybe we could add a commentary here explained the purpose of this file (ie the fact that it is to be ran by ocaml-bench).

bench.Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ocaml/opam:debian-10-ocaml-4.14
```
voodoos, line 1: Maybe we could add a short commentary here explained the purpose of this file.

tests/bench.sh
Zineb-Ada: yes exactly, and we can see this print in the benchmark result "as you can see in this link http://autumn.ocamllabs.io:8081/job/2022-06-23/125306-ocluster-build-9d17b9#L283-L296" and on the terminal.

Zineb-Ada:
In the last commit we added benchmarks for `ocamlmerlin type-enclosing command` following the same methodology as `locate command`.
To launch the benchmarks, you must install the current-bench application "https://github.com/marketplace/ocaml-benchmarks" on the Merlin repository. The benchmarks will launch on every PR.

art-w:
@voodoos > We'll soon be able to run the benchmarks only for PRs that have been marked with a specific github label: for example "bench", but you just have to let us know which tag name works for you? Currently it's a bit hard to watch for a list of modified files but we might get back to you on this in the future :)

Regarding the CI setup, you need someone with enough administrative power over the `ocaml` organization to install the `ocaml-benchmarks` GH app at https://github.com/marketplace/ocaml-benchmarks , then enable it only on the `merlin` repository (there's a dropdown to pick the repo right after the "installation").

voodoos:
I would say something like `need-bench` ?

ElectreAAS:
I know this PR is old, but current-bench has just merged the PR allowing us to only benchmark PRs with a specific label ([https://github.com/ocurrent/current-bench/pull/371]).
@voodoos should I go ahead and set `need-bench` as the necessary label?

PR #1488, opened 2022-06-27
Use weak strategy when reducing shapes

voodoos:
This prevent unnecessary computations when reducing shapes in Merlin.
For example, in the current released version, Merlin would completely reduce the whole shape of the module `Irmin_unix` when performing a jump-to-definition at `Irmin_u|nix.Git.Some(Stuff)` which is wasteful since we only need the UID attributed to the top-level structure of the shape.

This patch leads to a huge speed up in this case. Improving #1468 a lot.

cc @gasche

### DIFF
--- a/src/ocaml/typing/shape.mli
+++ b/src/ocaml/typing/shape.mli
@@ -152,6 +152,7 @@ module Make_reduce(Context : sig
     val find_shape : env -> Ident.t -> t
   end) : sig
   val reduce : Context.env -> t -> t
+  val weak_reduce : Context.env -> t -> t
 end
 
 val local_reduce : t -> t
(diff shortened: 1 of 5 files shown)


src/ocaml/typing/shape.mli
```diff
@@ -151,7 +151,7 @@ module Make_reduce(Context : sig
 
     val find_shape : env -> Ident.t -> t
   end) : sig
-  val reduce : Context.env -> t -> t
+  val reduce : ?weak:bool -> Context.env -> t -> t
```
gasche, line 154: I would rather expose two functions, `reduce` (or `strong_reduce`) and `weak_reduce`, because they do fairly different things.

voodoos: Thanks for having a look!

PR #1494, opened 2022-06-29
Fix duplicated prefix path in imenu entries

bcc32:
The entry already has the prefix path incorporated, so adding the
prefix again results in duplicate prefixes (in fact, n^2 module path
components!).


voodoos:
I am not able to reproduce the duplicated path.
What I did was:
- open a file
- call `M-x merlin-use-merlin-imenu`
- call `M-x imenu`
- press `<tab>` to show the completions and I don't see any duplication

voodoos (approved):
Thank you !

Prefix is indeed already added in `merlin-imenu-create-entry`

PR #1503, opened 2022-07-25
Replace custom "holes" AST nodes by extensions to restore binary compatibility

voodoos:
There is currently an issue with Merlin's new typed holes and ppxes.
Since these holes were added as new AST node ppxes crash in there presence. 

This PR removes these new parsetree nodes and restore the old extension-based ones.
We still keep the typedtree nodes that improve the typing if these holes.

I added a test with a custom ppx that shows the error. However this requires to have `ppxlib` as a test-only dependency for the `merlin` package. @trefis I know you were wary about doing that, but it really looks like the easiest way to test ppxes in the the testsuite, do you see another option ?

cc @lpw25


src/ocaml/preprocess/parser_raw.mly
```diff
@@ -1435,7 +1435,8 @@ module_expr [@recovery default_module_expr ()]:
         { Pmod_extension ex }
     | (* A hole. *)
       UNDERSCORE
-        { Pmod_hole }
+        { let id = mkrhs "merlin.hole" $loc in
+          Pmod_extension (id, PStr []) }
```
antalsz, line 1439: This could get factored out to avoid typos

src/ocaml/typing/untypeast.ml
```diff
@@ -517,7 +517,9 @@ let expression sub exp =
                              ])
     | Texp_open (od, exp) ->
         Pexp_open (sub.open_declaration sub od, sub.expr sub exp)
-    | Texp_hole -> Pexp_hole
+    | Texp_hole ->
+        let id = Location.mkloc "merlin.hole" loc in
+        Pexp_extension (id, PStr [])
```
antalsz, line 522: We've imported `Ast_helpers`, so this can just be `Exp.hole ()`, I believe

src/ocaml/typing/untypeast.ml
```diff
@@ -679,7 +681,9 @@ let module_expr (sub : mapper) mexpr =
           | Tmod_unpack (exp, _pack) ->
               Pmod_unpack (sub.expr sub exp)
               (* TODO , sub.package_type sub pack) *)
-          | Tmod_hole -> Pmod_hole
+          | Tmod_hole ->
+              let id = Location.mkloc "merlin.hole" loc in
+              Pmod_extension (id, PStr [])
```
antalsz, line 686: We've imported `Ast_helpers`, so this can just be `Mod.hole ()`, I believe

tests/test-dirs/with-ppx.t/run.t
```diff
@@ -0,0 +1,29 @@
+  $ cat >main.ml <<EOF
+  > match Some 3 with
+  > | None -> ()
+  > | Some _ -> print_int [%get_int 42]
+  > EOF
```
antalsz, line 5: I think this test should also include an actual typed hole; perhaps `_` instead of `()`?

antalsz (changes requested):
Thanks for making this change, it looks great!  I went ahead and gave it some review, since we're excited to roll it out internally; I just had a couple small comments.

src/ocaml/typing/untypeast.ml
voodoos: `Exp.hole ()` has type `expression` but `expression_desc` is expected here so we cannot use it. It is probably enough to only factor the `merlin.hole` name out.

src/ocaml/typing/untypeast.ml
voodoos: Same.

PR #1505, opened 2022-07-28
Improvement: jumping to a label's precise declaration.

voodoos:
When querying jump to declaration of a record's label `foo.ba|r` merlin would only jump to the beginning of the corresponding record type declaration.

This is due to the uid's of labels not being registered by the compiler in the `Env.uid_to_loc` table.

This PR adds these uid registration to `typedecl.ml`. This means that the fix will only work for the current buffer and we would need to be upstreamed for it to work when jumping to external compilation units. 

cc @rgrinberg

### DIFF
--- a/tests/test-dirs/locate/context-detection/cd-label.t/run.t
+++ b/tests/test-dirs/locate/context-detection/cd-label.t/run.t
@@ -25,3 +25,49 @@
     "notifications": []
   }
 
+  $ cat >record.ml <<EOF
+  > type t = { bar: int}
+  > let foo = { bar = 42 }
+  > let () = print_int foo.bar
+  > EOF
+
+  $ $MERLIN single locate -look-for mli -position 3:24 \
+  > -filename ./record.ml < ./record.ml | jq '.value'
+  {
+    "file": "$TESTCASE_ROOT/record.ml",
+    "pos": {
+      "line": 1,
+      "col": 11
+    }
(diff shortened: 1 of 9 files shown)


voodoos:
While making a test for "locating a label in another CU" I noticed that it was even more broken: after finding the Uid fails it should "fallback to the declaration's node's location" and it was not the case. This is fixed in the latest commit.

rgrinberg:
Looks good to me.

I guess the same problem exists for variant constructors?

voodoos:
> I guess the same problem exists for variant constructors?

I added a test and it worked as expected for constructors. However it relied on  the "fallback" so I did the same change as for labels.

However I tested the polymorphic variants' tags and Merlin doesn't even try to locate these. But this will probably be the subject of another PR.

tests/test-dirs/locate/context-detection/cd-label.t/run.t
```diff
@@ -25,3 +25,47 @@
     "notifications": []
   }
 
+  $ cat >record.ml <<EOF
+  > type t = { bar: int}
+  > let foo = { bar = 42 }
+  > let () = print_int foo.bar
```
rgrinberg, line 31: The EOF for this test isn't terminated.

rgrinberg:
@voodoos any ETA on when this will be merged?

PR #1509, opened 2022-08-23
Recovery caused nested outline

voodoos:
Fixes issue https://github.com/ocaml/ocaml-lsp/issues/841 reported for ocaml-lsp-server.

When typing of a module structure item failed the saved types from the structure were wrongly added to that sub module causing outline to misbehave:

```
  $ cat >test.ml <<EOF
  > module Make = struct
  >   module A = struct end
  >   module B = C.C1
  >   module D = struct end
  > end
  > EOF

FIXME: A is not a child of B ! This is likely due to the typing recovery.
  $ $MERLIN single outline -filename test.ml <test.ml
  {
    "class": "return",
    "value": [
        "name": "Make",
        "children": [
            "name": "D",
            "children": [],
          },
            "name": "B",
            "children": [
                "name": "A",
                "children": [],
              }
            ],
          },
            "name": "A",
            "children": [],
          }
        ],
      }
    ],
  }
```

### DIFF
--- a/src/ocaml/typing/typemod.ml
+++ b/src/ocaml/typing/typemod.ml
@@ -2231,6 +2231,10 @@ let simplify_app_summary app_view =
   | false, None -> Includemod.Error.Anonymous, mty
 
 let rec type_module ?(alias=false) sttn funct_body anchor env smod =
+  (* Merlin: when we start typing a module we don't want to include potential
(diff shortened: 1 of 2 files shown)


let-def (approved):
The patch looks good as it is.

PR #1511, opened 2022-09-09
vim: Load the plugin on the fly if necessary

shym:
Updates to 'runtimepath' don’t load newly-available plugins. So adding merlin to 'runtimepath' during a session (instead of your .vimrc), you end up with an exception in merlin#FindBinary next time you open an OCaml file because none of s:c.ocamlmerlin_path, s:c.merlin_home and s:c.merlin_parent are set.


voodoos:
Thank you for taking the time of opening a PR !
 
These changes looks ok to me, but I am not an expert. @trefis of @let-def  could you double-check ?

trefis (approved):
LGTM.

voodoos:
@shym could you add a changelog entry ?

shym:
I added the changelog entry in the same commit. Tell me whether it’d better be a separate commit.

PR #1516, opened 2022-10-20
fix #1513: variant constructor docs 'only_after'

### LINKED ISSUE #1513: Doc comments are associated with the wrong constructors
**Steps to rero**

Create foo.ml, with these contents:

```ocaml
type t =
    | A
    (**A Comment *)
    | B
    (**B Comment *)
    | C
```

Then open the file in VSCode, using the Ocaml Platform extension (tested with v1.10.7), which I think uses Merlin v4.6-413 via ocaml-lsp-server.

Hover over each constructor

**Actual behavior**

Here are the hovers ocaml-lsp-server provides for each constructor, respectively:
- A: "A comment"
- B: "A comment B comment"
- C: "B comment"

**Expected behavior**

Here are the hovers I'd expect ocaml-lsp-server to provide for each constructor, respectively:
- A: "A comment"
- B: "B comment"
- C: "C comment"

This expectation is based on the OCaml manual https://v2.ocaml.org/manual/doccomments.html#ss:label-comments:

> Comments which appear immediately after a labelled argument, record field, variant constructor, object method or polymorphic variant constructor are are converted into [attribute](https://v2.ocaml.org/manual/attributes.html#attribute)s.

**version**

I used `ocaml-lsp-server.1.13.2~5.0preview` which I think uses **Merlin v4.6-413**

mheiber:
Fix #1513  Group doc comments for variant constructors only with only the constructor before

Reference:
https://v2.ocaml.org/manual/doccomments.html#ss:label-comments

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -1015,8 +1015,16 @@ let get_doc ~config ~env ~local_defs ~comments ~pos =
               Location.print_loc l);
         Format.fprintf fmt "]\n"
       );
+    let (_, deepest_before) = List.hd @@ Mbrowse.deepest_before loc.loc_start [browse] in
+    (* based on https://v2.ocaml.org/manual/doccomments.html#ss:label-comments: *)
+    let after_only = begin match deepest_before with
+      | Browse_raw.Constructor_declaration _ -> true
+      (* The remaining `true` cases are currently not reachable *)
+      | Label_declaration _ | Record_field _ | Row_field _  -> true
+      | _ -> false
+    end in
     begin match
-      Ocamldoc.associate_comment comments loc !last_location
+      Ocamldoc.associate_comment ~after_only comments loc !last_location
     with
     | None, _     -> `No_documentation
     | Some doc, _ -> `Found doc
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ undefined
     - Do not change temporarily Merlin's cwd when starting a PPX (#1521)
     - Fix a parsing issue when declaring the `(??)` custom prefix operator.
       (#1507, fixes #1506)
+    - Fix variant constructors' comments grouping (#1516, @mheiber, fixes #1513)
   + editor modes
     - vim: load the plugin when necessary if it wasn’t loaded before (#1511)
   + test suite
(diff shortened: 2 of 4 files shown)


src/analysis/locate.ml
```diff
@@ -1016,7 +1016,16 @@ let get_doc ~config ~env ~local_defs ~comments ~pos =
         Format.fprintf fmt "]\n"
       );
     begin match
-      Ocamldoc.associate_comment comments loc !last_location
+      let (_, deepest_before) = List.hd @@ Mbrowse.deepest_before loc.loc_start [browse] in
+      (* based on https://v2.ocaml.org/manual/doccomments.html#ss:label-comments: *)
+      let after_only = begin match deepest_before with
+        | Browse_raw.Constructor_declaration _ -> true
+        (* The remaining `true` cases are *currently* inert, since iuc we don't find
+          doc comments for these nodes yet *)
```
voodoos, line 1024: Do you mean that Merlin does not show such comments or that they are not part of the specification ?
Could you add test cases showing the current behavior ?

src/analysis/locate.ml
```diff
@@ -1016,7 +1016,16 @@ let get_doc ~config ~env ~local_defs ~comments ~pos =
         Format.fprintf fmt "]\n"
       );
     begin match
-      Ocamldoc.associate_comment comments loc !last_location
+      let (_, deepest_before) = List.hd @@ Mbrowse.deepest_before loc.loc_start [browse] in
```
voodoos, line 1019: Nitpick: that's a lot going on in the `match ... with` argument, could you give it a name and then match on this ?
Or just stay closer to  the current code with something like:
```ocaml
let after_only = 
  .....
in 
begin match
  Ocamldoc.associate_comment ~after_only comments loc !last_location
with

voodoos (commented):
LGTM, thanks for having taken the time :-) 

Just a couple of minor comments

src/analysis/locate.ml
mheiber: > Do you mean that Merlin does not show such comments or that they are not part of the specification ?

As far as I can tell, Merlin does not show such comments, at least in this codepath

> Could you add test cases showing the current behavior ?

Good idea, will do

mheiber: I added these as separate test cases here: https://github.com/ocaml/merlin/pull/1526

PR #1521, opened 2022-11-03
Stop using chdir for ppx

### LINKED ISSUE #1420: Request: do not use chdir for ppx
This isn't something user facing, rather a matter of making it easy to use merlin correctly in ocamllsp. You might be aware that ocamllsp runs the merlin pipeline in a dedicated thread to make sure the server stays responsive while merlin is busy. That works well until preprocessing is involved. `chdir` is not threadsafe and therefore merlin's use of it for running pp/ppx breaks ocamllsp in some subtle (and painfully hard to debug) ways. Thus my humble request from the merlin team is to accept a patch that removes the use of chdir.

I haven't yet written the patch, but I'm thinking that it can be done by using the [spawn](https://github.com/janestreet/spawn) library. This library is successfully used by dune and ocamllsp itself so it is well tested. It's also light on the dependencies.

The only problem I could imagine is that merlin has some code for launching processes on Windows. So perhaps we can keep the current code for windows and add a cwd argument? Or improve spawn in a way that makes this custom code unnecessary? cc @MisterDA

voodoos:
Fixes #1420 

This is a cherry-pick from @rgrinberg's fork of Merlin for LSP.

@MisterDA could you test that this patch works well on Windows ? You need a project with a ppx.

### DIFF
--- a/src/kernel/mppx.ml
+++ b/src/kernel/mppx.ml
@@ -2,16 +2,6 @@ open Mconfig
 
 let {Logger. log} = Logger.for_section "Mppx"
 
-let change_directory dir =
-  log ~title:"changing_directory" "%s" dir;
-  match Sys.chdir dir with
-  | () -> true
-  | exception exn ->
-    log ~title:"changing directory"
-      "change_directory %S failed with %a" dir Logger.exn exn;
-    false
-
-
 let with_include_dir path f =
   let saved = !Clflags.include_dirs in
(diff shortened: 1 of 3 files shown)


let-def (approved):

MisterDA:
Just tested that merlin with this patch works correctly on a ppx project on Windows. It works, thanks.

PR #1522, opened 2022-11-04
Fix locate on module without implementation

### LINKED ISSUE #1519: locate fails with "didn't manage to find <Module>" but Module.<tab> list all the values inside the module correctly
I've been in this case so many times in my projects and I'm not too sure how to reproduce it. Any ideas?
I use the locate feature quite frequently but it's very finicky to get it working half the time.

voodoos:
Should fix #1519 

@kate could you check if this fixes both your issues or just the first one ?


kit-ty-kate:
Ah, I hadn't seen this. You pinged someone else.

voodoos:
Oups :-D

PR #1535, opened 2022-11-25
Update to latest merlin (4.7)

ulugbekna:
- Drops several commits that are no longer needed because of changes in the upstream

  - drop bf63d90d16ca13ff95a4ef7a813b4a7322ae5455
     there's now `Lib_config.program_name` that can be set to `OCaml LSP`

  - 9a3c4e9b2d666018f936fcd97ec17f861ce2ed4c disappears during rebase 

  - drop 2bb9e14f3fc8059b795e0b78464e687ad47af6fc 
    done upstream

  - drop 87e952dac24d1067348d9de08b6e69793d1c0e57
    `parenthesize_name` now lives in `Misc` module

  - drop 953a5354349157be4cb2f0a42eae2db875cc0916
    done in ocamllsp itself

Note: when upgrading to latest merlin in ocamllsp (I plan to do it myself if ok), need to keep in mind to adapt to these changes

### DIFF
--- a/src/dot-protocol/merlin_dot_protocol.ml
+++ b/src/dot-protocol/merlin_dot_protocol.ml
@@ -63,7 +63,7 @@ end
 type directive = Directive.Processed.t
 
 module Sexp = struct
-  type t = Atom of string | List of t list
+  type t = Csexp.t = Atom of string | List of t list
 
   let atoms_of_strings = List.map ~f:(fun s -> Atom s)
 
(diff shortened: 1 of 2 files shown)


nojb:
Is this on the right repository?

ulugbekna:
> Is this on the right repository?

Ah, no, it's not. Thanks and sorry for noise!

PR #1537, opened 2022-12-01
Locate constructor exact positions

voodoos:
Currently Merlin will locate the type of a constructor, not the constructor itself.
This is much improved in that PR.


voodoos:
After discussing this issue with @pitag-ha we realized that it only answers the problem for type's declarations, not definitions. I am turning this into a draft while waiting for a more robust answer.

PR #1544, opened 2022-12-07
Accept the `-safe-string` flag but make it a no-op

### LINKED ISSUE #1518: "unknown flag -safe-string" on OCaml 5.0
Using this project at this particular commit: https://github.com/kit-ty-kate/mastodon-archive-viewer/commit/bd94f22fc43d19b2b2f86a6418e7a0126b755f13 using OCaml 5.0.0~beta1
Then opening `src/main.ml` using emacs and reloading merlin will show the following error:
```
unknown flag -safe-string (0 pending errors, use C-c C-x to jump)
```
However the command is still around and works perfectly fine (it's no noop)

voodoos:
This should fix #1518 

@kit-ty-kate do you no if other configuration options have been removed from OCaml 5 but are still accepted by the compiler as a no-op ?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+merlin 4.7.1
+==========
+Thu Dec 13 12:49:42 CEST 2022
+
+  + merlin binary
+    - Restore compatibility with the compielr command line by accepting the
+      `-safe-string` flag as a no-op instead of rejecting it. (#1544, fixes
+      #1518)
+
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
As far as I know it's the only one but i haven't kept track of this too much. @dra27 do you know?

kit-ty-kate:
Yes I think so.

Issue #1543, opened 2022-12-07
`merlin-locate` fails in the presence of certain type errors


antalsz:
We've had some bugs about `merlin-locate` (jump to definition/implementation) reported internally, which are all about it not succeeding when there's some sort of type error present in the code.  We're not sure which of them come from the same root cause, but they all manifest very similarly; we've divided them into three cases below.  While Merlin's behavior in the presence of errors isn't super crisply defined, these do seem like bugs; in all cases, we see the following behavior.

1. We have a variable, ``let problem = `Problem``.

2. This variable is used in a way that provokes a type error.

3. If we ask Merlin for the type of this use of `problem` (`merlin-type-enclosing`), it happily and correctly reports "``[> `Problem]``".

4. If we attempt to use `merlin-locate` to jump to the definition or interface (i.e., with `merlin-locate-preference` set to either `'ml` or `'intf`) from the use of `problem`, we get the error message "`Not in environment 'problem'`" and don't jump anywhere.

The combination of (3) and (4) makes this seem like a genuine bug, as does the fragility exhibited by bugs 2 and 3 (which see), which require precise surrounding contexts to appear.

The bugs are as follows.

### Bug 1: In the argument to an incorrectly-refuted `match` expression

In this case, `merlin-locate` fails if a value of a non-empty type is being used as the argument to a refutable `match` expression.

```
module ERROR_locate_from_incorrectly_refutable_match = struct
  (* We expect to end up here.
   *  vvvvvvv *)
  let problem = `Problem

  let () =
    (* We jump to defn or intf from here:
     *    vvvvvvv *)
    match problem with
    | _ -> .
  ;;

  (* We get the message "Not in environment 'problem'" and go nowhere. *)
end
```

### Bug 2: Inside applications applied to a `fun` literal used in a non-function context

In this case, `merlin-locate` fails inside a `fun` function literal, but only if that function literal isn't being used at a function type.  This can be due to inference, due to a surrounding function call, or even due to a type annotation.

```
module ERROR_locate_from_inside_function_literal_used_as_non_function = struct
  (* We expect to end up here:
   *  vvvvvvv *)
  let problem = `Problem

  (* We jump to the definition or interface from here...:
   *                 vvvvvvv *)
  let () = fun () -> problem

  (* ...or from here...:
   *                                vvvvvvv *)
  let _ : int = Int.succ (fun () -> problem)
  
  (* ...or from here:
   *                                      vvvvvvv *)
[...]

PR #1546, opened 2022-12-20
Locate fails in ill-typed expressions

voodoos:
This PR tracks the progress on fixing issue #1543.

- [x] Bug 1
- [x] Bug 2
- [ ] Bug 3

Merlin tries to recover from typing errors, but large part of the typedtree are sometimes dropped. Fixing the error cases documented in #1543 requires improvement to the typing recovery.


voodoos:
Note: the current fixes simply ignore the type errors but this might not be sound. Maybe the "saved-types" mechanism would be a better way to go.

let-def (approved):

PR #1549, opened 2022-12-22
Broken `eq` comparison in merlin.el

mattiase:
Whether a string literal is `eq` to anything else is undefined and, in this case, very unlikely: use `equal` instead. This silences a (well-justified) warning in Emacs 30, although the comparison may have been subsumed by the additional use of `merlin--error-warning-p` in practice.


voodoos:
Thank you !

Issue #1553, opened 2022-12-26
OCaml 5.0 support


edwintorok:
I got this error from ocaml-lsp-server (which vendors merlin):
```
[...]

PR #1554, opened 2022-12-26
Recognize OCaml 5.0 cmi magic number

edwintorok:
And add an assertion that we can recognize our own magic number that we got compiled with.

Fixes https://github.com/ocaml/merlin/issues/1553

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+merlin 4.7.2
+============
+undefined
+
+  + merlin binary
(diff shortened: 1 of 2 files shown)


voodoos:
Thank you !
This function is used only to provide a nice error message when the project was not built with 5.0.
Could you add a changelog entry ? I will have a look a the (unrelated) CI failures.

edwintorok:
Added a new entry for 4.7.2 (assuming that is the next version you want to release).

voodoos (approved):

PR #1556, opened 2022-12-29
Add missing ocamlc flags to the ignore list

voodoos:
These flags where added to OCaml 4.14 but are not accepted by Merlin.

Note for the 5.0 backport: the flag `-force-tmc` was removed in 5.0.

CC @antalsz


voodoos:
Just noticed that `-force-tmc` was dropped in 4.14.1, so let's not bother with it.

PR #1563, opened 2023-02-02
Jump to declaration through aliases

voodoos:
This PR implement a suggestion by @lpw25 to traverse module aliases when jumping to declaration.

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -353,48 +353,6 @@ let uid_of_path ~config ~env ~ml_or_mli ~decl_uid path ns =
       Logger.fmt (fun fmt -> Shape.print fmt r);
     r.uid
 
-(** [module_aliasing] iterates on a typedtree to check if the provided uid
-  corresponds to a module alias. If it does the function returns the uid of the
-  aliased module. If not it returns None.
-  The intended use of this function is to traverse dune-generated aliases. *)
-let module_aliasing ~(bin_annots : Cmt_format.binary_annots) uid  =
-  let exception Found of Path.t * Env.t in
-  let iterator env = { Tast_iterator.default_iterator with
-    module_binding = (fun sub mb ->
-      begin match mb with
-      | { mb_id = Some id; mb_expr = { mod_desc = Tmod_ident (path, _); _ }; _ }
-        ->
-          let md = Env.find_module (Pident id) env in
-          if Shape.Uid.equal uid md.md_uid then
-            raise (Found (path, env))
-      | _ -> () end;
-      Tast_iterator.default_iterator.module_binding sub mb)
-    }
-  in
-  try
-    begin match bin_annots with
-    | Interface s ->
-        let sig_final_env = Envaux.env_of_only_summary s.sig_final_env in
-        let iterator = iterator sig_final_env in
-        iterator.signature iterator { s with sig_final_env }
-    | Implementation str ->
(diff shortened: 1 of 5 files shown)


src/analysis/locate.ml
```diff
@@ -310,6 +310,16 @@ let load_cmt ~config comp_unit ml_or_mli =
...
 
+let find_declaration_uid ~env ~decl_uid path =
+  let rec non_alias_declaration_uid path =
+    let md = Env.find_module path env in
+    match md.md_type with
+    | Mty_ident _ | Mty_signature _ | Mty_functor _ | Mty_for_hole -> md.md_uid
+    | Mty_alias path -> non_alias_declaration_uid path
+  in
+  try non_alias_declaration_uid path
+with Not_found -> decl_uid
```
trefis, line 321: IIUC with this implementation if we have a chain of aliases and we fail to resolve one of them at any step (could be the very last), we'll just jump to the initial declaration.

Perhaps it might be more useful to stop on the last step we managed to reach? Otherwise people would have to call `jump-to-decl` over and over again to reach that point...

Something like this:
```ocaml
  let rec non_alias_declaration_uid fallback_uid path =
    match Env.find_module path env with
    | { md_type = Mty_alias path; md_uid; _ } -> non_alias_declaration_uid md_uid path
    | { md_type = (Mty_ident _ | ...); md_uid; _ } -> md_uid
    | exception Not_found -> fallback_uid
```
?

src/analysis/locate.ml
```diff
@@ -310,6 +310,16 @@ let load_cmt ~config comp_unit ml_or_mli =
       Ok (source_file, cmt_infos)
   | None -> Error ()
 
+let find_declaration_uid ~env ~decl_uid path =
```
trefis, line 313: I find that name to be quite vague, especially since we give `decl_uid` as an input.
How about renaming it to `scrape_alias` or something of the sort?

tests/test-dirs/locate/module-aliases.t/run.t
```diff
@@ -65,12 +65,11 @@ Jump to an aliased module `A|.f`:
   }
 
 Jump to the declaration of an aliased module `A|.f`.
-With the new shape implementation it is expected that we jump to the 
-declaration of the alias and not to the aliased module itself.
+The alias is traversed.
```
trefis, line 68: Seeing this, it seems like we had _decided_ to have the "stop on alias" behavior. Perhaps (thanks to the move to dune) the whole community has switched to using module aliases extensively and the behavior is now more annoying than anything.
But I do wonder if we could not perhaps add a flag to keep the old behavior to stop on aliases for declarations... not sure whether it's worth it. I guess it will depend how ugly the change required to `locate.ml` are?

src/analysis/locate.ml
voodoos: Done in 31bbe7408e0767d4334955198fdb9ec377af5fe8

src/analysis/locate.ml
voodoos: Done in 5a3d79172a99d731aa257d21ea66616bc2eb6e67

tests/test-dirs/locate/module-aliases.t/run.t
voodoos: I feel like the most often expected behavior is to traverse the aliases. Let's keep it that way for now, and if the need to add a configuration flag arises we will do it at that time.

trefis (approved):

PR #1569, opened 2023-02-20
ocamlmerlin.c: chdir to root also on Windows to continue working if cwd is deleted

### LINKED ISSUE #1474: [BUG] Running `ocamlmerlin-server.exe` does not allow moving directory under Windows
After one starts using `merlin` to edit a file on Emacs under Windows it is not possible to move the containing directory around, which is quite inconvenient. Closing `emacs` does not solve the problem because the `merlin` process continues to run in the background. The only solution is to kill the `merlin` process running in the background.

image

- Why not shut down the `merlin` background process when closing Emacs?
- Typically this problems happens because a program (Merlin in this case) keeps open handles to the directory or some file within it; if this is indeed the case, perhaps this situation can be avoided? For example even if one is editing a file in a directory with Emacs, that does not keep the containing directory from being renamed.

**To reproduce**

1. `mkdir foo`
2. `emacs foo/foo.ml`
3. Perform some operation that invokes `merlin` on the `foo.ml` buffer
4. `mv foo foo2` in the terminal (fails)

nojb:
On Windows, the process `ocamlmerlin-server.exe` is launched with the directory of the current source file as current working directory. This causes problems if one tries to move or delete that directory.

To avoid this problem, this PR makes sure to `chdir` to a different directory (TMPDIR) before launching `ocamlmerlin-server.exe`. Note that the same logic already exists for Unix: https://github.com/ocaml/merlin/blob/f1b877cd20d5c99b24bb8d051a63088488d26846/src/frontend/ocamlmerlin/ocamlmerlin.c#L314-L317

cc @dra27 @MisterDA 

Fixes #1474

### DIFF
--- a/src/frontend/ocamlmerlin/new/new_merlin.ml
+++ b/src/frontend/ocamlmerlin/new/new_merlin.ml
@@ -147,10 +147,15 @@ let run = function
         prerr_endline ("Exception: " ^ Printexc.to_string exn);
         1
 
-let restore_cwd old f args =
-  match f args with
-  | exception e -> Sys.chdir old; raise e
-  | res -> Sys.chdir old; res
+let with_wd ~wd ~old_wd f args =
+  match Sys.chdir wd with
+  | () ->
+    log ~title:"run" "changed directory to %S (old wd: %S)" wd old_wd;
+    Fun.protect ~finally:(fun () -> Sys.chdir old_wd) (fun () -> f args)
+  | exception Sys_error _ ->
+    log ~title:"run" "cannot change working directory to %S (old wd: %S)"
+      wd old_wd;
+    f args
 
 let run ~new_env wd args =
   begin match new_env with
@@ -159,15 +164,13 @@ let run ~new_env wd args =
     Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()))
   | None -> () end;
   let old_wd = Sys.getcwd () in
-  let wd_msg, run = match wd with
-    | None -> "No working directory specified", run
-    | Some wd ->
-      try Sys.chdir wd; Printf.sprintf "changed directory to %S" wd, restore_cwd old_wd run
-      with _ -> Printf.sprintf "cannot change working directory to %S" wd, run
+  let run args () = match wd with
+    | Some wd -> with_wd ~wd ~old_wd run args
+    | None ->
+      log ~title:"run" "No working directory specified (old wd: %S)" old_wd;
+      run args
   in
   let `Log_file_path log_file, `Log_sections sections =
     Log_info.get ()
   in
-  Logger.with_log_file log_file ~sections @@ fun () ->
-  log ~title:"run" "%s (old wd: %S)" wd_msg old_wd;
-  run args
+  Logger.with_log_file log_file ~sections @@ run args
--- a/src/frontend/ocamlmerlin/new/new_merlin.ml
+++ b/src/frontend/ocamlmerlin/new/new_merlin.ml
@@ -147,6 +147,11 @@ let run = function
         prerr_endline ("Exception: " ^ Printexc.to_string exn);
         1
 
+let restore_cwd old f args =
+  match f args with
+  | exception e -> Sys.chdir old; raise e
+  | res -> Sys.chdir old; res
+
 let run ~new_env wd args =
   begin match new_env with
   | Some env ->
@@ -154,11 +159,11 @@ let run ~new_env wd args =
     Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()))
   | None -> () end;
   let old_wd = Sys.getcwd () in
-  let wd_msg = match wd with
-    | None -> "No working directory specified"
+  let wd_msg, run = match wd with
+    | None -> "No working directory specified", run
     | Some wd ->
-      try Sys.chdir wd; Printf.sprintf "changed directory to %S" wd
-      with _ -> Printf.sprintf "cannot change working directory to %S" wd
+      try Sys.chdir wd; Printf.sprintf "changed directory to %S" wd, restore_cwd old_wd run
+      with _ -> Printf.sprintf "cannot change working directory to %S" wd, run
   in
   let `Log_file_path log_file, `Log_sections sections =
     Log_info.get ()
--- a/src/frontend/ocamlmerlin/new/new_merlin.ml
+++ b/src/frontend/ocamlmerlin/new/new_merlin.ml
@@ -153,6 +153,7 @@ let run ~new_env wd args =
     Os_ipc.merlin_set_environ env;
     Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()))
   | None -> () end;
+  let old_wd = Sys.getcwd () in
   let wd_msg = match wd with
     | None -> "No working directory specified"
     | Some wd ->
@@ -163,5 +164,5 @@ let run ~new_env wd args =
     Log_info.get ()
   in
   Logger.with_log_file log_file ~sections @@ fun () ->
-  log ~title:"run" "%s" wd_msg;
+  log ~title:"run" "%s (old wd: %S)" wd_msg old_wd;
   run args
--- a/src/frontend/ocamlmerlin/ocamlmerlin.c
+++ b/src/frontend/ocamlmerlin/ocamlmerlin.c
@@ -279,7 +279,7 @@ static int connect_socket(const char *socketname, int fail)
 #ifdef _WIN32
 static void start_server(const char *socketname, const char* eventname, const char *exec_path)
 {
-  char buf[PATHSZ];
+  char buf[PATHSZ], lpSystemDir[PATHSZ];
   PROCESS_INFORMATION pi;
   STARTUPINFO si;
   HANDLE hEvent = CreateEvent(NULL, FALSE, FALSE, eventname);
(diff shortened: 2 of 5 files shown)


src/frontend/ocamlmerlin/ocamlmerlin.c
```diff
@@ -288,6 +288,10 @@ static void start_server(const char *socketname, const char* eventname, const ch
   ZeroMemory(&si, sizeof(si));
   si.cb = sizeof(si);
   ZeroMemory(&pi, sizeof(pi));
+  /* Change directory to TMPDIR, so that process still works if directory
+     is deleted. */
+  if (GetTempPath(PATHSZ, lpTempDir) == 0 || SetCurrentDirectory(lpTempDir) == 0)
```
MisterDA, line 293: I lightly suggest `GetTempPath2`.
> When calling this function from a process running as SYSTEM it will return the path C:\Windows\SystemTemp, which is inaccessible to non-SYSTEM processes. For non-SYSTEM processes, GetTempPath2 will behave the same as [GetTempPath](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha).

MisterDA (approved):

src/frontend/ocamlmerlin/ocamlmerlin.c
nojb: Thanks, done.

voodoos:
@let-def was there a reason why we didn't `chdir /` in the Windows case like we do for unixes ?

nojb:
> @let-def was there a reason why we didn't `chdir /` in the Windows case like we do for unixes ?

What is `/` :) ?

voodoos:
@nojb why not chdir to the root like in the unix cases ?

voodoos:
> What is `/` :) ?

Sorry, I was to lazy to write `chdir("/")` 😅

nojb:
> @nojb why not chdir to the root like in the unix cases ?

Yes, I also considered that by using [`PathCchStripToRoot`](https://learn.microsoft.com/en-us/windows/win32/api/pathcch/nf-pathcch-pathcchstriptoroot), but did not see a clear win over `GetTempPath`. @MisterDA any opinions?

nojb:
> > What is `/` :) ?
> 
> Sorry, I was too lazy to write `chdir("/")` 😅

I meant that `/` does not mean anything on Windows...

MisterDA:
No strong opinion. Switching to the temporary path seems fine.

voodoos:
> I meant that `/` does not mean anything on Windows...

Oh right, each dir has its own root. I would have expected that cd'ing to `\` cd's to the current drive's root, though.

let-def:
> I meant that / does not mean anything on Windows...

Correct me if I am wrong, I remember trying that on windows at some point, and I remember the effect was to move to the root of the current drive (so `C:\` or `D:\`). Though it might have been cygwin doing some magic, I can't tell and it is foggy. (And is it safe to assume that `C:\` is always valid ?)

However I don't know why we were not doing this on Windows before.

nojb:
@voodoos

> > I meant that `/` does not mean anything on Windows...
> 
> Oh right, each drive has its own root. I would have expected that cd'ing to `\` cd's to the current drive's root, though.

@let-def

> > I meant that / does not mean anything on Windows...
> 
> Correct me if I am wrong, I remember trying that on windows at some point, and I remember the effect was to move to the root of the current drive (so `C:\` or `D:\`).

Oops, you are both right; I misspoke.

> Though it might have been cygwin doing some magic, I can't tell and it is foggy. (And is it safe to assume that `C:\` is always valid ?)

If it works, using `chdir("\\")` may indeed be better as it is closer to what is done for Unix. I'll test it and (if it works) will update the PR accordingly.

nojb:
> If it works, using `chdir("\\")` may indeed be better as it is closer to what is done for Unix. I'll test it and (if it works) will update the PR accordingly.

Tried it, seems to work fine, and it is simpler. Updated the PR.

image

@MisterDA do you mind taking a second look, since you already did a first pass over the PR? Thanks!

MisterDA (approved):
I'm not sure I knew you could chdir to root, like this, that's good to know.
This version seems actually better than the last one, `chdir` is a CRT function and will set `errno` and `failwith_perror` uses `perror` and `errno`. The previous version was using Win32 functions, I don't think they set `errno` since the error code has to be retrieved with `GetLastError` (in that sense `failwith_perror`, a wrapper over CRT functions, is wrong for all Win32 functions…).
Reported the problem at #1570.

nojb:
Thanks for the review @MisterDA! I updated CHANGES. This is ready to go as far as I am concerned.

voodoos:
The windows CI failure is a bit surprising.
There is a warning:
```
(cd _build/default/src/frontend/ocamlmerlin && D:\cygwin\bin\x86_64-w64-mingw32-gcc.exe -O2 -fno-strict-aliasing -fwrapv -mms-bitfields -oocamlmerlin.exe ocamlmerlin.c)
ocamlmerlin.c: In function 'start_server':
ocamlmerlin.c:293:7: warning: implicit declaration of function '_chdir'; did you mean 'chdir'? [-Wimplicit-function-declaration]
  293 |   if (_chdir("\\") == -1)
      |       ^~~~~~
      |       chdir
      ```
      
But also a number of different results in the test suite. I will have a look later.

**Comment by nojb** [comment]
> There is a warning:
> 
> ```
> (cd _build/default/src/frontend/ocamlmerlin && D:\cygwin\bin\x86_64-w64-mingw32-gcc.exe -O2 -fno-strict-aliasing -fwrapv -mms-bitfields -oocamlmerlin.exe ocamlmerlin.c)
> ocamlmerlin.c: In function 'start_server':
> ocamlmerlin.c:293:7: warning: implicit declaration of function '_chdir'; did you mean 'chdir'? [-Wimplicit-function-declaration]
>   293 |   if (_chdir("\\") == -1)
>       |       ^~~~~~
>       |       chdir
>       ```

It may have been a missing include. I pushed a possible fix.

> But also a number of different results in the test suite. I will have a look later.

Thanks!

**Comment by voodoos** [comment]
So, one of the failing test is calling:
```
   $ ocamlmerlin server errors -filename test.ml -cmi-path sub < test.ml
 ```
 
 And it looks like with that PR the argument `-cmi-path sub` is ignored (it prints an error saying that some module is not in path).

Not sure what is causing that, there nothing obvious in the diff.

voodoos:
Right, maybe there is something wrong with how Merlin treat the relative path  given as an argument now that the server is not spawned at the same place. @let-def does that ring a 🔔 ?

nojb:
> Right, maybe there is something wrong with how Merlin treat the relative path given as an argument now that the server is not spawned at the same place. @let-def does that ring a 🔔 ?

I think I now what is going on. In the Unix code, we only change the working directory for the child process, but in the present verseion of this PR we are changing the working directory also for the parent process.

I am working on a fix.

dra27:
FWIW (sorry for late entry), `chdir("\\")` isn't strictly as on Unix... it's not impossible to be on a network or removable drive, and then Merlin is part of the "Windows cannot eject this drive right now" problem (i.e. this could give you `G:\` which could be unmounted, whereas `/` cannot for the most part be unmounted) 🙂 Given that there's no need to write to it, I'd go with `GetSystemDirectory` which is guaranteed to exist on a non-removeable drive, or the `GetTempPath2` but we're in hair-splitting territory!

dra27:
Oo, except `GetTempPath2` is _really_ recent (if that hasn't already been spotted), so `GetSystemDirectory` is probably the better one.

nojb:
Thanks @dra27 for the suggestion!

Switched to `GetSystemDirectory`; also, I made it so that only the current directory of the child process is modified (and not the parent's). This should fix the test suite errors. I also had to do a tiny fix to the code in `New_merlin` to restore the "old" working directory after handling a request.

Review welcome! cc @voodoos @dra27 @MisterDA

nojb:
(By the way, this PR fixes #1474.)

nojb:
Friendly ping.

src/frontend/ocamlmerlin/new/new_merlin.ml
```diff
@@ -147,17 +147,23 @@ let run = function
...
   | None -> () end;
-  let wd_msg = match wd with
-    | None -> "No working directory specified"
+  let wd_msg, run = match wd with
+    | None -> "No working directory specified", run
     | Some wd ->
-      try Sys.chdir wd; Printf.sprintf "changed directory to %S" wd
-      with _ -> Printf.sprintf "cannot change working directory to %S" wd
+      let oldwd = Sys.getcwd () in
+      try Sys.chdir wd; Printf.sprintf "changed directory to %S" wd, restore_cwd oldwd run
```
voodoos, line 165: So, do I understand well that without these change we change the server cwd before running a command without restoring it after that. Which means that the first cd to `/` is useless ?

Could you add a test illustrating this change's effect ?

nojb: > So, do I understand well that without these changes we update the server cwd before running a command without restoring it after that.

Yes.

> Which means that the first cd to `/` is useless ?

No, it is still necessary: otherwise after handling a request the server is left on the directory where it was first launched, and this can cause problems.

> Could you add a test illustrating this change's effect ?

I'm not sure how I would go about it, since we are talking about correctly (re)setting the "current directory" of the server process, but there should be no observable difference for the user. But if you have any suggestions, I'm happy to give it a try.

nojb: I pushed a test that shows that the current working directory was not being restored correctly before. I didn't try to exhibit a server misbehaviour because this was harder, it depends on the operating system, etc.

voodoos: Great, thank you !

src/frontend/ocamlmerlin/new/new_merlin.ml
```diff
@@ -147,10 +147,15 @@ let run = function
...
-  | exception e -> Sys.chdir old; raise e
-  | res -> Sys.chdir old; res
+let with_wd ~wd ~old_wd f args =
+  try
+    Sys.chdir wd;
+    log ~title:"run" "changed directory to %S (old wd: %S)" wd old_wd;
+    Fun.protect ~finally:(fun () -> Sys.chdir old_wd) (fun () -> f args)
+  with _ ->
+    log ~title:"run" "cannot change working directory to %S (old wd: %S)"
+      wd old_wd;
```
nojb, line 157: To avoid catching exceptions of `f` by mistake, I suggest to use `match with exception` instead:
```suggestion
let with_wd ~wd ~old_wd f args =
  match Sys.chdir wd with
  | () ->
    log ~title:"run" "changed directory to %S (old wd: %S)" wd old_wd;
    Fun.protect ~finally:(fun () -> Sys.chdir old_wd) (fun () -> f args)
  | exception Sys_error _ ->
    log ~title:"run" "cannot change working directory to %S (old wd: %S)"
      wd old_wd;
```

voodoos: Right, I will put your match back and then rebase and force puh to yoour branch if you don't mind

PR #1581, opened 2023-03-17
Allow monadic IO in dot protocol

3Rafal:
As discussed in: https://github.com/ocaml/ocaml-lsp/issues/1052

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@ merlin 4.9
 ==========
 unreleased
 
+  + merlin binary
+    - Allow monadic IO in dot protocol (#1581)
   + test suite
     - Add missing dependency to a test using ppxlib (#1583)
 
--- a/src/dot-protocol/merlin_dot_protocol.ml
+++ b/src/dot-protocol/merlin_dot_protocol.ml
@@ -63,7 +63,7 @@ end
 type directive = Directive.Processed.t
 
 module Sexp = struct
-  type t = Atom of string | List of t list
+  type t = Csexp.t = Atom of string | List of t list
 
   let atoms_of_strings = List.map ~f:(fun s -> Atom s)
 
@@ -127,8 +127,6 @@ module Sexp = struct
     List (List.map ~f directives)
 end
 
-module Csexp = Csexp.Make (Sexp)
-
 module Commands = struct
   type t = File of string | Halt | Unknown
 
@@ -163,3 +161,40 @@ let read ~in_channel =
 
 let write ~out_channel (directives : directive list) =
   directives |> Sexp.from_directives |> Csexp.to_channel out_channel
+
+module Make (IO : sig
+  type 'a t
+
+  module O : sig
+    val ( let+ ) : 'a t -> ('a -> 'b) -> 'b t
+  end
+end) (Chan : sig
+  type t
+
+  val read : t -> Csexp.t option IO.t
+
+  val write : t -> Csexp.t -> unit IO.t
+end) = struct
+  let read chan =
+    let open IO.O in
+    let+ res = Chan.read chan in
+    match res with
+    | None ->
+        Error (Unexpected_output "Eof")
+    | Some (Sexp.List directives) ->
+        Ok (List.map directives ~f:Sexp.to_directive)
+    | Some sexp ->
+      let msg = Printf.sprintf
+        "A list of directives was expected, instead got: \"%s\""
+        (Sexp.to_string sexp)
+      in
+      Error (Unexpected_output msg)
+
+  module Commands = struct
+    let send_file chan path =
+      Chan.write chan Sexp.(List [Atom "File"; Atom path])
+
+    let halt chan =
+      Chan.write chan (Sexp.Atom "Halt")
+  end
+end
--- a/src/dot-protocol/merlin_dot_protocol.ml
+++ b/src/dot-protocol/merlin_dot_protocol.ml
@@ -127,40 +127,31 @@ module Sexp = struct
     List (List.map ~f directives)
 end
 
-module Commands = struct
-  type t = File of string | Halt | Unknown
+module type S = sig
+  type 'a io
+  type in_chan
+  type out_chan
 
-  let read_input in_channel =
-    let open Sexp in
-    match Csexp.input in_channel with
-    | Ok (List [Atom "File"; Atom path]) -> File path
-    | Ok (Atom "Halt") -> Halt
-    | Ok _ -> Unknown
-    | Error _msg -> Halt
-
-  let send_file ~out_channel path =
-    Sexp.(List [Atom "File"; Atom path])
-    |> Csexp.to_channel out_channel
-end
-
-type read_error =
+  type read_error =
   | Unexpected_output of string
   | Csexp_parse_error of string
 
-let read ~in_channel =
-  match Csexp.input in_channel with
-  | Ok (Sexp.List directives) ->
-      Ok (List.map directives ~f:Sexp.to_directive)
-  | Ok sexp ->
-    let msg = Printf.sprintf
-      "A list of directives was expected, instead got: \"%s\""
-      (Sexp.to_string sexp)
-    in
-    Error (Unexpected_output msg)
-  | Error msg -> Error (Csexp_parse_error msg)
+  (** [read] reads one csexp from the channel and returns the list of
+      directives it represents *)
+  val read :
+    in_chan -> (directive list, read_error) Merlin_utils.Std.Result.t io
 
-let write ~out_channel (directives : directive list) =
-  directives |> Sexp.from_directives |> Csexp.to_channel out_channel
+  val write : out_chan -> directive list -> unit io
+
+  module Commands : sig
+    type t = File of string | Halt | Unknown
+    val read_input : in_chan -> t io
+
+    val send_file : out_chan -> string -> unit io
+
+    val halt : out_chan -> unit io
+  end
+end
 
 module Make (IO : sig
   type 'a t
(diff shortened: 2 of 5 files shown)


voodoos:
I think we should use the new interface in Merlin itself and remove the old io functions.
We should also add comments explaining how to use the new interface and maybe point to ocaml-lsp's usage.

src/dot-protocol/merlin_dot_protocol.mli
```diff
@@ -74,19 +74,42 @@ end
...
+    val read_input : Chan.in_chan -> t Chan.IO.t
+
+    val send_file : Chan.out_chan -> string -> unit Chan.IO.t
+
+    val halt : Chan.out_chan -> unit Chan.IO.t
+  end
+end
 
-val write : out_channel:out_channel -> directive list -> unit
+module StdChan : Chan with type in_chan = in_channel and type out_chan = out_channel and type 'a IO.t = 'a
```
voodoos, line 115: I think we should instantiate the functor here so that we can use it directly in Merlin.

Anyway, configuration loading is tightly tied to `merlin-lib` right now, we should make it more modular at some point. For exemple, `ocaml-lsp` will still need to re-implement `mconfig_dot.ml` after this patch: https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/merlin_config.ml 

(but at least with this patch they can stop vendoring)

3Rafal: Sure, I have created `Merlin_dot_protocol.Std` module.

src/dot-protocol/merlin_dot_protocol.mli
```diff
@@ -74,19 +74,42 @@ end
...
+
+  type in_chan
+  type out_chan
+
+  val read : in_chan -> (Csexp.t, string) result IO.t
+
+  val write : out_chan -> Csexp.t -> unit IO.t
+end
+
+module Make (Chan : Chan) : sig
```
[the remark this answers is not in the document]
3Rafal: Module `Chan` depends on `IO` so I thought they can be packed together.
The previous ocaml-lsp use case:
```ocaml
module Dot_protocol_io =
  Merlin_dot_protocol.Make
    (Fiber)
    (struct
      include Lev_fiber_csexp.Session

      let write t x = write t [ x ]
    end)
```

Now, after fitting Std io/chan there are small differences in types:
- Fiber uses the same type for in and out channel, Std uses in_channel and out_channel
- Lev_fiber_csexp.Session uses Option and Std uses Result for error handling

So my line of reasoning was:
1. Fiber and Lev_fiber_csexp doesn't exactly fit the interface
2. ocaml-lsp will already have to craft a modules instead using them directly
3. So we may merge two modules for conciseness.

However I see your point and I've changed it to fit your description.

src/dot-protocol/merlin_dot_protocol.ml
```diff
@@ -127,39 +127,74 @@ module Sexp = struct
...
+module StdChan = struct
+  type in_chan = in_channel
+  type out_chan = out_channel
+  let read = Csexp.input
+  let write = Csexp.to_channel
+end
 
-let write ~out_channel (directives : directive list) =
-  directives |> Sexp.from_directives |> Csexp.to_channel out_channel
+module Std = Make(StdIO)(StdChan)
```
voodoos, line 200: I think the most important feature of this instance is that it is blocking. A name like `Blocking` or `Blocking_std` might be better.

Also  let's inline the arguments, they are small enough for that.

src/dot-protocol/merlin_dot_protocol.ml
```diff
@@ -127,39 +127,74 @@ module Sexp = struct
...
 
-module Commands = struct
-  type t = File of string | Halt | Unknown
+module Make (IO : sig
+    type 'a t
+
+    module O : sig
+      val ( let+ ) : 'a t -> ('a -> 'b) -> 'b t
+    end
+  end) (Chan : sig
```
voodoos, line 140: Indentation of the arguments feels off. I would endline before each.
(hopefully we will soon ocamlformat the codebase and don't do this kind of comments anymore)

src/dot-protocol/merlin_dot_protocol.ml
```diff
@@ -127,39 +127,74 @@ module Sexp = struct
     List (List.map ~f directives)
 end
 
-module Csexp = Csexp.Make (Sexp)
+type read_error =
+  | Unexpected_output of string
+  | Csexp_parse_error of string
```
voodoos, line 132: This could probably be scoped to the functor's body.

src/dot-protocol/merlin_dot_protocol.mli
```diff
@@ -74,19 +74,53 @@ end
...
-(** [read inc] reads one csexp from the channel [inc] and returns the list of
-  directives it represents *)
-val read : in_channel:in_channel -> (directive list, read_error) Merlin_utils.Std.Result.t
+module Make (IO : sig
+  type 'a t
+
+  module O : sig
+    val ( let+ ) : 'a t -> ('a -> 'b) -> 'b t
+  end
+end) (Chan : sig
```
voodoos, line 87: Same indentation issue.

src/dot-protocol/merlin_dot_protocol.mli
```diff
@@ -74,19 +74,53 @@ end
...
+    val send_file : Chan.out_chan -> string -> unit IO.t
 
-val write : out_channel:out_channel -> directive list -> unit
+    val halt : Chan.out_chan -> unit IO.t
+  end
+end
+
+module Std : sig
+  module Commands :
+  sig
```
voodoos, line 113: We could name that sig to avoid duplicating it (module type S = sig end)

src/kernel/mconfig_dot.ml
```diff
@@ -279,12 +279,13 @@ let get_config { workdir; process_dir; configurator } path_abs =
       workdir
   in
   let query path (p : Configurator.Process.t) =
+    let open Merlin_dot_protocol.Std in
     log_query path;
-    Merlin_dot_protocol.Commands.send_file
-      ~out_channel:p.stdin
+    Commands.send_file
+      p.stdin
       path;
```
voodoos, line 286: This could be one-lined

voodoos (commented):
Mostly cosmetic changes...

src/dot-protocol/merlin_dot_protocol.mli
```diff
@@ -74,19 +74,53 @@ end
...
-end
-
 type read_error =
   | Unexpected_output of string
   | Csexp_parse_error of string
 
-(** [read inc] reads one csexp from the channel [inc] and returns the list of
-  directives it represents *)
-val read : in_channel:in_channel -> (directive list, read_error) Merlin_utils.Std.Result.t
+module Make (IO : sig
```
voodoos, line 81: I think we should add a docstring here to explain that this is provided for projects using merlin as a library

voodoos:
Also we will need a changelog entry :-)

src/dot-protocol/merlin_dot_protocol.ml
3Rafal: Done, used `Blocking`

src/dot-protocol/merlin_dot_protocol.ml
3Rafal: Done, I've used ocamlformat with config from #1586 . I hope it's fine now :)

src/dot-protocol/merlin_dot_protocol.ml
```diff
@@ -127,39 +127,74 @@ module Sexp = struct
...
+      val ( let+ ) : 'a t -> ('a -> 'b) -> 'b t
+    end
+  end) (Chan : sig
+          type in_chan
+          type out_chan
+
+          val read : in_chan -> (Csexp.t, string) result IO.t
+
+          val write : out_chan -> Csexp.t -> unit IO.t
+        end) = struct
```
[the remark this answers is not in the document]
3Rafal: done

src/dot-protocol/merlin_dot_protocol.mli
3Rafal: Sure, I've added a docstring. Do you think it's clear enough?

CHANGES.md
```diff
@@ -2,6 +2,9 @@ merlin 4.9
 ==========
 unreleased
 
+  + merlin binary
+    - Allow monadic IO in dot protocol (#1581)
+
```
voodoos, line 7: could you remove that space ?

voodoos:
Ok, let's stop worrying about the indentation and make this PR the first one to be ocamlformatted.

Can you rebase and clean the history ? Then I will take care of the formatting.

src/dot-protocol/merlin_dot_protocol.mli
voodoos: Forgot about this one: the signature of the `Blocking` module is, as expected, the same as the signature of the body of the functor Make.

You can give a name to that signature and use it both for Make and Blocking

voodoos:
Thanks for the changes @3Rafal !

@rgrinberg , could you confirm that the final interface suits your needs ?

src/dot-protocol/merlin_dot_protocol.mli
```diff
@@ -74,19 +74,53 @@ end
...
-  directives it represents *)
-val read : in_channel:in_channel -> (directive list, read_error) Merlin_utils.Std.Result.t
+  (** [read] reads one csexp from the channel and returns the list of
+      directives it represents *)
+  val read :
+    in_chan -> (directive list, read_error) Merlin_utils.Std.Result.t io
+
+  val write : out_chan -> directive list -> unit io
+
+  module Commands : sig
```
rgrinberg, line 93: If a type doesn't depend on any of the functor arguments, it would be cleaner to move it out and make it usable in a wider context. Same for `read_error`.

I think this is how I did in my original patch.

src/dot-protocol/merlin_dot_protocol.ml
```diff
@@ -127,39 +127,94 @@ module Sexp = struct
...
+  module O : sig
+    val ( let+ ) : 'a t -> ('a -> 'b) -> 'b t
+  end
+end) (Chan : sig
+  type in_chan
+  type out_chan
+
+  val read : in_chan -> (Csexp.t, string) result IO.t
+
+  val write : out_chan -> Csexp.t -> unit IO.t
```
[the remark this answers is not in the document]
3Rafal: I'm not sure. I think the fork uses a single csexp as well. 
https://github.com/ocaml/merlin/blob/209e5f24f9a3a98d4319717c7b1738eb78d2e175/src/dot-protocol/merlin_dot_protocol.mli#L105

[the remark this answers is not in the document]
rgrinberg: @ddickstein you might be using an old version somewhere.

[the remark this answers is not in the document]
ddickstein: I think you might be misreading. The functor takes a module whose `write` is implemented in terms of a single csexp. The module that is provided, `Lev_fiber_csexp.Session`, implements `write` in terms of a list, so there's a redefinition of `write` that implements it for the functor by packaging the csexp into a singleton list and calling `Lev_fiber_csexp.Session.write`. The incompatibility I was pointing out as an issue was on the read side, where the latest version of Merlin now expects the implementation that returns a result, but `Lev_fiber_csexp.Session.read` returns an option, and there's no re-implementation to satisfy the functor because ocaml-lsp hasn't yet updated to the latest version of Merlin. But I am now on latest Merlin and on latest ocaml-lsp, so I'm seeing an incompatibility.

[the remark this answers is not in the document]
voodoos: Yes, not sure why the `result` turned into an option, we can easily change that.

However even with that change it is important to note that the current Merlin master should not be expected to be compatible with ocaml-lsp. We still have an upstreaming PR to merge (#1585) and then we will have to make changes to ocaml-lsp to actually use the new api. Only then both projects will we be compatible for the first time.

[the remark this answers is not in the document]
3Rafal: Implementation used internally by merlin ("Blocking"), uses Csexp library. `read` is implemented as:
```ocaml
let read = Csexp.input
```
which has type of
```
val input : in_channel -> (Sexp.t, string) result
```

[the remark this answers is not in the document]
rgrinberg: Whoops, I should have taken a closer look.

> I decided to use result, and then adjust implementation in ocaml-lsp

How does one know that we've read until EOF in this current implementation?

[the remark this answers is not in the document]
rgrinberg: Whoops, sorry for not reading this closer.

> I decided to use result, and then adjust implementation in ocaml-lsp

What would be the purpose of the string in Error case?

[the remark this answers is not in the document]
3Rafal: Currently, the Error string is not used. I reused the old implementation, which worked that way. Do you think it would be a better idea to change the interface to use option?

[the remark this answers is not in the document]
rgrinberg: Nah, the result type is fine. I'm just curious what you plan to use it for. If it's an error message, It's not really possible to display it in a friendly way to an lsp user through the merlin API.

PR #1584, opened 2023-03-24
Add a cache mechanism for the PPX phase

pitag-ha:
There are three different cache mechanisms for the typer phase, but none for the parsetree-related phases. This commit adds a simple cache for the reader phase and for the PPX phase, the important one being the latter.

Some things that might call your attention are:
- It's a very simple cache. As opposed to file caches for example, it doesn't have any clearance or cache invalidation checks: for every single query, the old entry is simply overridden with the new parsetree.
- I'm using a reference to hold the old parsetree in memory. Let me know if you'd like me to use a different data type!
- The cache for the PPX phase depends on the reader phase: the expanded parsetree can only be taken from the cache, if the source hasn't changed. I pass that information through the mpipeline from the mpipeline reading step to the mpipeline PPX step. That has made the source component of the mpipeline a bit ugly: I've turned it from a tuple to a triple. Let me know if you want me to pass the information in a different way to the PPX step or if you want me to make a record out of that triple!
- I've extended the `File_id` module with a `get_res` function. The reason for that is that when in doubt whether the cache can be used or not, it's better to not use it IMO: correctness over performance IMO. The case of an error while computing `File_id` is definitely a reason to doubt whether the cache can be used. However, that has made this cache incoferent with the file caches, which go ahead with a generic file id when a file id couldn't be computed. Let me know if you want more coherence here!


rgrinberg:
Is there a way to manually coontrol this cache or toggle it somehow? I'm concerned that it isn't correct for ppx that relies on `preprocessor_deps`.

pitag-ha:
That's a very good point, @rgrinberg. Thanks!

Do you mean toggling the cache from the outside, i.e. by the instance that's calling Merlin? How would that instance know whether there are preprocessor dependencies?

Maybe instead of toggling it from the outside, another option could be the following (let me know if that's already what you had in mind). We could extend the `dot-merlin-reader` to capture information about `preprocessor_deps`. I guess, for that we'd also need to extend the `dune` command (or the equivalent in a different set-up than `dune`), that feeds that information to the `dot-merlin-reader`. Is that right?

So for new versions of `dune`/`jenga`, Merlin would have information about preprocessor dependencies in its configuration. Then, the PPX mechanism could be as follows: does the configuration have information about the preprocessor dependencies? If no, no PPX cache. If yes, cache which takes preprocessor dependencies into account.

What do you think? Is that what you had in mind?

rgrinberg:
> Do you mean toggling the cache from the outside, i.e. by the instance that's calling Merlin? How would that instance know whether there are preprocessor dependencies?

I meant just being able to turn it off or control it when using merlin as a library. In lsp, I would use dune rpc  to tell merlin when a preprocessed AST would be invalidated.

> . We could extend the dot-merlin-reader to capture information about preprocessor_deps. I guess, for that we'd also need to extend the dune command (or the equivalent in a different set-up than dune), that feeds that information to the dot-merlin-reader. Is that right?

That would be very difficult because it would mean re-implementing dune's DSL specification for deps inside merlin.

pitag-ha:
> In lsp, I would use dune rpc to tell merlin when a preprocessed AST would be invalidated.

That's interesting! Thanks for explaining. We also need a way to make this work correctly in a different context than lsp, though. I'm not very familiar with the different communication layers that are done in different contexts for Merlin to get all its config information. I'll ask @voodoos about that and come back to you.

pitag-ha:
> In lsp, I would use dune rpc to tell merlin when a preprocessed AST would be invalidated.

That's interesting! Thanks for explaining. We also need a way to make this work correctly in a different context than lsp, though. I'm not very familiar with the different communication layers that are done in different contexts for Merlin to get all its config information. I'll ask @voodoos about that and come back to you.

rgrinberg:
This PR reminded me of alternative method to speed up ppx that works without relying on cache hits (we won't get the cache hits in many cases). One of the reasons why ppx (and any other preprocessor) is slow is the cost of launching an external process every single time the input changes. I bet we could speed up preprocessing materially if we kept the ppx executable persistent, and just dispatched pp requests to it. It would require some changes to ppxlib to allow drivers to run in a `--server` mode, but it would be fairly universal.

Even in this server mode however, we'll need some co-operation from dune to re-run the server whenever the binary changes. On the lsp side, I have no problems with managing such ppx subprocesses. But perhaps it wouldn't work well with merlin?

In any case, I think it does make sense to make merlin a little more flexible and allow callers to customize how they would like to run preprocessors.

voodoos:
A server mode for the driver would be a nice feature. Additionally the ppxes currently write the ast to a file on disk that is then read by merlin. Skipping that intermediary file might also speed up things a little ?

pitag-ha:
@ddickstein, yes, I also think that's a good idea, for sure "as a start" and possibly permanently. In fact, that's what I understood @rgrinberg was referring to.

About the "speeding up the PPX phase by improving the interaction between Merlin and the PPX driver" idea: there are three different aspects to that point. As you, @rgrinberg, have pointed out, spawning a new process each time is time-consuming. As you, @voodoos, have pointed out, passing the AST between Merlin and the PPX process via disc is time-consuming. And the third one is that Merlin marshalling and the PPX process unmarshalling (and later vice-versa) is time-consuming. For the first, we can have the PPX driver run in server mode. For the second, we could pass the AST via a serial stream or similar. For all three of them, including the third, we could not do those two things and instead embed a PPX driver into Merlin that dynamically links the PPXs (that driver could be similar to the one `utop` uses). The latter also has timing downsides, though, so I'd need to find out what the time pay-off would be.

I've been analyzing some Merlin latency data during the last weeks, though, and I think there are several far more impactful things we can do currently to improve the PPX phase latency. One is that several AST passes could/should be gotten rid of, particularly for any Merlin query different than `errors`. And the other one is a cache for the PPX phase.

To give an idea about why I think those things are more impactful than improving the interaction between Merlin and the PPX driver: I had a look at the system time Merlin and its child process spend during the PPX phase, and it's marginal: I haven't found a case in which it was over 20 msecs, yet. In the cases in which the PPX phase is a bottleneck, it spends up to 500 msecs. And that without a cache, so it does so every single time for every single query.

rgrinberg:
>  For all three of them, including the third, we could not do those two things and instead embed a PPX driver into Merlin that dynamically links the PPXs (that driver could be similar to the one utop uses)

Is that really what utop does? I'm pretty sure it just uses compiler-libs which executes the external processes.
In any case, dynlink is very complex and brittle. I personally wouldn't consider it.

>  Is the complete-prefix situation the situation or one of the situations you had in mind?

You have the right idea. When the user is actively writing code, we can't rely on cache hits. As you've mentioned, providing completion will be triggered heavily in this phase, so we'll be doing a lot of preprocessing without a lot of cache hits.

rgrinberg:
>  For the second, we could pass the AST via a serial stream or similar.

Wouldn't the server mode already cover that? We'd be passing the AST over the socket. This should be faster than using a temporary file.

rgrinberg:
> does the VSCode plug-in or the lsp-server in general have a way to cancel queries?

The LSP protocol supports the client cancelling a request whenever the client is no longer interested in the response. It is relevant to ppx as well. For example, if the client requests diagnostics and immediately edits the document, the diagnostics request would be cancelled and we'd do well to stop whatever preprocessing is going on. Unfortunately, that means sending a signal to stop the ppx driver process, usually followed by re-running it on subsequent requests. That's all pretty inefficient, and we could do better without all these processes.

pitag-ha:
> The LSP protocol supports the client cancelling a request whenever the client is no longer interested in the response.

Cool! Does VSCode ever cancel old `complete-prefix` or other queries? I mean the situation where a new query comes in before the last query has replied.

rgrinberg:
I'm not sure unfortunately. I would hope so.

I would also not recommend relying on this cancellation mechanism for performance. It can definitely help us, but unfortunately not all editors (their corresponding lsp clients) have it and their implementations are not really consistent.

TyOverby:
I'm in favor of @ddickstein's suggestion to ignore the cache if `preprocessor_deps` is used due to how infrequently it appears.

This "ppx server" idea seems great too - for files that are being actively edited, it'll be necessary to reduce the cost of running necessary ppxes - but it's not going to be easy; some unknown number of ppxes use global state under the assumption that it's going to be invoked on a single ml/mli file at a time.

pitag-ha:
> I'm in favor of @ddickstein's suggestion to ignore the cache if preprocessor_deps is used due to how infrequently it appears.

:100: 
I'll do that as soon as I'm back from Easter vacation.

> some unknown number of ppxes use global state under the assumption that it's going to be invoked on a single ml/mli file at a time.

Hmm, this might be important for me to clarify, since I might be missing something. Could you explain this a bit more? I'm not sure what you mean by the "assumption that it's going to be invoked on a single ml/mli file at a time.". Isn't the PPX driver always just invoked on the AST (ml/mli file) of the current buffer?

Btw, I also think that if we make the PPX driver run in server mode, we can later improve the performance of the PPX phase significantly: we can make the PPX driver conscious of whether all the PPXs it's linking are "state-less". In the case of a state-less driver, we could invoke the driver on each structure/signature item separately (thanks to the server mode) instead of invoking it on the whole AST. So we could implement a structure/signature item based PPX cache, which keeps track of all expanded items and only re-expands the ones that were modified. That cache would even be more powerful than the typer cache of the current buffer,, since it really only need to re-expand the modified items, whereas the typer cache needs to re-type all items from the first modified item on. I haven't thought this idea though yet, though. It just came to my mind when reading about the server idea. Does it go in the same direction as what you have in mind, @TyOverby?

pitag-ha:
I've made the cache togglable now. @ulysse, could you review when you have a moment?

Btw, there seem to be two different workflows to add new config values to merlin: via a new config directive, such as [exclude-query-dir](https://github.com/ocaml/merlin/blob/master/src/dot-protocol/merlin_dot_protocol.ml#L42); or simply putting it into the generic bucket of [flags](https://github.com/ocaml/merlin/blob/master/src/dot-protocol/merlin_dot_protocol.ml#L38), such as `-ocamllib-path`, `-log-file` etc. The first seemed quite involved, and the latter was extremely simple, so I've done the latter for now. Does that sound good? And also: do you know when it's necessary/advised to add new dedicated directives?

Also, I've decided to dump the new config value "use-ppx-cache" when someone runs `dump-configuration`. Most config values are dumped, but some aren't (such as `config_path` and `exclude_query_dir`).

pitag-ha:
Btw, I forgot to mention: I've tested enabling the cache on a project with a `.merlin`-file by adding the new `FLG -use-ppx-cache` manually, and that works. When/if we merge this, we'll need to add support to dune, jenga and lsp-server to turn on the PPX cache for us whenever no PPX dependencies are used.

voodoos:
> Btw, I forgot to mention: I've tested enabling the cache on a project with a `.merlin`-file by adding the new `FLG -use-ppx-cache` manually, and that works. 

Could you write a test illustrating the cache behavior in the test suite ?
Maybe showing that Merlin gave the same answer if the cache was active but the ppx add a hidden dependency that changed. 

Note about test organization: so far the testsuite has only one test for ppxs named `with-ppx.t`. It is a directory test. I think we should rename that test `typed-holes.t` and make `with-ppx` a normal directory with all the tests depending on ppxlib. So that we would have `test-dirs/with-ppx/typed-holes.t`,  `test-dirs/with-ppx/your-new-test.t` etc.

pitag-ha:
> Maybe showing that Merlin gave the same answer if the cache was active but the ppx add a hidden dependency that changed.

With "hidden dependency", are you referring to PPX dependencies? Whether PPX dependencies influence the PPX cache, doesn't depend on Merlin with this implementation, but on the build system. Merlin only knows whether the cache is disabled or enabled, not why: the PPX cache is disabled by default and can be enabled by the merlin config, i.e. by the build system (the build system should enable it if and only if there are no PPX dependencies).

What Merlin knows about is:
- is the cache disabled (that's the default)?
- was the PPX binary modified?
- were the args passed to the PPX binary modified?
- was the workdir modified?
- was the source code modified?

I can add tests to show that when any of the above points is the case, the parsetree is expanded instead of being pulled from the cache. That would be a different approach of test than what you've suggested (if I've understood you correctly): iiuc, you were suggesting an "add a test showing the limits of the current implementation -> fix the implementation -> see the fix in the test" approach. Is that right? I'm suggesting to just add a "test that the implementation is right" approach. Do you want me to add some test like that?

voodoos (commented):
Did a first rough review.

src/kernel/mconfig.ml
```diff
@@ -355,6 +358,11 @@ let merlin_flags = [
     marg_path (fun path merlin -> {merlin with stdlib = Some path}),
     "<path> Change path of ocaml standard library"
   );
+  (
+    "-use-ppx-cache",
+    Marg.unit (fun merlin -> {merlin with use_ppx_cache = true}),
+    ""
```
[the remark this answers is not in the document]
pitag-ha: Good catch, thanks!

src/kernel/mpipeline.ml
```diff
@@ -82,7 +82,7 @@ type t = {
   state  : Mocaml.typer_state;
   raw_source : Msource.t;
   source : (Msource.t * Mreader.parsetree option) lazy_t;
-  reader : (Mreader.result * Mconfig.t) lazy_t;
+  reader : (Mreader.result * Mconfig.t * bool) lazy_t;
```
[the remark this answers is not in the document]
pitag-ha: Sounds good, I've done that.

PD: that's one of the things I mentioned in the first comment of the PR.

src/kernel/mpipeline.ml
```diff
@@ -111,14 +111,14 @@ let reader t = Lazy.force t.reader
...
-let reader_comments  t = (fst (reader t)).Mreader.comments
-let reader_lexer_keywords  t = (fst (reader t)).Mreader.lexer_keywords
-let reader_lexer_errors  t = (fst (reader t)).Mreader.lexer_errors
-let reader_parser_errors t = (fst (reader t)).Mreader.parser_errors
+let reader_config    t = let (_, config, _) = (reader t) in config
+let reader_parsetree t = let (pt, _, _) = (reader t) in pt.Mreader.parsetree
+let reader_comments  t = let (pt, _, _) = reader t in pt.Mreader.comments
+let reader_lexer_keywords  t = let (pt, _, _) = reader t in pt.Mreader.lexer_keywords
+let reader_lexer_errors  t = let (pt, _, _) = reader t in pt.Mreader.lexer_errors
+let reader_parser_errors t = let (pt, _, _) = reader t in pt.Mreader.parser_errors
```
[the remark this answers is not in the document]
pitag-ha: Indeed. It's a record now (see above)

src/kernel/parsetree_cache.ml
```diff
@@ -0,0 +1,82 @@
...
+    let equal { binary_id = b1; args = a1; workdir = w1 }
+        { binary_id = b2; args = a2; workdir = w2 } =
+      File_id.check b1 b2 && List.equal String.equal a1 a2 && String.equal w1 w2
+  end
+
+  type cache = { fingerprints : Fingerprint.t list; result : Mreader.parsetree }
+
+  let cache = ref None
+
+  let with_cache ~source_was_cached (config : Mconfig.t) f =
```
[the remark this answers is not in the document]
pitag-ha: No, I haven't thaught about it very much tbh. But I do think it's a good idea. It would make it more coherent. I don't have time for it today anymore. If you think it's important, I'll try to find time. I don't think it would be that difficult.

src/utils/misc.ml
```diff
@@ -134,6 +134,10 @@ let rec split_path path acc =
     dir :: acc
   | dir -> split_path dir (Filename.basename path :: acc)
 
+let add_segment ~dir ~seg =
+  let dir = if String.ends_with ~suffix:"/" dir then dir else dir ^ "/" in
+  dir ^ "/" ^ seg
+
```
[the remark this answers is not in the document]
pitag-ha: Thanks!

pitag-ha:
I've added the tests now, both the ones I had in mind and the one you had in mind, @voodoos (I think. it's the one at the end of the new test file). The test structure has ended up different than you've suggested. I don't have an opinion on the structure tbh. So if you'd like it the way you've suggested, let me know! I think it should be very simple a sec to change it.

And to come back to whether to add the new use-ppx-cache config via the FLG directive or a new directive. I've done the first, but that might be a problem:

When using .merlin-files, then Merlin `errors` complains when there are unknown flags in the .merlin-file. However, it doesn't complain when there are unknown directives. So adding the new config via a flag will lead to incompatibility problems between new config workflows and old Merlins.

Does anyone know whether that's also a problem when using dune? I'd like to think that the config compatibility story is better for the Merlin-Dune communication, but I don't have any idea. @voodoos, do you know? Also, @rudi, I'm assuming that for ocaml-lsp this isn't a problem. If it is, please let me know!

voodoos:
> And to come back to whether to add the new use-ppx-cache config via the FLG directive or a new directive. I've done the first, but that might be a problem:
> 
> When using .merlin-files, then Merlin `errors` complains when there are unknown flags in the .merlin-file. However, it doesn't complain when there are unknown directives. So adding the new config via a flag will lead to incompatibility problems between new config workflows and old Merlins.

Good catch !

I actually thought that when using `.merlin` files unknown directives would be tagged as error by Merlin and that only when Dune is used those would be ignored.

I see three possible ways to go:
1. Make future versions of Dune with the new flag conflict with older versions of Merlin 
2. Have Merlin ignore unknown flags when they come from Dune (and make future versions of Dune with the new flag conflict with versions of Merlin prior to the one that ignore unknown flags)
3. Use a new config directive

I am personally be in favor of 3 since we control which flags Dune sends to Merlin and if a wrong compiler flag is given it will be already told to the user by the compiler itself when building the project. This way future flag update won't always require a new conflict.

Also, what should happen if multiple ppxes are defined ? I guess we want to keep thing simple and have the flag apply to all of them ?

src/kernel/parsetree_cache.ml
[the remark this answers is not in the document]
pitag-ha: I've functorized the two caches now. It's a bit longer than before, but indeed more coherent and easier to read IMO. Thanks for the idea!

Btw, with this new abstraction, this all-or-nothing cache mechanism can be used for any phase of the pipeline. However, the typer phase already has a way more elaborate cache and the computation phase in the majority of cases isn't a bottleneck. Still, it might be interesting to look into using it for concrete cases of the computation phase at some point, such as `complete-prefix` queries (I haven't thought this through, though; it's just an idea).

src/kernel/phase_cache.mli
```diff
@@ -0,0 +1,42 @@
...
+  (** Phase computation *)
+
+  val title : string
+  (** Phase title for logging *)
+
+  module Fingerprint : sig
+    type input
+
+    type t
+    (** Fingerprint that determines whether the phase cache is invalidated *)
```
voodoos, line 20:
```suggestion
    (** Fingerprint used to determine whether the cache should be invalidated *)
```

src/kernel/phase_cache.mli
```diff
@@ -0,0 +1,42 @@
...
+  type t = { output : Phase.output; cache_was_hit : bool }
+
+  val apply :
+    cache_disabling:string option -> is_pre_invalidated:bool -> Phase.t -> t
+  (** [apply ~cache_disabling ~is_pre_invalidated phase_input] runs the phase
+      computation [Phase.f phase_input], if there's some [cache_disabling].
+      Otherwise, it runs that phase computation with a cache mechanism. Whether
+      the cache is invalidated depends on [is_pre_invalidated] and on the
+      outcome of a [Phase.Fingerprint] comparison between the current
+      fingerprint and the last one. *)
```
voodoos, line 41: Nitpick: I would go something like:

>  Otherwise, the phase computation is ran with a cache mechanism. Whether
> the cache is invalidated depends  on the     outcome of a [Phase.Fingerprint] comparison between the current    fingerprint and the last one. Additionaly the invalidation of the cache can be forced by setting the `force_invalidation`parameter to `true`.*)

src/utils/std.ml
```diff
@@ -346,6 +346,14 @@ module Result = struct
...
   | Ok of 'a
   | Error of 'e
+
+  let map ~f = function
+  | Error err -> Error err
+  | Ok x -> Ok (f x)
+
+  let bind ~f = function
+  | Error err -> Error err
+  | Ok x -> f x
```
voodoos, line 356: Those are already part of the stdlib (minus the labels).
If we want labelled versions we can simply reuse the official implementations:

```
let map ~f r = Result.map f r
```

src/kernel/mconfig.ml
```diff
@@ -617,6 +626,8 @@ let initial = {
 
     exclude_query_dir = false;
 
+    use_ppx_cache        = false;
```
voodoos, line 629: weird alignment

src/kernel/mconfig.ml
```diff
@@ -83,6 +83,8 @@ type merlin = {
   log_sections : string list;
   config_path : string option;
 
+  use_ppx_cache        : bool;
```
voodoos, line 86: Weird `: ` alignment

src/kernel/mpipeline.ml
```diff
@@ -156,18 +236,41 @@ let process
...
+      let lazy ((_, pp_result) as source) = source in
       let config = Mconfig.normalize config in
       Mocaml.setup_reader_config config;
-      let result = Mreader.parse ?for_completion config source in
-      result, config
+      let { Reader_with_cache.output; cache_was_hit } =
+      let cache_disabling =
+        match (config.merlin.use_ppx_cache, pp_result) with
+        | false, _ -> Some "configuration"
+        | true, Some _ -> Some "source preprocessor usage"
```
voodoos, line 246: Couldn't we use the preprocessed parse tree as a fingerprint when the source is preprocessed and have the cache working ?

src/kernel/mpipeline.ml
```diff
@@ -156,18 +236,41 @@ let process
...
     )) in
   let reader = timed_lazy reader_time (lazy (
-      let lazy source = source in
+      let lazy ((_, pp_result) as source) = source in
       let config = Mconfig.normalize config in
       Mocaml.setup_reader_config config;
-      let result = Mreader.parse ?for_completion config source in
-      result, config
+      let { Reader_with_cache.output; cache_was_hit } =
+      let cache_disabling =
```
voodoos, line 243: indentation

src/kernel/mpipeline.ml
pitag-ha: Btw, I've just added a small comment about this in the code.

src/kernel/mpipeline.ml
```diff
@@ -155,19 +235,47 @@ let process
...
+              match (config.merlin.use_ppx_cache, pp_result) with
+              | false, _ -> Some "configuration"
+              | true, Some _ ->
+                (* The cache could be refined in the future to also act on the PP phase.
+                   For now, let's disable the whole cache when there's a PP. *)
+                Some "source preprocessor usage"
+              | true, None -> None
+            in
+            Reader_with_cache.apply ~cache_disabling ~force_invalidation:false
+              { source; for_completion; config }
```
voodoos, line 255: It just occur to me why this feels weird: the input of a `Reader_phase` can contain preprocessed AST and we ignore that in the fingerprinting.

I think a more natural way to disable the cache wen there is actually preprocessing would be to have the fingerprint be `None` (or some better name type constructor) if it is the case of `Some digest` if it is the raw source. Then `Fingerprint.equal` should return false if one of the two compared fingerprint if `None`.

That way we can remove the `cache_disabling` logic and simply pass `config.merlin.use_ppx_cache` or even `config` to the apply function.

Does that sounds reasonable ?

src/kernel/mpipeline.ml
```diff
@@ -155,19 +235,47 @@ let process
...
       Msupport.catch_errors Mconfig.(config.ocaml.warnings) caught @@ fun () ->
-      let parsetree = Mppx.rewrite config parsetree in
-      { Ppx. config; parsetree; errors = !caught }
+      let {Ppx_with_cache.output = parsetree; cache_was_hit = _ } =
+        let cache_disabling =
+          if config.merlin.use_ppx_cache then None
+          else Some "configuration"
+        in
+        let force_invalidation = not source_is_unmodified in
+        Ppx_with_cache.apply ~cache_disabling ~force_invalidation {parsetree; config}
```
voodoos, line 276: Here we also bypass real fingerprinting with the `force_invalidation` parameter.

That means that we get ppx cache misses even for source changes that don't change the parsetree (like adding whitespace).

I don't know if it would be worth it to do it, but we could add a remark about that somewhere.

voodoos: That's room for improvement if the current cache misses too often.

src/kernel/mpipeline.ml
pitag-ha: I used to have the workflow you're proposing in the first commits (before the refactor). I changed it together with the more general refactor to what I have now, but that might have been a bad decision. Let me explain why I changed it. There are two things in which the two workflows differ from each other:

One is where the cache module captures the disabling of the cache. With the proposed/previous workflow, the disabling of the cache is captured in the fingerprint. With the current workflow, it's captured one level higher, i.e. in the cache itself (when the cache is disabled, then [the cache reference is set to `None`](https://github.com/ocaml/merlin/pull/1584/files#diff-7716196a07a66d255ac0cab613acff41943afd7c640a06b820a25d714954af99R32)). Capturing it at the level of the cache means that the cache neither stores a fingerprint nor a "last processing result" when being disabled. I think that makes sense since it's a bit weird to have a cache that stores a "last result" with a `None` fingerprint. What do you think? I'm not super familiar with how caches are usually implemented, so I might easily be missing something.

The other thing that's different is where the _logic of disabling_ the cache lives: with the current workflow, it lives outside the cache module and is passed to the `With_cache.apply` function from there. That has two upsides: 1. we stay more flexible in terms of disabling the cache; 2. it's visual from outside whether the cache is disabled or actually used. With the proposed/previous workflow, it is embedded in the cache logic: the cache is more of a black box from outside that's sometimes disabled and sometimes enabled, without knowing or being able to tweak it from outside which one of the two it is. The upside of that is that we don't need to pass the `cache_disabling` parameter. Btw, if you prefer the latter (which seems to be the case), I could change the workflow I'm talking about here even if I don't change the one from the last paragraph.

src/kernel/mpipeline.ml
pitag-ha: Yes, that's a good point. It brings us again to the question of whether it'd be easy to have a digest for the parsetree structure (as opposed to the source code, i.e. string). Do you know whether source code modifications that preserve the parsetree happen more or less frequently? For now, I've just added a comment about this.

src/kernel/mpipeline.ml
voodoos: Right, let's keep the more flexible version for now.

src/kernel/mpipeline.ml
voodoos: I think the most frequent cases are whitespace changes: adding or removing lines, adding or removing spaces.
There are also a few syntax that have an equivalent representation.

pitag-ha:
Cool! I've just rebased, squashed the commits, and added a changelog entry. After rebasing, I get a small formatting diff in `tests/test-dirs/type-enclosing/inside-tydecl.t`, also on the commit before my commit.

PR #1585, opened 2023-03-27
Configurable Unix system command

3Rafal:
As described in https://github.com/ocaml/ocaml-lsp/issues/1052

There was already a discussion about introducing Spawn dependency to replace `Sys.command` here: #1420


rgrinberg:
Btw, this fix isn't bringing us to an ideal long term situation. We don't want merlin imposing on lsp on how to lunch processes, nor do we want lsp imposing on merlin in the same way. Something closer to what we want would be an API for running preprocessors. E.g:

```
val pp : cwd:string -> prog:string -> args:string list -> (string, int) result
```

voodoos:
> Btw, this fix isn't bringing us to an ideal long term situation. We don't want merlin imposing on lsp on how to lunch processes, nor do we want lsp imposing on merlin in the same way. Something closer to what we want would be an API for running preprocessors. E.g:
> 
> ```
> val pp : cwd:string -> prog:string -> args:string list -> (string, int) result
> ```

We could make it customizable in https://github.com/ocaml/merlin/blob/master/src/utils/lib_config.mli

It's not the most elegant solution but it might provide the flexibility we need until a more involved refactor of "merlin-as-a-lib" is done.

3Rafal:
@rgrinberg , I'm not sure that I follow.

If we create the API with following function:
```
val pp : cwd:string -> prog:string -> args:string list -> (string, int) result
```

then how will we handle cancelation? I assume that returned int stands for exit code and PID is not exposed.

rgrinberg:
The implementation of that function on the lsp side would register the spawned pid somewhere and then kill it once cancellation is called. By the way, the error code should be a bit more accurate:

```
val pp : cwd:string -> prog:string -> args:string list -> (string, [`Finish of int | `Cancelled ]) result
```

3Rafal:
Ok, I'll just rebase it on current main branch and provide a changelog :-)

3Rafal:
@rgrinberg

> Is there a reason why we're in a rush to fix this issue?

The primary reason for prioritizing this issue is the potential for reducing maintenance costs on both projects. By upstreaming the merlin fork, we can simplify the overall development process and minimize duplicated efforts.

Moreover, this issue has been identified as a significant factor blocking adoption for some users in the OCaml community.

The proposed Pull Request offers a quick and efficient way to upstream this patch. This solution is not the final version. We intend to further refine the integration between ocaml-lsp and merlin once the initial version has been upstreamed.

voodoos:
@rgrinberg, about the return type: ```(string, [`Finish of int | `Cancelled ]) result```: 

Adding a ``` `Canceled``` case without properly handling it on Merlin's side wouldn't make much sense to me. 
Right now, if a PPX fails Merlin simply keeps going down the pipeline with the non-preprocessed source.

We should at some point (and probably sooner than later) add cancellation support to the `Mpipeline` itself, but I think this is out-of-scope for this PR that really simply aims to provide enough flexibility to support current `ocaml-lsp` usage without additional patches.

rgrinberg:
Sure, I was talking about the ideal UI that we should be looking towards. Of course merlin should handle cancellation in a smarter way than just continuing with the unpreprocessed ast.

voodoos:
@3Rafal could you add a changelog entry ? I will try to review by the end of the week.

src/ocaml/driver/pparse.ml
```diff
@@ -158,8 +158,12 @@ let apply_pp ~workdir ~filename ~source ~pp =
...
   end;
   let fn_out = fn_in ^ ".out" in
-  let comm = pp_commandline pp fn_in fn_out in
-  let ok = merlin_system_command comm ~cwd:workdir = 0 in
+  let args = [
+    Filename.quote fn_in;
+    Printf.sprintf "1>%s" (Filename.quote fn_out)
+  ] in
+  let comm = commandline pp args in
+  let ok = merlin_system_command ~prog:pp ~args:args ~cwd:workdir = 0 in
```
voodoos, line 166: I think this function should not expect its arguments to be shell-quoted.

Currently, merlin runs a shell command (`Sys.command "cd ... && ..."`) that does require specific quoting.

But users of the library that want to use something such as `Spawn` don't need  the arguments to be quoted since `Spawn` is directly using a C stub calling `execve` instead of relying on the shell.

src/ocaml/driver/pparse.ml
```diff
@@ -41,26 +41,30 @@ let report_error = function
 
 external windows_merlin_system_command : string -> cwd:string -> int = "ml_merlin_system_command"
 
-let merlin_system_command =
+let commandline prog args =
+  Printf.sprintf "%s %s" prog (String.concat ~sep:" " args)
+
+let merlin_system_command ~prog ~args ~cwd =
+  let cmd = commandline prog args in
   if Sys.win32 then
```
voodoos, line 49: We probably want to move that in the (appropriately renamed) `System_command.unix` function so that users of the lib can also redefine the windows spawning.

src/utils/std.ml
```diff
@@ -751,6 +751,13 @@ module Shell = struct
     List.rev !comps
 end
 
+module System_command = struct
+  let unix = ref @@
```
voodoos, line 755: With the other comments in mind, I think a better name would be:

`Std.System.run_in_directory` or something similar.

src/ocaml/driver/pparse.ml
3Rafal: Current merlin fork used by ocaml-lsp works with quoting:
```ocaml
let merlin_system_command =
  if Sys.win32 then
    windows_merlin_system_command
  else
    fun cmd ~cwd ->
      let prog = "/bin/sh" in
      let argv = ["sh"; "-c"; cmd] in
      let stdin = Unix.openfile "/dev/null" [ Unix.O_RDONLY ] 0x777  in
      let pid =
        let cwd : Spawn.Working_dir.t = Path cwd in
        let stdout = Unix.stderr in
        Spawn.spawn ~cwd ~prog ~argv ~stdin ~stdout ()
      in
      let (_, status) = Unix.waitpid [] pid in
      let res =
        match (status : Unix.process_status) with
        | WEXITED n -> n
        | WSIGNALED _ -> -1
        | WSTOPPED _ -> -1
      in
      Unix.close stdin;
      res

let ppx_commandline cmd fn_in fn_out =
  Printf.sprintf "%s %s %s%s"
    cmd (Filename.quote fn_in) (Filename.quote fn_out)
    (if Sys.win32 then "" else " 1>&2")

let apply_rewriter magic ppx (fn_in, failures) =
  let title = "apply_rewriter" in
  let fn_out = Filename.temp_file "camlppx" "" in
  let comm = ppx_commandline ppx.workval fn_in fn_out in
  log ~title "running %s from directory %S" comm ppx.workdir;
  Logger.log_flush ();
  let failure =
    let ok = merlin_system_command comm ~cwd:ppx.workdir = 0 in
[...]
```

The problem is that we can't move filename quoting to `run_in_directory` function, because it takes list of arguments. We're not able to decide which are filenames nad needs to be quoted without changing it's signature.

src/ocaml/driver/pparse.ml
```diff
@@ -41,26 +41,30 @@ let report_error = function
...
 let apply_rewriter magic ppx (fn_in, failures) =
   let title = "apply_rewriter" in
   let fn_out = Filename.temp_file "camlppx" "" in
-  let comm = ppx_commandline ppx.workval fn_in fn_out in
+  let args = [
+    Filename.quote fn_in;
+    Printf.sprintf "%s%s"
+      (Filename.quote fn_out)
+      (if Sys.win32 then "" else " 1>&2")
+  ] in
```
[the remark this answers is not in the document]
3Rafal: done

src/utils/std.ml
3Rafal: You've requested Windows hook as well. Should we name it `run_in_directory_unix` and `run_in_directory_windows`?

src/ocaml/driver/pparse.ml
voodoos: I wonder why is the shell still used with spawn. @rgrinberg ? 
Couldn't `[ppx_cmd; ppx_arg1; ...]` be used directly instead of `["sh"; "-c"; "quoted_ppx_cmd_with_args"]` ?

rgrinberg: I just never got around to changing it yet.

voodoos:
I gave a go at the latest changes we discussed with Rafal. This lead me to (re)-discover the source of our difficulties: in configuration, a ppx and its arguments are passed in one block to Merlin which makes it hard to distinguished the program from it's arguments. Additionally, when the `.merlin` file has been hand-crafted it can rely on various different quoting styles and even shell expansions / embedded commands. (free visit to the zoo: https://github.com/search?q=path%3A%22.merlin%22+FLG+-ppx&type=code&p=1).

That means that unless we decide to parse these and unquote them (which is something we really don't want to do) there is no way for us to use sane quotation functions like `Filename.quote_command`, or drop the call to the shell, if we want to ensure Merlin compatibility with the whole Zoo.

I documented the hook exposed in `Lib_config` with all that should be taken care of by implementers:

https://github.com/ocaml/merlin/blob/5d09af9d377fcbbabeeb54a16c1ce020bf1cd24f/src/utils/lib_config.mli#L22-L54

@rgrinberg We should at some point add a new directive to the configuration protocol to finally pass ppx configuration in a clean way when using Dune.

PR #1586, opened 2023-03-27
Ocaml-format the codebase

voodoos:
I spent some time playing with options.
I think these one are not too bad.
Huge diff still, maybe we could do it progressively ? 
Like having each PR format the files they touched ?

One option that makes me hesitate is `cases-exp-indent`:
- The codebase mostly use 2
- The default is 4, which is also the convention used in most parts of the compiler

I personally prefer 4 but choosing 2 reduce the diff significantly. From 10,000 additions / deletions to 9,000. Well, that's not so significant 😅.


.ocamlformat
```diff
@@ -0,0 +1,16 @@
+version=0.24.1
+break-cases=fit-or-vertical
+break-struct=natural
+doc-comments=before
+cases-exp-indent=2
+doc-comments-tag-only=fit
+dock-collection-brackets=false
+# Preserve beign/end
```
3Rafal, line 8: typo: begin 😉

3Rafal:
Hey @voodoos , I tried your config on #1581 and got weird results:
```ocaml
module Blocking =
  Make (struct
      type 'a t = 'a module O = struct let ( let+ ) x f = f x end
    end)
    (struct
      type in_chan = in_channel
      type out_chan = out_channel
      let read = Csexp.input
      let write = Csexp.to_channel
    end)
```

type definition in the same line as nested module definition looks confusing
 
my full config:
```
$ ocamlformat --print-config
profile=default
range=<whole input>
quiet=false
ocaml-version=4.04.0
max-iters=10
margin-check=false
disable=false
debug=false
comment-check=true
wrap-fun-args=true
wrap-comments=false
type-decl-indent=2
type-decl=compact
stritem-extension-indent=0
space-around-variants=false (file .ocamlformat:15)
space-around-records=false (file .ocamlformat:14)
space-around-lists=false (file .ocamlformat:13)
space-around-arrays=true
single-case=compact
sequence-style=terminator
sequence-blank-line=preserve-one
parse-toplevel-phrases=false
parse-docstrings=false (file .ocamlformat:12)
parens-tuple-patterns=multi-line-only
parens-tuple=always
parens-ite=false
ocp-indent-compat=false
nested-match=wrap
module-item-spacing=preserve (file .ocamlformat:11)
max-indent=68
match-indent-nested=never
match-indent=0
margin=80
line-endings=lf
let-module=compact
let-binding-spacing=compact
let-binding-indent=2
let-and=compact
leading-nested-match-parens=false
infix-precedence=indent
indicate-nested-or-patterns=unsafe-no
indicate-multiline-delimiters=no
indent-after-in=0
if-then-else=fit-or-vertical (file .ocamlformat:10)
function-indent-nested=never
function-indent=2
field-space=loose
extension-indent=2
exp-grouping=preserve (file .ocamlformat:9)
dock-collection-brackets=false (file .ocamlformat:7)
doc-comments-tag-only=fit (file .ocamlformat:6)
doc-comments-padding=2
doc-comments=before (file .ocamlformat:4)
disambiguate-non-breaking-match=false
cases-matching-exp-indent=normal
cases-exp-indent=2 (file .ocamlformat:5)
break-struct=natural (file .ocamlformat:3)
break-string-literals=auto
break-sequences=true
break-separators=after
break-infix-before-func=false
break-infix=wrap
break-fun-sig=wrap
break-fun-decl=wrap
break-collection-expressions=fit-or-vertical
break-cases=fit-or-vertical (file .ocamlformat:2)
break-before-in=fit-or-vertical
assignment-operator=end-line
```

You can see that some of the values were correctly overwritten by `.ocamlformat` file. Is this behavior expected, or am I missing something?

PR #1587, opened 2023-03-29
Locally enable ocamlformat

voodoos:
This PR add an ocamlformat configuration file and an empty `.ocamlformat-enable`.

The idea is to progressively format more and more files. When merging PR all changed file should be added to the `.ocamlformat-enable` file and formatted.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,3 +77,9 @@ jobs:
           opam exec -- dune clean
           opam exec -- dune build
           git diff --exit-code
+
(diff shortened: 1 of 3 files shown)


voodoos:
Status: investigating the CI issue

- OCaml-CI doesn't install project deps before running `dune build @fmt`
- In Merlin this command does trigger the generation of some files that break because of unmet dependencies

voodoos:
Found the CI problem. Opened a Dune issue: https://github.com/ocaml/dune/issues/7454

PR #1598, opened 2023-05-02
Elisp: make merlin-construct use the user's configured completion interface

bcc32:
The current `merlin-construct` interface is somewhat awkward.  It requires the user to switch buffers, and forces the use of the old-style built-in completion buffer, rather than using whatever completion interface the user has set up (e.g., ivy, helm).

Also make the following small improvements:
1. Fix some docstring typos
2. Clean up unused variable
3. Signal an error if the hole has no constructors, instead of just presenting an empty list of completions

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@ unreleased
       `construct` command, for consistency (#1599).
     - Add a hook to configure system command for spawning ppxes when Merlin is
       used as a library. (#1585)
+  + editor modes
+    - emacs: call the user's configured completion UI in
+      `merlin-construct` (#1598)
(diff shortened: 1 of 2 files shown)


voodoos:
Thank you, the code is much cleaner that way !

@3Rafal could you test it on your setup ?

3Rafal:
@voodoos , **it works on my machine™** :wink:

thanks @bcc32 for this contribution! your implementation works so much better 🙌🏻

voodoos (approved):
LGTM.

bcc32:
Done, and rebased.  Thanks!

PR #1599, opened 2023-05-05
Construct bools as false instead of true

bcc32:
Other primitive types are constructed as "zero" or "empty" values, e.g., ints are `0`, strings are `""`.  By analogy, `bool` should be constructed as `false` rather than `true` (or it should propose both).


voodoos (changes requested):
That would indeed be more coherent, thanks for the change.

Could you update the tests accordingly and add a changelog entry ?

bcc32:
Thanks, done.  I got the following error when running the tests, but they also happen on my machine at before this PR, so I don't think it's related:

```
❯ dune runtest 
File "tests/test-dirs/with-ppx.t/run.t", line 1, characters 0-0:
-tests/test-dirs/with-ppx.t/run.t
+tests/test-dirs/with-ppx.t/run.t.corrected
File "tests/test-dirs/with-ppx.t/run.t", line 9, characters 0-1:
 The ppx works as expected without any typed-hole:
   $ cat >main.ml <<EOF
   > match Some 3 with
   > | None -> ()
   > | Some _ -> print_int [%get_int 42]
   > EOF
 
   $ dune exec ./main.exe 2>/dev/null
-  42
+  [1]
 
   $ $MERLIN single errors \
   > -filename main.ml < main.ml 
+  sh: $TESTCASE_ROOT/_build/default/.ppx/a56cb746ce56d7c281c7d9796f4166ed/ppx.exe: No such file or directory
   {
     "class": "return",
!    "value": [
!      {
!        "start": {
!          "line": 3,
!          "col": 24
!        },
!        "end": {
!          "line": 3,
!          "col": 31
!        },
!        "type": "typer",
!        "sub": [],
!        "valid": true,
!        "message": "Uninterpreted extension 'get_int'."
!      }
!    ],
     "notifications": []
   }
 
 and with type-holes (since #1503)
   $ cat >main.ml <<EOF
   > match Some 3 with
   > | None -> _
   > | Some _ -> print_int [%get_int 42]
   > EOF
 
   $ $MERLIN single errors \
   > -filename main.ml < main.ml 
+  sh: $TESTCASE_ROOT/_build/default/.ppx/a56cb746ce56d7c281c7d9796f4166ed/ppx.exe: No such file or directory
   {
     "class": "return",
!    "value": [
!      {
!        "start": {
!          "line": 3,
!          "col": 24
!        },
!        "end": {
!          "line": 3,
!          "col": 31
!        },
!        "type": "typer",
!        "sub": [],
!        "valid": true,
!        "message": "Uninterpreted extension 'get_int'."
!      }
!    ],
     "notifications": []
   }
```

voodoos:
Thanks! 

Yes, these tests depend on ppxlib, I guess it's not installed on your switch.

voodoos (approved):

Issue #1604, opened 2023-05-15
`make build` overwrites preprocess files


3Rafal:
Each time I run `make build` command on a clean branch I get following files to be overwritten:
```
src/ocaml/preprocess/menhirLib.ml
src/ocaml/preprocess/menhirLib.mli
src/ocaml/preprocess/parser_raw.ml
```

Diff: 
[my_diff.txt](https://github.com/ocaml/merlin/files/11476374/my_diff.txt)

This is a bit confusing and painful... 
I've found a workaround, which is to run `dune build -p merlin` which was suggested in a related issue: #1479 
I wonder if there is a way to fix it? Or maybe we should just edit a Makefile or/and add a note in readme?

voodoos:
My "workaround" is to pin the correct version of Menhir: `opam pin menhir 20201216`

PR #1605, opened 2023-05-15
Add readme note about menhir pin

3Rafal:
As suggested by @voodoos in #1604


README.md
```diff
@@ -41,7 +41,10 @@ dune build -p dot-merlin-reader,merlin
 
 Note: if you want to work on Merlin, you'll want to avoid the `-p merlin` to
 build in dev mode, with some extra warnings enabled. In that case, you'll also
-need an extra dependency: Menhir.
+need an extra dependency: Menhir. You should pin it by running:
```
voodoos, line 44: "We recommend that you pin it to version `20201216` which is the version of Menhir that was used to generate the parser  that is currently present in the sources.

Issue #1529, opened 2022-11-22
Memory leak on completion


art-w:
With OCaml 4.14 and the latest merlin (`4.6-414` and git master (*)), when editing a file in the Irmin repository and querying for completion, the `ocamlermlin-server` quickly reaches >5go of memory usage in just a few seconds.

To reproduce, it's enough to edit [this test file from the Irmin repository](https://github.com/mirage/irmin/blob/main/test/irmin-pack/test_tree.ml) and insert a line anywhere, even at the very top:

```ocaml
let _ = Irmin_pack_unix. (* repeatedly ask for completion *)
```

From a quick look with @voodoos, it's not super clear if the typechecker is responsible for holding onto memory for too long, but it contributes to the size of the leak. For example in the test file, removing the functor instantiations in `module Store = ...` lowers how fast merlin is leaking by a lot. I'm guessing other projects are affected, but that the increasing memory consumption isn't noticeable if one doesn't abuse functors (?)

(*) I believe that @voodoos reproduced the issue on an older version of ocaml/merlin that predates the module shapes

voodoos:
I think I found the culprit !

There is a memoization mechanism for functors:
```
and functor_components = {
  fcomp_arg: functor_parameter;
  (* Formal parameter and argument signature *)
  fcomp_res: module_type;               (* Result signature *)
  fcomp_shape: Shape.t;
  fcomp_cache: (Path.t, module_components) Hashtbl.t;  (* For memoization *)
  fcomp_subst_cache: (Path.t, module_type) Hashtbl.t
}
```

Deactivating that memoization makes the typing very slow on Irmin but stabilize memory consumption.

PR #1609, opened 2023-05-17
Cleanup functor caches

let-def:
This PR adds a mechanism to track the stamp of functor components cached in `fcomp_cache` and `fcomp_subst_cache`. They can they be removed from the cache when backtracking.

Its probably only necessary to track the components added to the persistent_env caches, this feature might be added later if this patch makes things too slow (or just worsen the problem :D).

### DIFF
--- a/src/ocaml/typing/env.ml
+++ b/src/ocaml/typing/env.ml
@@ -496,6 +496,28 @@ type type_descriptions = type_descr_kind
 
 let in_signature_flag = 0x01
 
+let stamped_changes =
+  s_table Stamped_hashtable.create_changes ()
+
+let stamped_add table path value =
+  let rec path_stamp = function
+    | Pident id -> Ident.stamp id
+    | Pdot (t, _) -> path_stamp t
+    | Papply (t1, t2) -> Int.max (path_stamp t1) (path_stamp t2)
+  in
+  let stamp = path_stamp path in
+  let stamp = if stamp = 0 then None else Some stamp in
+  Stamped_hashtable.add table ?stamp path value
+
+let stamped_mem table path =
+  Stamped_hashtable.mem table path
+
+let stamped_find table path =
+  Stamped_hashtable.find table path
+
+let stamped_create n =
+  Stamped_hashtable.create !stamped_changes n
+
 type t = {
   values: (value_entry, value_data) IdTbl.t;
   constrs: constructor_data TycompTbl.t;
@@ -556,8 +578,8 @@ and functor_components = {
   (* Formal parameter and argument signature *)
   fcomp_res: module_type;               (* Result signature *)
   fcomp_shape: Shape.t;
-  fcomp_cache: (Path.t, module_components) Hashtbl.t;  (* For memoization *)
-  fcomp_subst_cache: (Path.t, module_type) Hashtbl.t
+  fcomp_cache: (Path.t, module_components) Stamped_hashtable.t;  (* For memoization *)
+  fcomp_subst_cache: (Path.t, module_type) Stamped_hashtable.t;
 }
 
 and address_unforced =
@@ -1008,7 +1030,7 @@ let modtype_of_functor_appl fcomp p1 p2 =
   | Mty_alias _ as mty -> mty
   | mty ->
       try
-        Hashtbl.find fcomp.fcomp_subst_cache p2
+        stamped_find fcomp.fcomp_subst_cache p2
       with Not_found ->
         let scope = Path.scope (Papply(p1, p2)) in
         let mty =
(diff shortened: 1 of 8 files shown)


voodoos:
This PR should fix #1529

src/ocaml/typing/env.ml
```diff
@@ -496,6 +496,28 @@ type type_descriptions = type_descr_kind
...
+  Stamped_hashtable.add table ?stamp path value
+
+let stamped_mem table path =
+  Stamped_hashtable.mem table path
+
+let stamped_find table path =
+  Stamped_hashtable.find table path
+
+let stamped_create n =
+  Stamped_hashtable.create !stamped_changes n
```
voodoos, line 519: I guess the answer to my previous comment is: "yes but we need to distinguish between the records made when typing different buffers, this way they will each have their own set of changes from their local store"

src/utils/stamped_hashtable.mli
```diff
@@ -0,0 +1,13 @@
+type ('a, 'b) t
```
voodoos, line 1: I think we could add a bit more documentation to this module and the reason it exists.

voodoos (approved):
Thanks for the patch @let-def !
I have a few questions, but the changes looks good to me.

Also I have confirmation that it fixes a usability issue in the Frama-C codebase: https://github.com/ocaml/ocaml-lsp/issues/1032#issuecomment-1556872227

Do you think these changes could be accepted upstream ? (I can make the PR)
Also, could you add a changelog entry ?

src/ocaml/typing/env.ml
```diff
@@ -496,6 +496,28 @@ type type_descriptions = type_descr_kind
...
+
+let stamped_add table path value =
+  let rec path_stamp = function
+    | Pident id -> Ident.stamp id
+    | Pdot (t, _) -> path_stamp t
+    | Papply (t1, t2) -> Int.max (path_stamp t1) (path_stamp t2)
+  in
+  let stamp = path_stamp path in
+  let stamp = if stamp = 0 then None else Some stamp in
+  Stamped_hashtable.add table ?stamp path value
```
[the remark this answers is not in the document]
let-def: There are two issues, they might be addressable but I didn't find how:

1) We don't just need to iterate on the keys, we also need to iterate on the hashtables, I am not sure what would be a simple way to iterate on all `fcomp_cache`/`fcomp_subst_cache` in `Persistent_env`?
This is why I introduced the `stamped_changes` indirection: it gathers the pairs `(table, key)` for all the recently added stamped keys.

2) Even if we find a way to iterate on all tables and keys, the amortized complexity will change from something like $O(\Delta n)$ to $O(n)$, which is probably something we want to avoid if we want to care about performance. The logic in `Stamped_hashtable` is careful to maximize the work that can be shared between snapshots to avoid doing redundant computation. 

If we find a better way to identify precisely the tables in Persistent_env, a simpler design might be possible.

src/utils/stamped_hashtable.mli
let-def: Will do :).

src/ocaml/typing/env.ml
let-def: Not sure which comment you are referring to :)?

let-def:
> Do you think these changes could be accepted upstream ?

I am not sure about that. The issue is very specific to Merlin.
It will never be problematic to the batch compiler, and an artificial example could show that the top-level suffers from the same problem but I am not sure it will be convincing enough to justify the added complexity.

The patch is designed to not be to intrusive, and so "easy" to port across different versions of Merlin and OCaml:
- mostly of the change is code addition,
- the only changes to existing code are the direct references to `fcomp_cache` and `fcomp_subst_cache` that needs to refer to `Stamped_hashtable` rather than `Hashtbl`.

Maybe a middle-ground could be to make the compiler defines aliases `Stamped_hashtable = Hashtbl` such that Merlin only needs to replace the `Stamped_hashtable`?

I am conjecturing a lot, we should talk to the compiler developers to get their actual thoughts :).

> (I can make the PR) Also, could you add a changelog entry ?

Yep.

src/ocaml/typing/env.ml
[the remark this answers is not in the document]
let-def: To summarize:
(1) unless we find a way to iterate the tables, we need a separate structure to keep track of the changes
(2) by keeping it lazily sorted, we get strong guarantees on the performance

(1) complicates the interface, which can impact maintenance.
(2) complicates the implementation, but it has to be done only once, so it's mostly free now :).

src/ocaml/typing/env.ml
voodoos: I was referring to https://github.com/ocaml/merlin/pull/1609#discussion_r1200284145

let-def: Ah yes. Indeed, we have to keep track of the changes for each buffer independently.
Though the hashtables are already buffer local, so if we could find a way to quickly iterate them, the changelog would be only needed for perf purpose.

Issue #1610, opened 2023-05-17
Merlin jumps to incorrect position (or fails to jump) for type that comes from functor


ddickstein:
Tested on 8f1d3f1be970663495a21b83c66bac9934351c82
```
module type T = sig
  type 'a t
end

module M (T : T) = struct
  type t = int T.t
end

module N (T : T) = struct
  type t = M(T).t
end
```

If you try to jump to the definition of `t` inside functor `N`, it jumps to the current type definition, even though this is not a recursive definition of `t`. If I specify `nonrec` and then try it, I get `"Not in environment 't'"`.

voodoos:
Status: 
- this is still reproducible
- it is possible that the new location information in the typedtree for longident will allow us to improve the situation ? Thats unclear.

PR #1611, opened 2023-05-22
Illustrate issue #1610 

voodoos:
This PR add tests illustrating issue #1610
There are two things leading to this: 
1. `reconstruct_identifer` looks broken on module paths
2. Locate used `Longident.parse` to parse which is also broken on module path (and infix operators).

This PR improve 2 by using `Parser.parse_longident`. This make locate work when the path `M(C).t` is manually inputted by the user. This also fixes another issue related to infix operators.

I will keep that PR as a draft until 1 is also fixes (and thus the issue).


voodoos:
@let-def I tried to get `reconstruct_identifier` to work with `Papply` components in [0d8bdd9](https://github.com/ocaml/merlin/pull/1611/commits/0d8bdd9bf017771597970d69ae6752b0f86cd24a) but it's not very satisfying. I guess we should really have a different output type than a flat list ? Have you thought about it already ? I am surprised that this was not handled so far.

In any case, the current hacky solution works on the test suite but it has an unexpected side-effect:
When looking fo the type of `t` in `M(N).t` reconstruct identifier now correctly returns the list of reconstructed identifiers: `["M(N).t"]` instead of an empty list. However parsing that expression fails: `Parser_raw.parse_expression` mistakenly returns `Pexp_construct` and `Pexp_field` nodes and so type-enclosing is wrong. Do you now how I could fix this ?

PR #1613, opened 2023-05-24
Fix text in test

ddickstein:
Text in test referenced -use-ppx-cache flag, which was changed to USE_PPX_CACHE directive.


voodoos (approved):
Thank you !

PR #1614, opened 2023-05-25
Add a CI that check a PR's compatibility with ocaml-lsp

voodoos:


tmattio:
Would it make sense to do it the other way around: clone LSP and pin merlin, so we can run LSP tests in addition to testing that it builds?

voodoos:
> Would it make sense to do it the other way around: clone LSP and pin merlin, so we can run LSP tests in addition to testing that it builds?

Thats a good idea, yes !

PR #1616, opened 2023-05-30
Add changelog check to CI

3Rafal:
This PR adds changelog check action, that fails CI when there is no entry to `CHANGES.md`.
For PRs that doesn't contain changes that should be logged, there should be a label `no changelog`. It needs to be added to merlin project by a maintainer, so that contributors can use it.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,13 @@
+name: Changelog check
+
+on:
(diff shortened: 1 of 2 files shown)


voodoos:
✅✅✅✅🟠✅

voodoos:
Thanks @3Rafal !

PR #1622, opened 2023-06-08
Handle concurrent server start

### LINKED ISSUE #1621: Concurrent server queries race condition - bind fails with "File exists"
Bug occurs on macOS.
Reproduction:
1. Cold start - make sure there is no running instance of ocamlmerlin-process
2. Run below command in terminal and see an error
```
rafal@machine ~ % ocamlmerlin server -version & ocamlmerlin server -version
[1] 86823
bind: File exists
merlin path: /Users/rafal/.opam/default/bin/ocamlmerlin-server
socket path: /var/folders/99/ktsgp3fs5n5cxlw0syckw9g80000gn/T//ocamlmerlin_501_16777235_61952075.socket
The Merlin toolkit version v4.9-414, for Ocaml 4.14.1
[1]  + done       ocamlmerlin server -version
```

A reason seems to be that both processes try to spawn the server and bind to the same socket. One wins and creates a server, another one fails with `bind: File exists`.

3Rafal:
Fixes #1621 

```
rafal@machine ~ % ocamlmerlin server -version & ocamlmerlin server -version
[1] 89118
The Merlin toolkit version %VERSION%, for Ocaml 4.14.1
The Merlin toolkit version %VERSION%, for Ocaml 4.14.1
[1]  + done       ocamlmerlin server -version
```


3Rafal:
@let-def , could you please review this? :) I think you were the author of code that changed

let-def (approved):

let-def:
This seems good, thanks.

3Rafal:
I have removed cram test, because it can't reproduce the behavior. Running with and without changes yields same result.
See:
https://github.com/ocaml/merlin/pull/1623

voodoos:
Great, thanks for the patch @3Rafal  and the review @let-def !

Not sure why the CI is failing but it looks unrelated.

PR #1625, opened 2023-06-09
chore: add a nix flake

rgrinberg:
This setup allows one to setup a development environment with a single command that's guaranteed to succeed (if you have nix):

```
$ nix develop
$ dune runtest
$ code . # not included, insert your favorite editor here
```

Sharing this because I know @3Rafal uses nix. Perhaps he will find it useful.

This can also be turned to a very quick CI with the appropriate github action.


voodoos:
By "very quick" you mean faster than the one based on `setup-ocaml` ? I see that you still use both in `OCaml-LSP`, what is the reason (at least for the unix-based ones) ?

3Rafal:
Nice, thanks @rgrinberg :)

voodoos (approved):
> Just because I know that our users install lsp via opam. So I want to make sure I'm testing the setup that they're commonly using.

That is sensible. I think adding a nix-dev environment is a good addition. I don't really understand how it works and what maintenance will be required however. (When the dependencies change I guess ? Also precise dependencies are not detailed in the file, does it use opam at some point to get these ? etc.) 

Maybe we could have a call (with @rgrinberg or @3Rafal ) at some point so you can explain me what is the stack and what happens under the hood :-)

rgrinberg:
> When the dependencies change I guess ? Also precise dependencies are not detailed in the file, does it use opam at some point to get these ? etc.

Precise dependencies are in fact specified by the lock file. The lock file specifies the version of nixpkgs, which has particular versions of all the dependencies we've specified (ocaml, csexp, yojson, jq, etc.)

> Maybe we could have a call (with @rgrinberg or @3Rafal ) at some point so you can explain me what is the stack and what happens under the hood :-)

Good idea. You might want to install nix before hand so we can walk you through using it.

PR #1626, opened 2023-06-14
Emacs: fix logging when merlin call is interrupted

bcc32:
Currently, "interrupted" is only logged when the result parsing is interrupted (by C-g), but it seems pretty clear that the original intent (and more useful behavior) was to log "interrupted" when the actual merlin call was interrupted.

We do this by setting up separate condition handlers for (1) the result parse failing; and (2) the merlin call or result parse being interrupted.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@ unreleased
 ==========
   + merlin binary
     - Handle concurrent server start (#1622)
+  + editor modes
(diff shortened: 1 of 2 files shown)


voodoos (approved):
These changes looks ok to me, thanks.

bcc32:
Sure, done.  Thanks for reviewing.

PR #1627, opened 2023-06-15
Benchmark init

3Rafal:


voodoos:
Shouldn't the setup like installing merl-an be specified in the docker file to take advantage of caching ?

voodoos:
Shouldn't the setup like installing merl-an be specified in the docker file to take advantage of caching ?

PR #1628, opened 2023-06-19
Override menhir version in flake.nix

3Rafal:
As described [here](https://github.com/ocaml/merlin/tree/master#compilation), menhir should be pinned to specific version. This PR overrides menhir in flake.nix, so that it matches.

@rgrinberg , could you take a look please? :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,22 +7,20 @@
   outputs = { self, nixpkgs, flake-utils }:
     flake-utils.lib.eachDefaultSystem (system:
       let
-        pkgs = nixpkgs.legacyPackages."${system}";
-        menhirOver = self: _: rec {
-          version = "20201216";
-          src = pkgs.fetchFromGitLab {
-            domain = "gitlab.inria.fr";
-            owner = "fpottier";
-            repo = "menhir";
-            rev = version;
-            sha256 = "sha256:04lnd3qxwma4l5jcv79f9bbl5849l6bhg2rzrrsvdzabdplfrxcb";
-          };
-          name = "ocaml${pkgs.ocaml.version}-${self.pname}-${version}";
-        };
-        menhirLib = pkgs.ocamlPackages.menhirLib.overrideAttrs menhirOver;
-        menhirSdk = pkgs.ocamlPackages.menhirSdk.overrideAttrs menhirOver;
-        menhir = (pkgs.ocamlPackages.menhir.overrideAttrs menhirOver)
-          .overrideAttrs(_: _: { buildInputs = [ menhirLib menhirSdk ]; });
+        pkgs = nixpkgs.legacyPackages."${system}".extend (self: super: {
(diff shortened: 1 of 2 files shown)


rgrinberg:
Looks OK to me, but why not use a flake input to fetch the source?

flake.nix
```diff
@@ -8,6 +8,21 @@
...
+            repo = "menhir";
+            rev = version;
+            sha256 = "sha256:04lnd3qxwma4l5jcv79f9bbl5849l6bhg2rzrrsvdzabdplfrxcb";
+          };
+          name = "ocaml${pkgs.ocaml.version}-${self.pname}-${version}";
+        };
+        menhirLib = pkgs.ocamlPackages.menhirLib.overrideAttrs menhirOver;
+        menhirSdk = pkgs.ocamlPackages.menhirSdk.overrideAttrs menhirOver;
+        menhir = (pkgs.ocamlPackages.menhir.overrideAttrs menhirOver)
+          .overrideAttrs(_: _: { buildInputs = [ menhirLib menhirSdk ]; });
```
anmonteiro, line 25:
```suggestion
        pkgs = nixpkgs.legacyPackages."${system}".extend (self: super: {
          ocamlPackages = super.ocamlPackages.overrideScope' (oself: osuper: {
            menhirLib = osuper.menhirLib.overrideAttrs (_: rec {
              version = "20201216";
              src = pkgs.fetchFromGitLab {
                domain = "gitlab.inria.fr";
                owner = "fpottier";
                repo = "menhir";
                rev = version;
                sha256 = "sha256:04lnd3qxwma4l5jcv79f9bbl5849l6bhg2rzrrsvdzabdplfrxcb";
              };
            });
          });
        });
```

anmonteiro (commented):
- You can just override `menhirLib`, everything uses its `src` (check https://github.com/NixOS/nixpkgs/blob/182d471a23be33536eb130c630be6c70d07c1cb4/pkgs/development/ocaml-modules/menhir/default.nix#L10)
- you're lucky your version works because nothing else in the scope needs this specific menhir version, but it's generally good practice to override menhir in the whole scope (otherwise you could get multiple versions of e.g. menhirLib in the linking closure)

3Rafal:
@anmonteiro , thanks for help and explanation! :) I'm in the process of learning nix, so I was happy to hack something that (luckily) works. It's great that you've managed to correct it.

@rgrinberg, I'm not sure how would the fetching in input work. Could you elaborate?

rgrinberg:
You can specify the menhir repo as a flake input (even though it's not a flake) and then use it as a source for the menhir package. For an example, you can see how I did it for jsonoo https://github.com/ocamllabs/vscode-ocaml-platform/pull/1143/files

I didn't pin the repo, but you can specify the revision in the flake input url.

3Rafal:
@rgrinberg , I've applied your suggestions. thanks! :)

anmonteiro (approved):

PR #1629, opened 2023-06-22
Nix CI

3Rafal:
Fast CI (less than 2 minutes) as mentioned by @rgrinberg in #1625

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,37 @@
+name: "Nix"
+
+on:
+  push:
+    branches: [ master ]
+    paths-ignore:
+      - '**.md'
+      - '**.txt'
+      - '.git*'
+      - 'doc/**'
+      - 'emacs/**'
+      - 'vim/**'
+      - '**/emacs-lint.yml'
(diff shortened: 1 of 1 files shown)


.github/workflows/nix.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        submodules: true
+    - name: nix
+      uses: cachix/install-nix-action@v21
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - run: nix develop -c dune runtest -p merlin-lib,dot-merlin-reader,merlin
```
rgrinberg, line 37: `dune runtest` is enough to run the tests.

A little better would be to do `dune build @check @runtest` to make sure everything builds as well.

3Rafal: Good idea, changed

voodoos:
Right, I am a bit worry about the additional maintenance, but we can try it: it's nice to have a faster CI, but it we should aim toward replacing other CIs by this one :-)

PR #1639, opened 2023-06-26
merl-an benchmark

3Rafal:
This PR introduces benchmarking based on `merl-an`.

Note:
Results are shown for multiple groups, because one commit has used wrong workdir. Currently, we generate only results for buffer-typed.


3Rafal:
Closing, because of problems with current-bench results

Issue #1634, opened 2023-06-26
General Behavior CI


pitag-ha:
Merlin already has quite a few end-to-end tests (which are also run in CI). All of those are behavior-specific: we want to test one concrete behavior of Merlin, create concrete input that's relevant for the behavior in question and test Merlin on that input. Additionally to those tests, we now also want to test Merlin's behavior non-specifically on a large randomized input scale. Similarly as for the benchmark CI, here we can use `merl-an` as well.

## Purpose

We want a CI on big scale that monitors Merlin's behavior on a big amount of randomized (but deterministic) input. A change in behavior doesn't necessarily need to be a regression. The CI serves as an observation tool on general global behavior, not as a regression CI.

## Implementation

`merl-an` already has a command `merl-an behavior` for this General Behavior CI. What's missing is the integration into a CI workflow

### Data to take into account

Result of running `merl-an behavior` on Irmin for a 1 sample on each file we get:
- 50MB (we assume that it will scale linearly)
- 15 mins (we assume that it won't scale linearly, will be faster)

So the "simple approach" of having the tests locally as dune cram tests won't scale by any means. We'll need to write a more individualized CI.

### Next action points

- [ ] Fix `merl-an` to avoid OOM problems. See https://github.com/pitag-ha/merl-an/issues/18.
- [ ] Gather more data to see how the resource consumption scales.
- [ ] Design the CI workflow. We can imagine different approaches:
    - Have a DB to store the result from last time. Create the new results. Diff the new result with the last result (less time and CPU consuming, but more space consuming and requires a DB).
    - Without DB, create both the results from the tip before the PR and the ones based on the PR. Diff the two results (more time and CPU consuming, but simpler).
- [ ] Write a PoC for the CI.
- [ ] Implement the CI.

pitag-ha:
@3Rafal, is there anything you'd add?

pitag-ha:
Implemented in https://github.com/ocaml/merlin/pull/1716

PR #1642, opened 2023-06-28
WIP: error-regression

3Rafal:


regression/run.t
```diff
@@ -0,0 +1,2847 @@
+
+  $ build-irmin >> /dev/null 2>&1
+
+  $ merl-an error-regression -s 1 --data=test-data 2> /dev/null
```
pitag-ha, line 4:
```suggestion
  $ merl-an error-regression --data=test-data 2> /dev/null
```
One sample per file is very little. `merl-an`'s default is 30. It'd be interesting to see how long the run takes with 30 samples per file.

(Side-note: this has reminded me that `merl-an` should distribute the samples better throughout the project by making the number of samples per file depend on the size of the file/AST. I've just opened an issue on `merl-an` to remember: https://github.com/pitag-ha/merl-an/issues/28)

pitag-ha (commented):
To me, this looks very nice, @3Rafal! @voodoos , what do you think?

A couple of general thoughts (that we can think about after merging this):

Even though this CI is more resource efficient than the [General Behavior CI](https://github.com/ocaml/merlin/issues/1634) will be, it will still be resource heavy once we make it run on a significant sample set:
- Increase the number of samples per file (see one of the comments below)
- Add a larger/different code-base than only Irmin to act on.
- Add more Merlin queries to `merl-an`.

### When to run the CI?

Given what I've just said, would you run this CI on every PR? Or would you enable it on concrete PRs via a github flag or similar?

### Which code base to act on?

About the point "Add a larger/different code-base than only Irmin to act on" above: So far, we've chosen the Irmin code-base to act on. The reasoning is that it seems an interesting bode-base because of deep functor depth and wide general module type usage. So that's particularly interesting wrt Merlin performance, but also wrt Merlin behavior. What other OCaml features are interesting to test Merlin behavior and which projects do we know that use them? E.g.
- Classes? Which projects use classes? The first ones that come to my mind are `eio` and `ppxlib`.
- Interesting type stuff? Which projects could we use to test that? My OCaml code-base knowledge is quite limited. I only know [`ppxlib`'s `Version`](https://github.com/ocaml-ppx/ppxlib/blob/main/ast/versions.ml) module with "interesting type stuff". Maybe we can use Sherlocode to find something?
- Something else?

### Which Merlin queries to test?

About the point "Add more Merlin queries to `merl-an`" above: The queries `merl-an` supports so far were chosen with performance in mind. For this CI tha ttests whether Merlin gives a successful response, I can also think of `destruct` and `construct` as interesting. What do you two think?

regression/dune
```diff
@@ -0,0 +1,9 @@
+(env (_
+ (binaries build-irmin)))
```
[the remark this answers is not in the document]
3Rafal: no, it doesn't. big parts of our pipelines aren't cached. I don't know about easy way to do this, but I agree it would be nice.

.github/workflows/regression.yml
```diff
@@ -0,0 +1,48 @@
+name: Regression
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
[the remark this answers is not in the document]
3Rafal: I will change it to master, so that it's consistent with other pipelines

regression/run.t
3Rafal: Also, having -s 30 means, that I have wait so long to get results and slows my dev process.

regression/run.t
```diff
@@ -0,0 +1,2847 @@
...
+
+  $ cat test-data/results.json
+  {"sample_id":3113,"merlin_id":0,"cmd":"ocamlmerlin server errors -filename ./irmin/test/ppx_irmin/test_logs.ml < ./irmin/test/ppx_irmin/test_logs.ml","success":true}
```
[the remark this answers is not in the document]
3Rafal: done

regression/run.t
pitag-ha: > Also, I think that having more than 2,5k tests isn't small.

I agree that even with one sample per file, this isn't small. In fact, that's the reason why I've suggested to make it a bit more realistic: That way, we could find out if this approach scales or if we should consider choosing a different approach.

> You can test bigger runs on different branches.

That sounds good. Have you done that already?

If we merge it like this without investigating more, we just need to be aware that we might end up removing this and implementing a different approach later. I have a tendency to first investigate things and then merge them, but I know that both you and Thibaut like doing it the other way around and that's ok as well.

regression/dune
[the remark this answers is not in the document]
pitag-ha: > I don't know about easy way to do this, but I agree it would be nice.

If we had the script be called by the GH action rather than by Dune, could we make use of GH action caching? (It's a genuine question. I don't know myself either.)

regression/run.t
3Rafal: I have tested this on my machine:
```
-- Results --
machine: Laptop with Intel i7 11 gen
sample size: 30 (default)

time:
real 203m8.524s
user 0m35.541s
sys 0m17.571
```
The results between CI and my dev env are comparable, because we use single thread.

> If we merge it like this without investigating more, we just need to be aware that we might end up removing this and implementing a different approach later.

I'm surprised, that we are considering using different approach at this point. I thought that we already discussed this CI and agreed on its implementation. I don't see why we would need to remove it, instead of simply adding some tweaks.

> I have a tendency to first investigate things and then merge them, but I know that both you and Thibaut like doing it the other way around and that's ok as well.

It's hard for me to agree with this remark. I think I have investigated a lot before creating this PR. What I'm suggesting, is that we already provide a value with this PR, so we can merge it and then improve according to specific needs.

regression/dune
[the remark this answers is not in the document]
3Rafal: Running this script via GH actions / dune seems orthogonal to caching. I run it via dune, because checking out Irmin code in test directory makes `dune test` run Irmin tests as well. This could probably be fixed, but I didn't want to spend too much time on this.

Caching build dependency can be done via GH action, but I'm not sure about an actual implementation. It would be a nice optimisation, but realistically we would shave couple of minutes from the build job that currently takes ~40 minutes. I don't think it 's the most valuable think to spend time on at this moment.

regression/run.t
pitag-ha: Oh, I'm sorry, I wasn't aware you had already tested it with sample-size 30. How big is the test file when you do that?

If you've already investigated enough on your side and are convinced that this approach is the right way to go, I'm happy to believe you and stop the thinking/discussion around the approach itself to focus only on details/concrete improvements. Is that what you were saying?

3Rafal:
```
artifact size:
$ ls -l --block-size=M
total 11M
-rwxr-xr-x 1 opam opam 8M Jun 25 23:43 commands.json
-rwxr-xr-x 1 opam opam 1M Jun 25 23:43 logs.json
-rwxr-xr-x 1 opam opam 2M Jun 25 23:43 results.json
```
It was measured without cmds, so it could be bigger (let's say x2).

I think that this approach looks promising. I can't guarantee that it will scale perfectly to all requirements. I think it's something that we can ask about @voodoos too.

.github/workflows/regression.yml
```diff
@@ -0,0 +1,48 @@
+name: Regression
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
voodoos, line 7: We should restrict the ci's run to changes made to .ml / .mli files

voodoos (commented):
It would be better to put the test in a subfolder of the `tests` folder: `tests/regression`

voodoos:
I think you can restrict the sample testing to `type-enclosing`, `locate` and `occurrences`.

Also, it looks like it would be painful to promote the results manually (and this happens quite a lot).

Is there a way to have a tag `promote-regression-changes` or something like that that makes github action create a commit with the diff ?

PR #1644, opened 2023-06-28
Emacs: Optionally pass -with-values to construct

bcc32:
Add the ability to pass a prefix argument to merlin-construct, to include local values in the options presented.  Alternatively, the user can always include local values by setting
merlin-construct-with-local-values non-nil.


emacs/merlin.el
```diff
@@ -1458,19 +1464,30 @@ strictly within, or nil if there is no such element."
...
+(defun merlin--construct-point (with-local-values point)
+  "Execute a construct at POINT.
+
+If WITH-LOCAL-VALUES is non-nil, pass \"-with-values local\" to
+include local values in the candidate list."
   (when-let ((result (merlin-call "construct"
-                                  "-position" (merlin-unmake-point point))))
+                                  "-position" (merlin-unmake-point point)
+                                  "-with-values"
+                                  (if (or with-local-values merlin-construct-with-local-values)
```
3Rafal, line 1475: Would simplified version work the same?
```lisp
  (when-let ((result (merlin-call "construct"
                                  "-position" (merlin-unmake-point point)
                                  (if with-local-values merlin-construct-with-local-values
                                    "-with-values"
                                    ""))))
```

3Rafal (approved):

voodoos:
Thanks @bcc32 !

PR #1651, opened 2023-07-05
chore: update nix flake

Alizter:


voodoos:
I am not a nix user, but I thought there was no need for an update of the lock file unless the builds was actually broken, is it ?

Alizter:
@voodoos There is no need, but there is a newer version of ocaml-lsp which is always nice to use. The OCaml platform extension for VSCode kindly reminds us of this.

voodoos:
Oh I see, it also installs the tools, fair enough :-)

PR #1653, opened 2023-07-05
test: test enclosing type for let+ ... and+ ... in expression

Alizter:
This demonstrates that type enclosing has the incorrect behaviour for letop's with multiple ands.


voodoos:
First diagnostic: The issue is due to the typedtree nodes not respecting location inclusion / disjointness rules. Some locations issues by the parser are also very imprecise.

voodoos:
I have a first improvement: it is now possible to get the type of the patterns !
On the `let+` and `and+` however the whole expression is selected and it's type shown. 
That's coherent but not very informative.

src/analysis/type_utils.ml
```diff
@@ -30,6 +30,10 @@ open Std
 
 module Verbosity = Mconfig.Verbosity
 
+let protect expr =
```
Alizter, line 33: What does protect do?

tests/test-dirs/type-enclosing/letop2.t/run.t
Alizter, line 1: You've removed some of the tests here, are they no longer needed?

src/analysis/type_utils.ml
voodoos: To get the "generic" type of the operator Merlin needs to parse it and type it. But `let++` is not valid syntax so adding parenthesis `(let++)` is required for it to parse correctly. The protect function uses a function from compiler libs to determine is the expression requires parenthesizing before parsing.

Alizter:
Thanks @voodoos!

PR #1654, opened 2023-07-11
feat(emacs): Add support for opam-switch-mode (>= 1.3)

### LINKED ISSUE #1591: Unexpected error "The current Merlin configuration has been generated by another, incompatible, version of Dune"
Dear all,

I recently reinstalled merlin in a local opam switch for working on [learn-ocaml](https://github.com/ocaml-sf/learn-ocaml),
and I systematically got the following error:

```
The current Merlin configuration has been generated by another, incompatible, version of Dune.
Please rebuild the project. (Using the same version of Dune as the one running the `ocaml-merlin` server.)
(431 pending errors, use C-c C-x to jump)
```

Thus:

* either there is a true incompatibility and an opam constraint should be added in **opam-repository**
* or I'm doing something wrong.

## Additional info

* Here is the list of our dependencies: [learn-ocaml.opam](https://github.com/ocaml-sf/learn-ocaml/blob/master/learn-ocaml.opam), thus requiring OCaml 4.12, and allowing dune upto 2.9.3
* `opam show merlin.4.7-412` currently says `"dune" {>= "2.9.0"}`
* Doing `rm -fr _opam` or `rm -fr _build` and starting over is not enough.
* ~`opam install merlin.4.6-412` solves the issue.~ [actually not :frowning_face:]

Thanks for any feedback allowing to fix this!

erikmd:
## Overall description

This PR adds support for `opam-switch-mode` (initially written by @hendriktews for [proof-general](https://proofgeneral.github.io) then as a separate ELPA package, further extended by @monnier and myself).

This new feature makes it possible to fully fix #1591, given it is now as easy as clicking on either of the "OPSW" menu-bar or the "OPSW" mode-bar, which changes Emacs underlying PATH accordingly (`exec-path`), and triggers the functions registered in the opam-switch-mode hook (typically, `merlin-stop-server`).

See the following screenshot of opam-switch-mode version 1.3 for a glimpse of the minor mode.

2023-07-11_16-04-24_Screenshot_OPSW_menus

## Remarks

* This PR does NOT add opam-switch-mode as a dependency of merlin: it is an optional, accompanying package (only useful for ocaml developers that rely on opam), and it is mostly language-agnostic (apart from using the opam tool).
* The [opam-switch-mode](https://github.com/ProofGeneral/opam-switch-mode) package is available in MELPA, MELPA Stable, NonGNU-devel, [NonGNU](https://elpa.nongnu.org/nongnu/opam-switch-mode.html).
* ~~It requires to use `(setq merlin-command "ocamlmerlin")`, but I was wary of backward-compatibility, so I didn't change the default value of `merlin-command` but documented the requirement instead.~~ this constraint is now lifted.
* The only `Package-Requires:` of opam-switch-mode is `(emacs "25.1")`, which matches with Merlin's ones.

## Details of the commit

* Add option `merlin-stop-server-on-opam-switch` (default: t) and related function `(merlin-stop-server-on-opam-switch)` that kills the merlin server just before we change the opam switch using the OPSW menu-bar or the OPSW mode-bar from opam-switch-mode.

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -605,6 +605,35 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
         (kill-local-variable 'merlin-buffer-configuration)
         (kill-local-variable 'merlin-erroneous-buffer)))))
 
+(defcustom merlin-stop-server-on-opam-switch t
+  "If t, stops the Merlin server before the opam switch changes.
+If the user changes the opam switch using `opam-switch-set-switch'
+or an `\"OPSW\"' menu from `opam-switch-mode', this option asks to
+stop the Merlin server process, so that the next Merlin command
+starts a new server, typically with a different Merlin version
+from a different opam switch.
+
+See https://github.com/ProofGeneral/opam-switch-mode
+
+Note: `opam-switch-mode' triggers automatic changes for `exec-path' and
+`process-environment', which are useful to find the `\"ocamlmerlin\"'
+binary (its filename can be overriden in `merlin-command') and the
+binary of Merlin's subprocesses, in the ambient opam switch."
+  :type 'boolean)
+
+(defun merlin--stop-server-on-opam-switch ()
+  "Stop the Merlin server before the opam switch changes.
+This function is for the `opam-switch-mode' hook
+`opam-switch-before-change-opam-switch-hook', which runs just
+before the user changes the opam switch through `opam-switch-mode'."
+  (when (and merlin-mode merlin-stop-server-on-opam-switch)
+    (condition-case _sig
+        (merlin-stop-server)
+      (t (message "Info: (merlin-stop-server) failed in the previous opam switch")))))
+
+(add-hook 'opam-switch-before-change-opam-switch-hook
+          #'merlin--stop-server-on-opam-switch t)
+
 ;;;;;;;;;;;;;;;;;;;;
 ;; FILE SWITCHING ;;
 ;;;;;;;;;;;;;;;;;;;;
@@ -1925,16 +1954,12 @@ Empty string defaults to jumping to all these."
 
                  ;; this was originally done via `opam exec' but that does not
                  ;; work for opam 1, and added a performance hit
-                 (setq bin-path (list (concat "PATH=" bin-dir)))
-                 "ocamlmerlin")
-
+                 (setq bin-path (list (concat "PATH=" bin-dir))))
              ;; best effort if opam is not available, lookup for the binary in
              ;; the existing env
-             (progn
-               (message "merlin-command: opam var failed (%S)"
-                        (buffer-string))
-               "ocamlmerlin"))))))
-
+             (message "merlin-command: opam var failed (%S)"
+                      (buffer-string)))
+           "ocamlmerlin"))))
       ;; cache command in merlin-buffer configuration to avoid having to shell
       ;; out to `opam` each time.
       (push (cons 'command command) merlin-buffer-configuration)
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -258,8 +258,6 @@ The association list can contain the following optional keys:
 ;; Internal variables ;;
 ;;;;;;;;;;;;;;;;;;;;;;;;
 
-(defvar merlin-opam-bin-path nil)
-
 ;; If user did not specify its merlin-favourite-caml-mode, try to guess it from
 ;; the buffer being edited
 (defvar merlin-guessed-favorite-caml-mode nil)
@@ -512,7 +510,13 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
   ;; Really start process
   (let ((binary      (merlin-command))
         ;; (flags       (merlin-lookup 'flags merlin-buffer-configuration))
-        (process-environment (cl-copy-list process-environment))
+        (process-environment
+         ;; for simplicity, we use a mere append here (leading to a
+         ;; duplicate binding), it does work because only the first
+         ;; occurrence is considered, one can check this by running
+         ;; (call-process "printenv" nil t)
+         (append (merlin-lookup 'env merlin-buffer-configuration)
+                 process-environment))
         (dot-merlin  (merlin-lookup 'dot-merlin merlin-buffer-configuration))
         ;; FIXME use logfile
         ;; (logfile     (or (merlin-lookup 'logfile merlin-buffer-configuration)
@@ -522,16 +526,6 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
         (packages    (merlin--map-flatten (lambda (x) (cons "-I" x))
                                           merlin-buffer-packages-path))
         (filename    (buffer-file-name (buffer-base-buffer))))
-    ;; Update environment
-    (dolist (binding (merlin-lookup 'env merlin-buffer-configuration))
-      (let* ((equal-pos (string-match-p "=" binding))
-             (prefix (if equal-pos
-                       (substring binding 0 (1+ equal-pos))
-                       binding))
-             (is-prefix (lambda (x) (string-prefix-p prefix x))))
-        (setq process-environment (cl-delete-if is-prefix process-environment))
-        (when equal-pos
-          (setq process-environment (cons binding process-environment)))))
     ;; Compute verbosity
     (when (eq merlin-verbosity-context t)
       (setq merlin-verbosity-context (cons command args)))
@@ -1910,7 +1904,8 @@ Empty string defaults to jumping to all these."
             (merlin-lookup 'do-not-cache-config merlin-buffer-configuration))
     (setq merlin-buffer-configuration (merlin--configuration)))
 
-  (let ((command (merlin-lookup 'command merlin-buffer-configuration)))
+  (let ((command (merlin-lookup 'command merlin-buffer-configuration))
+        bin-path)
     (unless command
       (setq
        command
@@ -1921,7 +1916,7 @@ Empty string defaults to jumping to all these."
          (with-temp-buffer
            (if (eq (call-process-shell-command
                     "opam var bin" nil (current-buffer) nil) 0)
-               (let ((bin-path
+               (let ((bin-dir
                       (replace-regexp-in-string "\n$" "" (buffer-string))))
                  ;; the opam bin dir needs to be on the path, so if merlin
                  ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
@@ -1930,8 +1925,8 @@ Empty string defaults to jumping to all these."
 
                  ;; this was originally done via `opam exec' but that does not
                  ;; work for opam 1, and added a performance hit
-                 (setq merlin-opam-bin-path (list (concat "PATH=" bin-path)))
-                 (concat bin-path "/ocamlmerlin"))
+                 (setq bin-path (list (concat "PATH=" bin-dir)))
+                 "ocamlmerlin")
 
              ;; best effort if opam is not available, lookup for the binary in
              ;; the existing env
@@ -1943,8 +1938,8 @@ Empty string defaults to jumping to all these."
       ;; cache command in merlin-buffer configuration to avoid having to shell
       ;; out to `opam` each time.
       (push (cons 'command command) merlin-buffer-configuration)
-      (when merlin-opam-bin-path
-        (push (cons 'env merlin-opam-bin-path) merlin-buffer-configuration)))
+      (when bin-path
+        (push (cons 'env bin-path) merlin-buffer-configuration)))
 
     command))
 
--- a/merlin.opam
+++ b/merlin.opam
@@ -51,8 +51,11 @@ Add opam emacs directory to your load-path by appending this to your .emacs:
     (add-hook 'tuareg-mode-hook 'merlin-mode t)
     (add-hook 'caml-mode-hook 'merlin-mode t)
     ;; Use opam switch to lookup ocamlmerlin binary
-    (setq merlin-command 'opam)))
-
+    (setq merlin-command 'opam)
+    ;; To easily change opam switches within a given Emacs session, you can
+    ;; install the minor mode https://github.com/ProofGeneral/opam-switch-mode
+    ;; and use one of its \"OPSW\" menus.
+    ))
 Take a look at https://github.com/ocaml/merlin for more information
 
 Quick setup with opam-user-setup
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@ unreleased
       local values in the completion options.  Alternatively, this
       behavior can be enabled permanently by customizing
       `merlin-construct-with-local-values` (#1644)
+    - emacs: add support for opam-switch-mode (#1654, fixes #1591).
+      See <https://github.com/ProofGeneral/opam-switch-mode>
 
 merlin 4.9
 ==========
--- a/README.md
+++ b/README.md
@@ -105,6 +105,8 @@ Just add the following to your *.emacs* file:
 ;; (require 'merlin-iedit)       ; iedit.el editing of occurrences
 ;; (require 'merlin-company)     ; company.el completion
 ;; (require 'merlin-ac)          ; auto-complete.el completion
+;; To easily change opam switches and pick the ocamlmerlin binary accordingly,
+;; you can use the minor mode https://github.com/ProofGeneral/opam-switch-mode
 ```
 
 More comprehensive documentation can be found on the [emacs-from-scratch wiki](https://github.com/ocaml/merlin/wiki/emacs-from-scratch).
@@ -127,6 +129,8 @@ Emacs startup file is sufficient:
 ;; (require 'merlin-iedit)       ; iedit.el editing of occurrences
 ;; (require 'merlin-company)     ; company.el completion
 ;; (require 'merlin-ac)          ; auto-complete.el completion
+;; To easily change opam switches and pick the ocamlmerlin binary accordingly,
+;; you can use the minor mode https://github.com/ProofGeneral/opam-switch-mode
 ```
 
 ### Other Editors


monnier:
> This PR adds support for `opam-switch-mode` (initially written by
> @hendriktews for [proof-general](https://proofgeneral.github.io) then
> as a separate ELPA package, further extended by @monnier and myself).

Thanks, a few comments below:

> +(defcustom merlin-stop-server-on-opam-switch t
> +  "If t, stop ocamlmerlin server before the opam switch changes.
> +If `(opam-switch-mode)' is loaded and the user changes the opam

Why parentheses around `opam-switch-mode`?

> +switch using `opam-switch-set-switch' or an `\"OPSW\"' menu, this

Actually, scratch my previous comment: if the user changes the
opam switch with `opam-switch-set-switch`, then `opam-switch-mode`
*is* loaded&available&installed.

> +option asks to stop the merlin server process, so that the next
> +merlin command starts a new server, typically with a different
> +merlin version from a different opam switch.

[ FWIW, I think Merlin deserves to be capitalized :-)  ]

> +Based on the minor mode `opam-switch-mode', see:
> +https://github.com/ProofGeneral/opam-switch-mode

"Based on" doesn't sound right.  I think I'd just say:

    See https://github.com/ProofGeneral/opam-switch-mode

> +Note: `(opam-switch-mode)' expects `merlin-command' to be `\"ocamlmerlin\"'."

[ Why the parens, again?  ]
What happens if `merlin-command' is something else?
Also, is it really `opam-switch-mode` which expects it?
My impression is that it's this PR's code which expects it.
If the code doesn't work when `merlin-command' is something else,
then maybe the default value should not be t but:

    (equal merlin-command "ocamlmerlin")

Better yet: can we make `opam-switch-mode` work right when
`merlin-command` is `opam`?

> +  :group 'merlin :type 'boolean)

The `:group` arg is redundant.

> +(defun merlin-stop-server-on-opam-switch ()
> +  "Stop ocamlmerlin before the opam switch changes.
> +This function is for the `opam-switch-mode' hook
> +`opam-switch-before-change-opam-switch-hook', which runs just
> +before the user changes the opam switch through
> +`opam-switch-mode'.
> +
> +If `merlin-stop-server-on-opam-switch' is t, then the merlin
> +server process is stopped, so that the next merlin command starts
> +a new server, typically with a different merlin version from a
> +different opam switch.
> +
> +Based on the minor mode `opam-switch-mode', see:
> +https://github.com/ProofGeneral/opam-switch-mode
> +
> +Note: `(opam-switch-mode)' expects `merlin-command' to be `\"ocamlmerlin\"'."

AFAICT this is an internal function, so I'd use a "merlin--" prefix and
I'd cut down the docstring significantly (especially to try and reduce
duplication): no need to repeat what the code says fairly clearly.

> +  (when (and merlin-mode merlin-stop-server-on-opam-switch)
> +    (condition-case sig
> +        (merlin-stop-server)
> +      (t (message "Info: (merlin-stop-server) failed in the previous opam switch")))))

Unused variable `sig`.  Maybe include its value in the message?

> +(add-hook 'opam-switch-before-change-opam-switch-hook #'merlin-stop-server-on-opam-switch t)

Cut the line to better fit within the usual 80 columns?


        Stefan

erikmd:
Thanks @monnier for your blazingly fast review! I'll address your remarks soon.

— BTW the code of this PR is similar to https://github.com/ocaml/tuareg/pull/307, so maybe, no need to review the other one yet… and I'll take your remaks into account for both!

erikmd:
Dear @monnier, thanks for your remarks! Here are some feedback:

>  +(defcustom merlin-stop-server-on-opam-switch t
>  +  "If t, stop ocamlmerlin server before the opam switch changes.
>  +If `(opam-switch-mode)' is loaded and the user changes the opam
> Why parentheses around `opam-switch-mode`?

Good point, it was preventing navigation in Emacs help buffer.

>  +switch using `opam-switch-set-switch' or an `\"OPSW\"' menu, this
> Actually, scratch my previous comment: if the user changes the
> opam switch with `opam-switch-set-switch`, then `opam-switch-mode`
> *is* loaded&available&installed.

Indeed, I just meant that the `opam-switch-set-switch` etc. are not
directly part of merlin.el but are found in an 3rd party feature.

But admittedly this is clear given Emacs' prefix-based naming convention!

So I rephrased this docstring.

>  +option asks to stop the merlin server process, so that the next
>  +merlin command starts a new server, typically with a different
>  +merlin version from a different opam switch.
> [ FWIW, I think Merlin deserves to be capitalized :-)  ]

Indeed! ;-)

>  +Based on the minor mode `opam-switch-mode', see:
>  +https://github.com/ProofGeneral/opam-switch-mode
> "Based on" doesn't sound right.  I think I'd just say:
> 
>     See https://github.com/ProofGeneral/opam-switch-mode

Done.

>  +Note: `(opam-switch-mode)' expects `merlin-command' to be `\"ocamlmerlin\"'."
> [ Why the parens, again?  ]

Sorry ^^

> What happens if `merlin-command' is something else?

Namely:

* if `merlin-command` is a basename, it works also
* if `merlin-command` is an absolute path, there is no "error"
  but opam-switch-mode is useless (as the `exec-path` changes done
  became no-op).
* if `merlin-command` is 'opam, likewise: there is no "error"
  but opam-switch-mode is useless.

> Also, is it really `opam-switch-mode` which expects it?

Yes.

> My impression is that it's this PR's code which expects it.

Nope: the PR code ensures that the previous ocamlmerlin server is
stopped in due time (before the exec-path is changed), and then
the new linting (`C-x C-s`) can start the new ocamlmerlin server
in the new switch

> If the code doesn't work when `merlin-command' is something else,
> then maybe the default value should not be t but:
> 
>     (equal merlin-command "ocamlmerlin")

Not sure that this would be more intuitive.

> Better yet: can we make `opam-switch-mode` work right when
> `merlin-command` is `opam`?

Maybe yes (or maybe not that easily).
Actually this seems a bit beyond the scope of this PR, whose main goal
is to add compatibility w.r.t. opam-switch-mode, which can be viewed
as an enhanced, externalized version of the `'opam` option.
In particular, I didn't want to break backward compatibility or so.

But maybe a **subsequent** PR could be made to deprecate the `'opam`,
or make "ocamlmerlin" the default option, or… just detect that the
opam-switch-mode feature is present, and override merlin-command
to "ocamlmerlin" in this case? What do you think?

(and IINM, the Merlin server binary does not have other typical basenames)

FWIW, note that thanks to this PR #1654, we could dramatically simplify
the .emacs we usually ship in OCaml labwork sessions, because
[these lines](https://github.com/erikmd/tapfa-init.el/blob/b15dcd7a84cf69ca6e11a8366b43bc67443bb285/.emacs#L209-L221):
```
(if tapfa-opam-available
    (use-package merlin
      :ensure t
      :hook
      ((tuareg-mode caml-mode) . merlin-mode)
      :config
      (setq merlin-command 'opam))
  (use-package merlin
    :ensure t
    :hook
    ((tuareg-mode caml-mode) . merlin-mode)
    :config
    (setq merlin-command "ocamlmerlin")))
```
could be replaced with:
```
(use-package merlin
  :ensure t
  :hook
  ((tuareg-mode caml-mode) . merlin-mode)
  :config
  (setq merlin-command "ocamlmerlin"))
```

>  +  :group 'merlin :type 'boolean)
> The `:group` arg is redundant.

Fixed.

>  +(defun merlin-stop-server-on-opam-switch ()
>  +  "Stop ocamlmerlin before the opam switch changes.
>  +  …
>  + Note: `(opam-switch-mode)' expects `merlin-command' to be `\"ocamlmerlin\"'."
> AFAICT this is an internal function, so I'd use a "merlin--" prefix and
> I'd cut down the docstring significantly (especially to try and reduce
> duplication): no need to repeat what the code says fairly clearly.

Thanks, done.

>  +  (when (and merlin-mode merlin-stop-server-on-opam-switch)
>  +    (condition-case sig
>  +        (merlin-stop-server)
>  +      (t (message "Info: (merlin-stop-server) failed in the previous opam switch")))))
> Unused variable `sig`.  Maybe include its value in the message?

Actually I don't think that'd be useful.

Indeed this case handles a very typical error in practice (the current switch
does not have Merlin installed, so at the time we change switches, we cannot
access the ocamlmerlin server, hence the expected error, which deserves a small
message, but not a too verbose message given the error message is not useful…
But, YMMV :)

>  +(add-hook 'opam-switch-before-change-opam-switch-hook #'merlin-stop-server-on-opam-switch t)
> Cut the line to better fit within the usual 80 columns?

Indeed, thanks!

merlin.opam
```diff
@@ -51,8 +51,11 @@ Add opam emacs directory to your load-path by appending this to your .emacs:
     (add-hook 'tuareg-mode-hook 'merlin-mode t)
     (add-hook 'caml-mode-hook 'merlin-mode t)
     ;; Use opam switch to lookup ocamlmerlin binary
-    (setq merlin-command 'opam)))
-
+    (setq merlin-command 'opam)
+    ;; For a better UX you can instead write
+    ;; (setq merlin-command \"ocamlmerlin\")
```
voodoos, line 56: I don't see why `(setq merlin-command 'opam)` would be problematic when using `opam-switch-mode`.
- As far as I understand, `(setq merlin-command 'opam)` means that the path to the Merlin binary will be asked to opam, and opam, according to the currently selected switch, should answer whith the correct binary (regardless of the `$PATH`).
- Using `opam-switch-mode` allows one to easily change the active switch and load the correct `$PATH`. This means that Merlin can indeed be started by calling `ocamlmerlin`, but I don't see why it would prevent the opam-based detection to work ?

emacs/merlin.el
```diff
@@ -611,6 +611,34 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
         (kill-local-variable 'merlin-buffer-configuration)
         (kill-local-variable 'merlin-erroneous-buffer)))))
 
+(defcustom merlin-stop-server-on-opam-switch t
```
voodoos, line 614: What is the point of this configuration option ? If the user decides to use the `opam-switch-mode` we probably always want to stop the server before switching.

voodoos (commented):
Thanks for your contribution @erikmd, I have a few questions:-)

merlin.opam
erikmd: @voodoos thanks for your comments!

> I don't see why (setq merlin-command 'opam) would be problematic when using opam-switch-mode.

> As far as I understand, (setq merlin-command 'opam) means that the path to the Merlin binary will be asked to opam, and opam, according to the currently selected switch, should answer whith the correct binary (regardless of the $PATH).

Yes and no: there is indeed a machinery in merlin@master to guess the Merlin binary when merlin-command = `'opam` **but** it is static: it relies on the current switch chosen at emacs startup, which implies the current behavior is buggy/distracting, see https://github.com/ocaml/merlin/issues/1591 as well as https://github.com/ocaml/merlin/issues/1591#issuecomment-1518894314 for the current workaround :-|

> Using opam-switch-mode allows one to easily change the active switch and load the correct $PATH. This means that Merlin can indeed be started by calling ocamlmerlin, but I don't see why it would prevent the opam-based detection to work ?

I didn't mean merlin's `'opam`-based static finding of ocamlmerlin wouldn't work, I meant opam-switch-mode's dynamic finding of ocamlmerlin by changing the `$PATH` **only** works if the minor mode lookups for a basename (roughly, emacs' equivalent of `$ which ocamlmerlin`)

emacs/merlin.el
erikmd: > we probably always want to stop the server before switching.

Exactly, this is what most users want, so it is the default option. But in some cases, the option might be useful (e.g., if a user starts two emacs instances, initially with the same opam switch, then want to change the switch in one instance… without killing the server for the other instance :)

monnier:
> I didn't mean merlin's `'opam`-based static finding of ocamlmerlin wouldn't
> work, I meant opam-switch-mode's dynamic finding of ocamlmerlin by changing
> the `$PATH` **only** works if the minor mode lookups for a basename
> (roughly, emacs' equivalent of `$ which ocamlmerlin`)

Yeah that code in `merlin.el` is weird.
We could start with a simplification like below (won't fix the problem yet).


        Stefan


(merlin-opam-bin-path): Remove global var since it was only used locally.
(merlin--call-command): Don't reimplement by hand the semantics already
provided by `process-environment`.
(merlin-command): In the `opam` case, return a name relative to $PATH
since we know what $PATH will be set to.

> diff --git a/emacs/merlin.el b/emacs/merlin.el
> index 62e33631..791fe333 100644
> --- a/emacs/merlin.el
> +++ b/emacs/merlin.el
> @@ -258,8 +258,6 @@ The association list can contain the following optional keys:
>  ;; Internal variables ;;
>  ;;;;;;;;;;;;;;;;;;;;;;;;
>  
> -(defvar merlin-opam-bin-path nil)
> -
>  ;; If user did not specify its merlin-favourite-caml-mode, try to guess it from
>  ;; the buffer being edited
>  (defvar merlin-guessed-favorite-caml-mode nil)
> @@ -512,7 +510,9 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
>    ;; Really start process
>    (let ((binary      (merlin-command))
>          ;; (flags       (merlin-lookup 'flags merlin-buffer-configuration))
> -        (process-environment (cl-copy-list process-environment))
> +        (process-environment
> +         (append (merlin-lookup 'env merlin-buffer-configuration)
> +                 process-environment))
>          (dot-merlin  (merlin-lookup 'dot-merlin merlin-buffer-configuration))
>          ;; FIXME use logfile
>          ;; (logfile     (or (merlin-lookup 'logfile merlin-buffer-configuration)
> @@ -522,16 +522,6 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
>          (packages    (merlin--map-flatten (lambda (x) (cons "-I" x))
>                                            merlin-buffer-packages-path))
>          (filename    (buffer-file-name (buffer-base-buffer))))
> -    ;; Update environment
> -    (dolist (binding (merlin-lookup 'env merlin-buffer-configuration))
> -      (let* ((equal-pos (string-match-p "=" binding))
> -             (prefix (if equal-pos
> -                       (substring binding 0 (1+ equal-pos))
> -                       binding))
> -             (is-prefix (lambda (x) (string-prefix-p prefix x))))
> -        (setq process-environment (cl-delete-if is-prefix process-environment))
> -        (when equal-pos
> -          (setq process-environment (cons binding process-environment)))))
>      ;; Compute verbosity
>      (when (eq merlin-verbosity-context t)
>        (setq merlin-verbosity-context (cons command args)))
> @@ -1910,7 +1900,8 @@ Empty string defaults to jumping to all these."
>              (merlin-lookup 'do-not-cache-config merlin-buffer-configuration))
>      (setq merlin-buffer-configuration (merlin--configuration)))
>  
> -  (let ((command (merlin-lookup 'command merlin-buffer-configuration)))
> +  (let ((command (merlin-lookup 'command merlin-buffer-configuration))
> +        bin-path)
>      (unless command
>        (setq
>         command
> @@ -1921,7 +1912,7 @@ Empty string defaults to jumping to all these."
>           (with-temp-buffer
>             (if (eq (call-process-shell-command
>                      "opam var bin" nil (current-buffer) nil) 0)
> -               (let ((bin-path
> +               (let ((bin-dir
>                        (replace-regexp-in-string "\n$" "" (buffer-string))))
>                   ;; the opam bin dir needs to be on the path, so if merlin
>                   ;; calls out to sub binaries (e.g. ocamlmerlin-reason), the
> @@ -1930,8 +1921,8 @@ Empty string defaults to jumping to all these."
>  
>                   ;; this was originally done via `opam exec' but that does not
>                   ;; work for opam 1, and added a performance hit
> -                 (setq merlin-opam-bin-path (list (concat "PATH=" bin-path)))
> -                 (concat bin-path "/ocamlmerlin"))
> +                 (setq bin-path (list (concat "PATH=" bin-dir)))
> +                 "ocamlmerlin")
>  
>               ;; best effort if opam is not available, lookup for the binary in
>               ;; the existing env
> @@ -1943,8 +1934,8 @@ Empty string defaults to jumping to all these."
>        ;; cache command in merlin-buffer configuration to avoid having to shell
>        ;; out to `opam` each time.
>        (push (cons 'command command) merlin-buffer-configuration)
> -      (when merlin-opam-bin-path
> -        (push (cons 'env merlin-opam-bin-path) merlin-buffer-configuration)))
> +      (when bin-path
> +        (push (cons 'env bin-path) merlin-buffer-configuration)))
>  
>      command))

erikmd:
@monnier you said:
> Yeah that code in `merlin.el` is weird.
> We could start with a simplification like below (won't fix the problem yet).

Thanks, but good news, it appears your patch fix the `'opam` problem!

I mean, opam-switch-mode works well even with the default `(setq merlin-command 'opam)`.

I tested this this way:

* <kbd>M-x merlin-mode RET</kbd>
* <kbd>M-x opam-switch-mode RET </kbd>
* <kbd>M-: (setq merlin-command 'opam) RET</kbd>
* <kbd>C-x C-s</kbd>
* `$ ps aux | grep ocamlmerlin`
* <kbd>M-x opam-switch-set-switch system RET</kbd>
* <kbd>C-x C-s</kbd>
* `$ ps aux | grep ocamlmerlin`

As a result, I pushed a commit that removes the constraint from the README/etc., that I had initially added in the PR.

So, do you approve the latest version fo the PR?
— AFAICT, it looks good to me (it fixes #1591, it ensures that opam-switch-mode smoothly integrate with merlin to add the "dynamic choice" regarding opam switch, and it is backward-compatible :)

monnier:
>> -        (process-environment (cl-copy-list process-environment))
> +        (process-environment
> +         (append (merlin-lookup 'env merlin-buffer-configuration)
> +                 process-environment))
>
> Note that your simplification of the (local) modification of
> process-environment (vs. [the older
> code](https://github.com/ocaml/merlin/pull/1654/files#diff-eb86342d6ca81aa067fe94ef9182841e48adbfd73994c7e71f6fa2fe3d737473L525-L534)
> that was complex, but conservative) is a bit abrupt: it can lead to several
> `"PATH=…"` items; and even if on my testing it worked, I am afraid the new
> `PATH` added in front of the code will skip the other useful `PATH`
> directories (maybe used by subprocesses).

That behavior is what the code currently does:

    (dolist (binding (merlin-lookup 'env merlin-buffer-configuration))
      (let* ((equal-pos (string-match-p "=" binding))
             (prefix (if equal-pos
                       (substring binding 0 (1+ equal-pos))
                       binding))
             (is-prefix (lambda (x) (string-prefix-p prefix x))))
        (setq process-environment (cl-delete-if is-prefix process-environment))
        (when equal-pos
          (setq process-environment (cons binding process-environment)))))

(as you can see, the PATH=... item will cause all the other
PATH=... present in `process-environment` to be removed altogether):

My patch does not change it (I agree that it seems problematic).

> So if you don't object, I will revert this small part of your patch.

Be my guest, but I don't think it'll make any difference (other than
keep the code more complex than it needs to be).

erikmd:
Salut Stefan !

I'm not sure I fully understood your reply.

> (as you can see, the PATH=... item will cause all the other
> PATH=... present in `process-environment` to be removed altogether):

I agree.

> My patch does not change it

Really? IIUC, the value of `(length process-environment)` of the current PR, and the length with your initial patch differ.

> > So if you don't object, I will revert this small part of your patch.
> 
> Be my guest, but I don't think it'll make any difference (other than
> keep the code more complex than it needs to be).

Indeed, I didn't expect a difference, but "better safe than sorry"…? 🙃

But I applied all the rest of your patch… Anyway, let me know if you're OK with the current version of the PR! (or if I need to revert the revert :)

Kind regards

monnier:
>> My patch does not change it
> Really? IIUC, the value of `(length process-environment)` of the current PR,
> and the length with your initial patch differ.

Of course, with the original code we start with

    (blabla "PATH=hello" blibli)

and we end with the fresh new list

    ("PATH=/the/opam/bin" blabla blibli)

whereas with my code we end up with

    ("PATH=/the/opam/bin" blabla "PATH=hello" blibli)

[where the `cdr` is the original list.]
But if you try to (call-process "printenv" nil t) you'll see that it
results in the exact same environment.


        Stefan

erikmd:
Dear @monnier,
Yes, this is exactly what I had in mind regarding the form of the list.
> But if you try to (call-process "printenv" nil t) you'll see that ...

OK! this is the info I was missing.

So, I'll revert the revert by a git-rebase just now.
Thanks again for your feedback ^^

monnier:
>> But if you try to (call-process "printenv" nil t) you'll see that ...
> OK! this is the info I was missing.

That's what I meant by "Don't reimplement by hand the semantics already
provided by `process-environment`."

erikmd:
Thanks Stefan! (and then, sorry for the noise w.r.t. the revert of the revert :'-)

erikmd:
Hi all,
I believe all questions and suggestions were addressed.
@voodoos, does the current version of the PR look good to you?

emacs/merlin.el
```diff
@@ -1921,7 +1941,7 @@ Empty string defaults to jumping to all these."
          (with-temp-buffer
            (if (eq (call-process-shell-command
```
voodoos, line 1942: Now both branches of this `if` return the same value: `"ocamlmerlin"` while before the opam bin dir was prepended to it in one case. 

I guess it was useless to return the absolute path to the executable since the `$PATH` variable is replaced by the path to the opam bin folder. Is that correct ? If it is, then it's a good thing to get rid of it, but we should switch that `if` to only perform the side effect of modifying the path, returning the same value in both branches is misleading.

emacs/merlin.el
```diff
@@ -611,6 +601,35 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
         (kill-local-variable 'merlin-buffer-configuration)
         (kill-local-variable 'merlin-erroneous-buffer)))))
 
+(defcustom merlin-stop-server-on-opam-switch t
+  "If t, stop the Merlin server before the opam switch changes.
```
voodoos, line 605:
```suggestion
  "If t, stops the Merlin server before the opam switch changes.
```

voodoos (commented):
The changes looks good to me, especially the simplification of the `env` replacement. I am not sure why it was made so complex before.

> it seems the last CI run was partial though (github says "5 workflows awaiting approval")
Sadly, we do not really test the editor modes in the CI right now. @3Rafal do you think you could try these changes on your setup ? The main thing to look for is that the correct Merlin is started depending on the currently selected opam switch.

emacs/merlin.el
```diff
@@ -512,7 +510,9 @@ argument (lookup appropriate binary, setup logging, pass global settings)"
   ;; Really start process
   (let ((binary      (merlin-command))
         ;; (flags       (merlin-lookup 'flags merlin-buffer-configuration))
-        (process-environment (cl-copy-list process-environment))
+        (process-environment
+         (append (merlin-lookup 'env merlin-buffer-configuration)
+                 process-environment))
```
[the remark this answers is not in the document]
erikmd: Good point! I will push a commit to fix this.

emacs/merlin.el
erikmd: Updated.

emacs/merlin.el
erikmd: I guess it was useless to return the absolute path to the executable since the `$PATH` variable is replaced by the path to the opam bin folder. Is that correct ?

Yes!

> If it is, then it's a good thing to get rid of it, but we should switch that `if` to only perform the side effect of modifying the path, returning the same value in both branches is misleading.

Agree; I just updated the function accordingly; thanks!

voodoos: Thanks for the changes! I think we can go one step further and make the value much closer to the binding name:
```
unless command
  setq command "ocamlmerlin"
  with-temp-buffer-query-opam
```

erikmd: Hi @voodoos!

I don't think so… because there are other cases after the current `(setq command` :

```
       (cond
        ((functionp merlin-command) (funcall merlin-command))
        ((stringp merlin-command) merlin-command)
```

3Rafal:
@voodoos , I've played with it on my machine and it works:)

voodoos (approved):
Thanks for your contribution @erikmd !
Could you rebase and squash the commits when it makes sense before the merge ?

erikmd:
@voodoos done:
I rebased the PR to only keep 2 commits (and I made sure that the diff was empty)

monnier:
> I guess it was useless to return the absolute path to the executable
> since the `$PATH` variable is replaced by the path to the opam bin
> folder.  Is that correct ?

Exactly, yes.

> +(defcustom merlin-stop-server-on-opam-switch t
> +  "If t, stop the Merlin server before the opam switch changes.
>
> ```suggestion
>   "If t, stops the Merlin server before the opam switch changes.
> ```

FWIW, a variable can't stop the server, only code can do that, which
is why in Emacs we often use the imperative to document the effects of
a variable :-)


        Stefan

erikmd:
Thanks for the merge! 😊

PR #1655, opened 2023-07-20
Add support for parsing negative numbers in sexps

ddickstein:
The motivation for this came from failing to parse a response that Merlin returned due to the `-1` values for `col` in the `start` and `end` fields:

```
((assoc)
 (class . "return")
 (value (
   (assoc)
   (start (assoc) (line . 0) (col . -1))
   (end (assoc) (line . 0) (col . -1))
   (type . "typer")
   (sub)
   (valid . true)
   (message . "omitted")))
 (notifications)
 (timing (assoc) (clock . 166) (cpu . 145) (query . 1) (pp . 0) (reader . 21) (ppx . 36) (typer . 76) (error . 11)))
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ unreleased
       fixes #1647)
     - Prevent `short-path` from looping in some cases related to recursive type
       definitions (#1645)
+    - Support parsing negative numbers in sexps (#1655)
   + editor modes
     - emacs: call merlin-client-logger with "interrupted" if the
       merlin binary itself is interrupted, not just the parsing of the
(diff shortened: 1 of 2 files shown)


voodoos:
Hi @ddickstein, the patch looks okay, but I'm curious to know how the issue is triggered exactly. 
The output you show is an `error` query result which contains dummy positions (as defined in `Lexing.dummy_pos`, for reference). What is you use-case for re-feeding part of this output to Merlin ?

Also, this change will require a changelog entry :-)

ddickstein:
The use case is that we pass the returned data from Merlin to a process that extracts and records telemetry information, but Merlin serializes sexps differently from sexplib (maps, in particular), so we need to use Merlin's logic to deserialize it.

voodoos:
Ok, thanks for the bit of context. We could consider moving to an established sexp library at some point...

Issue #1601, opened 2023-05-11
case-analysis crashes with closed variant type


ddickstein:
```ocaml
let foo : [< `Foo ] option = None

let () =
  match foo with
  | None | Some _ -> ()
;;
```
Using the Merlin from 9fc6355db49fbb8b59e633e5d149935295d2df25, destructing on the _ gives:
```
File "src/analysis/destruct.ml", line 558, characters 20-26: Assertion failed
Raised at Merlin_analysis__Destruct.node in file "src/analysis/destruct.ml", line 558, characters 20-32
Called from Merlin_utils__Misc.try_finally in file "src/utils/misc.ml", line 45, characters 8-15
Re-raised at Merlin_utils__Misc.try_finally in file "src/utils/misc.ml", line 62, characters 10-24
Called from Merlin_utils__Misc.protect_refs.(fun) in file "src/utils/misc.ml", line 82, characters 10-14
Re-raised at Merlin_utils__Misc.protect_refs.(fun) in file "src/utils/misc.ml", line 84, characters 38-45
Called from Ocaml_typing__Persistent_env.without_cmis in file "src/ocaml/typing/persistent_env.ml", line 151, characters 10-109
Called from Merlin_utils__Std.let_ref in file "src/utils/std.ml", line 690, characters 8-12
Re-raised at Merlin_utils__Std.let_ref in file "src/utils/std.ml", line 692, characters 30-39
Called from Dune__exe__New_commands.run in file "src/frontend/ocamlmerlin/new/new_commands.ml", line 65, characters 15-53
Called from Merlin_utils__Std.let_ref in file "src/utils/std.ml", line 690, characters 8-12
Re-raised at Merlin_utils__Std.let_ref in file "src/utils/std.ml", line 692, characters 30-39
Called from Merlin_utils__Misc.try_finally in file "src/utils/misc.ml", line 45, characters 8-15
Re-raised at Merlin_utils__Misc.try_finally in file "src/utils/misc.ml", line 62, characters 10-24
Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
Called from Merlin_kernel__Mocaml.with_state in file "src/kernel/mocaml.ml", line 18, characters 8-38
Re-raised at Merlin_kernel__Mocaml.with_state in file "src/kernel/mocaml.ml", line 20, characters 42-53
[...]

PR #1657, opened 2023-07-25
[RFC] Fuzzy behavior tests/CI for Merlin

### LINKED ISSUE #1634: General Behavior CI

### LINKED ISSUE #1635: Specified Regression CIs
The General Behavior CI (see https://github.com/ocaml/merlin/issues/1634) needs a bit of thinking to scale: It compares on the whole Merlin output and the Merlin output of some samples is very very big. As a lower-hanging fruit, we can first implement a concrete Behavior Regression CI.

## Purpose

Some concrete changes in behavior can almost certainly be considered a regression when happening. Two examples:
- Merlin now errors where before it would succeed (`error-regression CI`).
- The Merlin server now crashes where before it wouldn't (`crash-regression CI`).

So additionally to the General Behavior CI https://github.com/ocaml/merlin/issues/1634, we can also add concrete Regression CI workflows monitoring these concrete changes. Adding this level of output simplicity to the CI will make it easier to scale in terms of size and help avoid missing concrete regressions in behavior.

## Implementation

Let's focus on the `error-regression CI` for now.

`merl-an` already has a command `merl-an error-regression` for the `error-regression CI`. What's missing is the integration into a CI workflow.

### Data to take into account

For the first and most direct PoC, @3Rafal has already gathered data:

- Running 7 different queries, each on 1 sample per file on the whole of Irmin (700-800 files):
 In [CI](https://github.com/ocaml/merlin/actions/runs/5413597474/jobs/9839392402?pr=1642):
  ```
  38m 12s
  ```

- Running 7 different queries, each on 30 samples per file on the whole of Irmin (700-800 files):
  Locally on a Laptop with Intel i7 11 gen:
  ```
  203m8.524s
  ```

This is the data for the first and most direct PoC. We can optimize the CI in terms of time on different levels (see below).

### Next action points

@3Rafal has already written a [PoC for the `error-regression CI`](https://github.com/ocaml/merlin/pull/1642). There are several things we'd need to improve to make the CI more useful. The next action points are

- [ ] Discuss and decide if these CIs are really useful and worth the effort.
  - [ ] For the `error-regression CI`: Find out if there have been any changes in the past that made Merlin return an error where before it would return suffesscully. If so, which ones? And how likely is it that this will happen again in the future?
  - [ ] Similarly for the `crash-regression CI`: Have there been any changes in the past which made Merlin crash where before it wouldn't? If so, which ones?
- [ ] If we think it might be worth it, optimize the CI in terms of time:
    - [ ] On the CI-side: e.g. cache the set-up.
    - [ ] On the `merl-an`-side: Make sure we use the Merlin cache as much as possible (e.g. if the traversal is done in query -> file order, do it in file -> query order instead).
    - [ ] On the sample-side:
      - [ ] For Merlin cache, better to have lots of samples in few files than having few samples in a lot of files.
      - [ ] Do we want to run it on all `mli`-files as well?
- [ ] Have a look at how long the optimized CI takes and decide a good sample set and workflow for it.

pitag-ha:
Merlin already has quite a few end-to-end tests. All of those are behavior-specific: we want to test one concrete Merlin behavior, create specific input for that and test Merlin on that input. Additionally to those tests, we now also want to test Merlin's behavior non-specifically on a large randomized input scale.

This PR builds on top of PR #1642. The main differences are:
- It's considerably faster: About 10-15 min for 23482 samples, as opposed to 30 min for 3113 samples (or >200 min for 93390 samples).
- It gives more meaningul output (see  _Two in One_ section below):
    - The category `return` data is finer-grained.
    - It also adds category data for server crashes.
    - It also outputs the full Merlin responses.
- The test code base and the sample size are chosen in a way to get a good maintenance-impact payoff (see the _Test cost maintenance data_ and _Impact_ sections below)
- CI integration is left out for now for the RFC: Given that it also cats the full Merlin responses, the CI integration requires filtering out switch-specific noise etc.

This implements both issues #1634 and #1635. Here, I'm doing both things in one workflow, which I think is the right thing to do. Closes #1634, closes #1635.

## Approach

I'd like to discuss the approach. So, what do you think about the following?

### Local tests

Similarly to PR #1642, I'm also proposing to have the tests locally as well, not only in CI. Same as there: they would _not_ run every time on `dune test`.

### Test cost and maintenance data

I've optimized the tool `merl-an` we use to generate the samples etc and I've chosen the sample base carefully. With that, the time seems reasonable to me for the CI, while the space still seems quite big to me to add it to the repo:

#### Time
If the test base is already fetched and built (that should be the case most of the time now thanks to Dune caching):
```
real    11m41,585s
user    2m18,031s
sys     0m20,412s
```

If the base needs to be fetched and built:
```
real    14m38,179s
user    2m52,283s
sys     0m27,659s
```

Both timings come from my local computer.

#### Space

There are two big test files that would get into the Merlin repo with this approach: `category_data.t` and `full_responses.t`. They could be made more compact, but at a pricee: see _Handle the JSON output_ below. We could also make them smaller by choosing less samples. Currently their size is:

`category_data.t`:
4,7M
189460 lines

`full_responses.t`:
36M
1168705 lines

If we do go forward with this, we should probably use Git LFS. Does anyone have experience with that?

#### CI
For 23482 samples as in this RFC, I expect the CI to take about 15-20 min (side-note: I could push the github yml from the #1642 PR, even though currently the CI would fail, to make sure this is true).

There are other CIs on Merlin that also seem to take that long. Is that true? If so, we could try running the CI on every PR (fulfilling a few things such as ml/mli-files were modified etc) and see if that ends up being annoying or ok. We can also optimize the sample size further.

Alternatively, I have a potential alternative workflow in mind, if you think it's too long to run on every PR.

#### Locally
Locally, `dune test` won't run these tests. People can run them via `FUZZING=true dune test`, which they should do iff expecting or fearing a behavior change. That's one of the weak points of this approach, since that will take about 10 min locally - or even 15 min when the sample base isn't cached.

### Two in one

This test generates two different data sets: category data in `category_data.t`, i.e. the implementation of issue https://github.com/ocaml/merlin/issues/1635, and full Merlin responses in `full_responses.t`, i.e. the implementation of issue https://github.com/ocaml/merlin/issues/1634.

#### Category data

The Merlin responses can face regressions/improvements on different levels. When treating such big amount of test data as here, it's helpful to reduce the data and seperate the different levels.

##### Return classes

One level a regression can happen on is the the return class. That's why in `category_data.t`, the responses are reduced to "belongs to one of these six return categories", e.g. category "successful return", or category "properly reported error" etc.

##### Server crashes

Another interesting regression is server crashes. That's why `category_data.t` also contains the server's query number.

#### Full responses

By reducing the response, regressions/behavior changes can also be missed, though. That's why the test also contains the full responses in `full_responses.t`.

#### Concrete regressions we would have caught in the past

I've checked how analyzing this test would look like by running it on a couple of concrete `locate` regressions that have happened in the past. Both regressions would have been caught by the new test.

For the first one, the test diff is [here](https://github.com/ocaml/merlin/compare/5e37babd220e4063da0a83a15716fcce187b946e...098096b29be10ecbd890cc8dbb7a992ad8e7a9b4). It's short and easy to read. The full responses data would have caught the regression, the category data wouldn't. The regression was about [`locate` in the context of functor applications](https://github.com/ocaml/merlin/pull/1329).

For the second one, the test diff is [here](https://github.com/ocaml/merlin/compare/696be9648b04588cd555f08f206b61a98096417f...3bd0bdd94ba101df48543709c0ab478339e9562c). It's also short and easy to read. Both the full responses data and the category data would have caught the regression. The regression was about [`locate` in the context of modules without implementation](https://github.com/ocaml/merlin/issues/1519).

#### Behavior changes on big PRs such as compiler bumps

I've also had a look how the test diffs would look like on big PRs such as compiler bumps.

The diff for the 4.13 bump is [here](https://github.com/ocaml/merlin/compare/631c958d3779eee553877b211dd2845d1f902686...d487a50653649416d76ecb438f8f8c13c987c91d). Unfortunately, it contains a bit of unnecessary noise (we need to fix that). The category data is easy to read and it directly shows a regression in `occurrences `: It sometimes returns an empty list now, where before it would find occurrences. The full responses are a bit harder to read, but also contain interesting changes.

The diff for the 4.14 bump is [here](https://github.com/ocaml/merlin/compare/697245e17b09416227ba3e7a5616bac93d43785e...f89dd850ac9dcde007e457007ed77c9966fca17e). It's similar.

#### General

I also had to update the test between the mentioned PRs. Interestingly, all of them contain regressions as well (while also containing quite a few improvements of course). E.g.:

One interesting diff is [this one](https://github.com/ocaml/merlin/compare/f4eba5df1142239968f2d8ae65ec7b7b1ee1c5ab...2a2b885fb27e42e5943d0f1eaa10b3db04009018). It contains one of the `locate` regressions whose fix is mentioned above. Additionally, it contains a few more `locate` regressions. One of them: The return is now an exception, where before it was a value. Those regressions were introduced at some point between the 4.14 bump and [this PR](https://github.com/ocaml/merlin/pull/1522).

#### Test code base

The test code base is buildable, in particular it doesn't have errors and doesn't contain holes. So no regression in those contexts can be found.

To give one example: The regression in the context of holes and PPXs wouldn't have been caught. I didn't even bother to try. We could "overfit" this test to that regression by adding a workflow to `merl-an` producing that situation. I don't think that's worth it. For those cases, the normal input-specific end-to-end tests are more suitable.

#### Sampe size

The samples can't cover all edge cases, so we'll always miss things.

To give one example: Originally, I included `case-analysis` in the queries that are run. I've checked if [a regression making the server crash on certain `case-analysis` queries](https://github.com/ocaml/merlin/issues/1601) would have been caught. It wouldn't have.

## Implementation details and their consequences

This is a quick PoC with the goal to gather some data (see _Impact_) and to discuss this approach/workflow I'm suggesting. There are several implementation details we'll need to approach before thinking about merging this. E.g.:

#### Where does `merl-an` live
The current workflow relies on having `merl-an` installed. Instead, I suggest upstreaming it to Merlin and having Dune build it as part of the test.

#### Where do the Irmin dependencies come from
The current workflow relies on having the Irmin dependencies installed. If we wanted that workflow, we'd have to add all Irmin dependencies as Merlin test dependencies. Instead, I suggest vendoring Irmin and its dependencies. Note:

#### CI integration

The CI integration can be taken from PR #1642. We just need to make sure the tests are unspecific to locally vs CI.

#### Filter out noise
Related: By not using `merlin-wrapper`, the tests aren't filtering out the noise coming from e.g. switch-specific and compiler-specific things.

Furthermore, the current [query_num] adds a lot of noise as well: As soon as there was one server crash, all the following queries will be marked in the diff as well.

#### Handle the JSON output
Currently, the tests pipe the JSON to `jq`. That makes both the JSON's themselves and their diffs very readable. However, it makes the files extremely long. We need to find a good balance between strongly formatted JSON (piping to `jq`) and very compact JSON (`merl-an`'s output directly).

#### Sample set choise
The current sample choice is already quite conscious and adjusted and I don't think it's bad (I can give detail if that's interesting). However, there are still 3 ways in which we can improve it: Better sample distribution: https://github.com/pitag-ha/merl-an/issues/28; optimize the test code base even more (can it be smaller catching the same? Bigger catching more?). Optimize the amount of samples per file (can it be less catching the same? Bigger cathing more?).

#### Queries

E.g.: Do we want to include `case-analysis`?

#### Minor improvements/details
There are several details we could improve if we want to, e.g:

The `return` category data could be even finer grained. E.g. when a list is returned, it's interesting to know its size. For example, that's interesting for `complete-prefix`, where changes often consist in having more or less suggestions than before.

Also, we could use dune diff workflow directly instead of cram tests. The cram tests just cat files, t


pitag-ha:
I'm closing this as it has long been superseded by https://github.com/ocaml/merlin/pull/1716 .

PR #1668, opened 2023-08-31
fix(merlin.el): Fix and improve emacs bindings

### LINKED ISSUE #1386: Emacs bindings
After [discussing](https://github.com/ocaml/tuareg/issues/273) with @monnier, @mattiase, and @erikmd about Tuareg _documentation_ bindings, we would like to expand this discussion with the Merlin developers.

- There is unfortunately no standard Emacs' documentation keybinding but we are thinking to use <kbd>C-c C-d</kbd> (we cannot use [<kbd>C-c C-h</kbd>](https://github.com/ocaml/merlin/issues/860) as <kbd>C-h</kbd> is [reserved](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) for discoverability).  This however conflicts with the key for `m︭erlin-d︭estruct`.  May we suggest that <kbd>C-c M-d</kbd> be used for this (which is even more mnemonic 😀).
- Merlin shadows <kbd>C-c C-r</kbd> (bound in Tuareg and Caml-mode to `...-eval-region`).  What do you think of binding `merlin-error-check` to  <kbd>C-c M-r</kbd>?
- Given that <kbd>M-.</kbd> (resp. <kbd>M-,</kbd>) are standard in Emacs (and already supported by Merlin), one can remove <kbd>C-c C-l</kbd> (resp. <kbd>C-c &</kbd>) — I'm thinking to do the same in Tuareg after making `caml-types-show-ident` available through `xref`.
- How about binding `merlin-occurrences` to [<kbd>M-?</kbd> (through `xref`)](https://www.gnu.org/software/emacs/manual/html_node/emacs/Identifier-Search.html)?
- Contrarily to [what is asserted on the wiki](https://github.com/ocaml/merlin/wiki/emacs-from-scratch), <kbd>C-c t</kbd> is no longer bound to `merlin-type-expr` (a [good thing as it is reserved to the user](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html)).  May we suggest to bind it to <kbd>C-c C-t</kbd> with a prefix?

erikmd:
(following the discussion in https://github.com/ocaml/merlin/issues/1386)

Cc @monnier @voodoos FYI

For a quick recap of the changes and their motivation, see also the table from comment:

https://github.com/ocaml/merlin/issues/1386#issuecomment-1701567716

Close #1386

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,8 +5,14 @@ unreleased
   + merlin binary
     - Improve error messages for missing configuration reader (#1669)
     - Fix regression causing crash when using ppxes under Windows (#1673)
-
-
+  + editor modes
+    - emacs: fix/improve keybindings (#1668, fixes #1386):
+      Unbind <kbd>C-c C-r</kbd> (to avoid shadowing `tuareg-eval-region`)
+      and bind <kbd>C-c C-v</kbd> instead to `merlin-error-check`;
+      rebind <kbd>C-c C-d</kbd> to `merlin-document`
+      and bind <kbd>C-c M-d</kbd> and <kbd>C-c |</kbd> instead to `merlin-destruct`;
+      bind <kbd>C-u C-c C-t</kbd> to `merlin-type-expr`.
+      See also <https://github.com/ocaml/merlin/issues/1386#issuecomment-1701567716>
 
 merlin 4.10
 ==========
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -1316,15 +1316,18 @@ If QUIET is non nil, then an overlay and the merlin types can be used."
       (set-temporary-overlay-map merlin-type-enclosing-map t
                                  'merlin--type-enclosing-reset))))
 
-(defun merlin-type-enclosing ()
+(defun merlin-type-enclosing (&optional manual)
   "Print the type of the expression under point (or of the region, if it exists).
-If called repeatedly, increase the verbosity of the type shown."
-  (interactive)
-  (if (region-active-p)
-      (merlin--type-region)
-    (when (merlin--type-enclosing-query)
-      (merlin-type-enclosing-go-up)
-      (merlin--type-enclosing-after))))
+If called repeatedly, increase the verbosity of the type shown.
+With prefix argument MANUAL, call `merlin-type-expr' interactively."
+  (interactive "P")
+  (if manual
+      (call-interactively #'merlin-type-expr)
+    (if (region-active-p)
+        (merlin--type-region)
+      (when (merlin--type-enclosing-query)
+        (merlin-type-enclosing-go-up)
+        (merlin--type-enclosing-after)))))
 
 (defun merlin--find-extents (list low high)
   "Return the smallest extent in LIST that LOW and HIGH fit
@@ -1979,9 +1982,11 @@ Empty string defaults to jumping to all these."
     (define-key merlin-map (kbd "C-c C-x") #'merlin-error-next)
     (define-key merlin-map (kbd "C-c C-l") #'merlin-locate)
     (define-key merlin-map (kbd "C-c &"  ) #'merlin-pop-stack)
-    (define-key merlin-map (kbd "C-c C-r") #'merlin-error-check)
+    (define-key merlin-map (kbd "C-c C-v") #'merlin-error-check)
     (define-key merlin-map (kbd "C-c C-t") #'merlin-type-enclosing)
-    (define-key merlin-map (kbd "C-c C-d") #'merlin-destruct)
+    (define-key merlin-map (kbd "C-c C-d") #'merlin-document)
+    (define-key merlin-map (kbd "C-c M-d") #'merlin-destruct)
+    (define-key merlin-map (kbd "C-c |") #'merlin-destruct)
     (define-key merlin-map (kbd "C-c C-n") #'merlin-phrase-next)
     (define-key merlin-map (kbd "C-c C-p") #'merlin-phrase-prev)
     (define-key merlin-menu-map [customize]
--- a/.github/workflows/emacs-lint.yml
+++ b/.github/workflows/emacs-lint.yml
@@ -18,7 +18,9 @@ jobs:
         emacs_version:
           - '27.2'
           - '28.2'
+          - '29.1'
           - snapshot
+      fail-fast: false  # don't stop jobs if one fails
     env:
       EMACS_PACKAGE_LINT_IGNORE: ${{ matrix.package_lint_ignore }}
       EMACS_BYTECOMP_WARN_IGNORE: ${{ matrix.bytecomp_warn_ignore }}


erikmd:
Hi; does this wrap-up PR looks good to you?

BTW, sorry for not opening it earlier… I realized it could have been part of the latest release;
anyway, I believe it wouldn't be required to do yet another small release just for this PR for the time being,
given that most emacs users use merlin from MELPA (i.e., from the latest master commit), not MELPA stable.

voodoos:
Thanks for the thorough work on the summary !

The proposed changes looks reasonable to me, however I am not an emacs user myself and bindings involve some very old muscle memory for users. I will take some the time to gather more feedback before reviewing.

We might also think of a way to teach users these new bindings, or maybe have a transition period ?

erikmd:
Hi @voodoos, thanks for your feedback!

> and bindings involve some very old muscle memory for users.

I agree! but regarding the proposed changes, concretely only 2 existing bindings are changed:

* `C-c C-r` → it looks sensible to change it because as pointed out by @Chris00, this minor-mode binding was normally a major-mode (Tuareg) binding, unexpectedly hidden by merlin (and AFAICT, I was not using that binding that often because merlin's typechecking is already triggered upon save `C-c C-s`)

* `C-c C-d` → this rebind may indeed cause more "surprise" at first sight because I expect it is more used than `C-c C-r`, but I was not the only person to suggest binding that one to a documentation-related binding (on this point, we already got @mattiase, @bbatsov and @Chris00's advice)
 
> I will take some the time to gather more feedback before reviewing.

Sure! maybe we could post this question on https://discuss.ocaml.org/ ? (if ever if you think that's needed/useful and you'd prefer that I post the question myself, let me know)

> We might also think of a way to teach users these new bindings, or maybe have a transition period ?

I'm not aware of standard and simple ways to do new-bindings-advertisement or transition-period in Emacs, maybe @monnier could tell.

Actually what I generally do is to suggest installing modes such as https://github.com/justbur/emacs-which-key (there are similar modes that do almost the same) which, after typing a prefix such as `C-c`, display a popup with the main bindings:

2023-09-06_19-37-49_Screenshot_which-key

monnier:
> I'm not aware of standard and simple ways to do new-bindings-advertisement
> or transition-period in Emacs, maybe @monnier could tell.

Emacs does not have mechanisms for that, indeed.
Usually I'd expect to solve this problem by making sure the new binding
of the key is not dangerous and that its behavior is sufficiently
self-explanatory that the user will understand what's going.

You could also add a message in the command stating something like
"blabla has moved to `<newkey>`".  This will need some way to make sure
the message is not emitted too often to become annoying.


        Stefan

erikmd:
> You could also add a message in the command

1. Yes, why not: for `C-c C-d`, we could show a message in the mini-buffer such as `The command "merlin-destruct" can now be run with C-c |`,

   similary to: https://github.com/emacsmirror/emacs/blob/8255b9c1f10a0789c724dbdbe2c47eaa633f6697/lisp/simple.el#L2559

> This will need some way to make sure the message is not emitted too often to become annoying.

3. Yes, maybe we could stop printing the message after `C-c |` has been run twice (e.g., thanks to a defcustom that will count that number). WDYT?

4. But question: note that with the current version of the PR, the function previously triggered by `C-c C-d` is not moved to a single binding, but to 2 bindings, both TTY-compliant:

   ```diff
    (define-key merlin-map (kbd "C-c M-d") #'merlin-destruct)
    (define-key merlin-map (kbd "C-c |") #'merlin-destruct)
   ```

   Do you think we should only select one binding? (say, the simplest one to type, `C-c | ?`)

erikmd:
On the other hand, I'm wary of adding too much business logic just for this `C-c C-d` keybinding change… I guess we should avoid any "over-engineering" 😅

monnier:
> On the other hand, I'm wary of adding too much business logic just for this
> `C-c C-d` keybinding change… I guess we should avoid any "over-engineering"
> 😅

Exactly.  Maybe this deserves a feature request on Emacs side, since
it's something we often need/want: some standardized way to emit
a message that should not be repeated ad-nauseam.

Doing that specifically for this binding falls very quickly into
"over-engineering", but if you do it once and forall in Emacs itself and
it can be reused in many places, the tradeoff changes significantly.

erikmd:
OK, thanks @monnier for your comment!

BTW, some advice about my question 3. above?

> … with the current version of the PR, the function previously triggered by C-c C-d is not moved to a single binding, but to 2 bindings, both TTY-compliant:
> 
>     (define-key merlin-map (kbd "C-c M-d") #'merlin-destruct)
>     (define-key merlin-map (kbd "C-c |") #'merlin-destruct)
> 
> Do you think we should only select one binding? (say, the simplest one to type, C-c | ?)

monnier:
> BTW, some advice about my question 3. above?
>
>> … with the current version of the PR, the function previously triggered by
>> C-c C-d is not moved to a single binding, but to 2 bindings, both
>> TTY-compliant:
>> 
>>     (define-key merlin-map (kbd "C-c M-d") #'merlin-destruct)
>>     (define-key merlin-map (kbd "C-c |") #'merlin-destruct)
>> 
>> Do you think we should only select one binding? (say, the simplest one to type, C-c | ?)

No opinion on this.  It's not uncommon to provide several bindings for
the same command.  I can't think of any convention that would object to
either (or both) of those two bindings.

bbatsov:
> It's not uncommon to provide several bindings for
the same command. 

Indeed. I do it all the time in several packages I maintain.

bbatsov:
While I agree that changes to keybindings can be painful, we also have to consider that:

- the `merlin-destruct` command doesn't seem like something very widely used, so probably the impact of the change won't be big
- it's easy for the end users to change some keybindings if they are unhappy with them

My point is that we're probably overthinking this.

emacs/merlin.el
```diff
@@ -1316,15 +1316,18 @@ If QUIET is non nil, then an overlay and the merlin types can be used."
...
+If called repeatedly, increase the verbosity of the type shown.
+With prefix argument MANUAL, call `merlin-type-expr' interactively."
+  (interactive "P")
+  (if manual
+      (call-interactively #'merlin-type-expr)
+    (if (region-active-p)
+        (merlin--type-region)
+      (when (merlin--type-enclosing-query)
+        (merlin-type-enclosing-go-up)
+        (merlin--type-enclosing-after)))))
```
voodoos, line 1330: These changes looks out-of-place in this PR.
Additionally I don't see the benefit of thi new argument instead of calling directly `merlin-type-expr` ?

emacs/merlin.el
```diff
@@ -1979,9 +1982,11 @@ Empty string defaults to jumping to all these."
     (define-key merlin-map (kbd "C-c C-x") #'merlin-error-next)
     (define-key merlin-map (kbd "C-c C-l") #'merlin-locate)
     (define-key merlin-map (kbd "C-c &"  ) #'merlin-pop-stack)
-    (define-key merlin-map (kbd "C-c C-r") #'merlin-error-check)
+    (define-key merlin-map (kbd "C-c C-v") #'merlin-error-check)
     (define-key merlin-map (kbd "C-c C-t") #'merlin-type-enclosing)
-    (define-key merlin-map (kbd "C-c C-d") #'merlin-destruct)
+    (define-key merlin-map (kbd "C-c C-d") #'merlin-document)
+    (define-key merlin-map (kbd "C-c M-d") #'merlin-destruct)
+    (define-key merlin-map (kbd "C-c |") #'merlin-destruct)
```
voodoos, line 1989: How does that work on macos ? The `|` character is a bit hard to type already (`alt+maj+l`)

voodoos (commented):
Thanks for the changes ! Let's try these new bindings...
Could you also update the documentation website on the `gh-pages` branch ?

emacs/merlin.el
erikmd: @voodoos I don't think so, this follows the [initial suggestion by @Chris00](https://github.com/ocaml/merlin/issues/1386#issue-993828239):

> Contrarily to [what is asserted on the wiki](https://github.com/ocaml/merlin/wiki/emacs-from-scratch), `C-c t` is no longer bound to `merlin-type-expr` (a [good thing as it is reserved to the user](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html)). May we suggest to bind it to `C-c C-t` with a prefix?

To be more precise, @Chris00 meant:

* *Keeping the `C-c C-t` binding as is for `merlin-type-enclosing`* (so, no backward-compatibility issue),
* and *bind `C-u C-c C-t` to `merlin-type-expr`*.

Given `C-u` is the universal prefix, TTBOMK this latter binding should be implemented the way I did: accordingly it's a bit verbose in the diff, but I think this code is idiomatic (otherwise @monnier would have told otherwise I guess :)

emacs/merlin.el
erikmd: Indeed: this `|`-based binding (especially chosen to suggest the pattern-matching construct) is easier to type on QWERTY and AZERTY keyboards! I guess macOS users would rather use the other keybinding for `merlin-destruct`…

So maybe, this could be another argument to keep these two bindings for `merlin-destruct`?
(see also [this previous comment](https://github.com/ocaml/merlin/pull/1668#issuecomment-1709132411) on that point)

Anyway, feel free to let us know if you're also OK with this dual-binding idea… or if you'd prefer that I remove `C-c |` to only keep `C-c M-d`.
Both choices would be fine with me.

erikmd:
Hi @voodoos,
> Could you also update the documentation website on the gh-pages branch ?

To sum up, do you suggest that I open a PR that targets this file?:
https://github.com/ocaml/merlin/blob/gh-pages/editor/emacs/index.markdown

emacs/merlin.el
voodoos: The dual-binding seems fine.

CHANGES.md
```diff
@@ -5,8 +5,9 @@ unreleased
   + merlin binary
     - Improve error messages for missing configuration reader (#1669)
     - Fix regression causing crash when using ppxes under Windows (#1673)
-
-
+  + editor modes
+    - emacs: fix/improve keybindings (#1668, fixes #1386).
+      See <https://github.com/ocaml/merlin/issues/1386#issuecomment-1701567716>
```
voodoos, line 10: One last thing: we should probably list the binding changes here in addition to the link.

voodoos:
> To sum up, do you suggest that I open a PR that edits this file? https://github.com/ocaml/merlin/blob/gh-pages/editor/emacs/index.markdown

Exactly yes :-) 

Thanks for all the effort on the PR :+1:, I will merge it when the changelog is updated.

PR #1688, opened 2023-09-27
add test for dynamic python3 installation

nilsbecker:
MacVim comes with options +python3/dyn and -python3. This means currently merlin fails to load. This patch fixes that by testing for the dynamic python3.

I did not include the same for python/dyn because by now I think it's better not to support python 2.


voodoos (approved):
Thanks !

PR #1690, opened 2023-10-05
Break ties between ghost nodes by looking at cursor position

### LINKED ISSUE #1689: #1664 may select a worse ghost node if both nodes are ghost
#1664 (which fixes #1660) attempts to prioritize non-ghost nodes. It does so with this code:

https://github.com/ocaml/merlin/pull/1664/files#diff-bcb7135781bd818b310e2c37a0018ebd78f5f857cb7e8989fb3b4000a2be12ddR126-R129

```ocaml
    if l1.Location.loc_ghost then 1
    else if l2.Location.loc_ghost then -1
    else (* compare nodes based on location *)
```

However, I think this ends up always picking `l2` if `l1` is ghost, even if `l2` is *also* ghost and a worse ghost location than `l1`. I'm seeing this in a large test case, but it's difficult (and I argue not worth it) to minimize.

I think this should look something more like a lexicographic compare, where ghostiness is considered before location. I plan on opening a PR in a moment.

ncik-roberts:
Fix #1689 by choosing the best ghost node when both are ghost.

It would take me many times longer to write a test than it took to write this PR. If anyone has ideas for a simple test, I'm open to it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 merlin NEXT_VERSION
 ===================
 
+  + merlin binary
+    - Fix a follow-up issue to the preference of non-ghost nodes introduced in #1660 (#1690, fixes #1689)
   + editor modes
     - vim: load merlin when Vim is compiled with +python3/dyn (e.g. MacVim)
 
--- a/src/kernel/mbrowse.ml
+++ b/src/kernel/mbrowse.ml
@@ -120,11 +120,13 @@ let compare_locations pos l1 l2 =
     Location_aux.compare_pos pos l2
   with
   | 0, 0 ->
-    (* Cursor inside both locations: favor closer to the end *)
-    if l1.Location.loc_ghost then 1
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -3,6 +3,8 @@ merlin NEXT_VERSION
 
   + editor modes
     - vim: load merlin when Vim is compiled with +python3/dyn (e.g. MacVim)
+  + merlin binary
+    - Fix a follow-up issue to the preference of non-ghost nodes introduced in #1660 (#1690, fixes #1689)
```
voodoos, line 7: We usually put the merlin binary section first.

voodoos (approved):
Looks good to me, thanks for your contribution !

(I cannot think of a simple way to make a test appart from writing some custom ppx...)

CHANGES.md
ncik-roberts: Fixed

Issue #1663, opened 2023-08-23
emacs' merlin-mode highlighting of code w/ errors is just bad


UnixJunkie:
Putting a lot of code in a bold pinkish font is just not helpful.
That makes the whole code completely unreadable.
I have to disable merlin-mode so that I can continue
working on the code in question...

https://discuss.ocaml.org/t/emacs-merlin-mode-and-highlighting-of-code-w-errors/12845/2

dbuenzli:
Add `(setq merlin-error-on-single-line t)` to your `.emacs`. Personally I think this should be the default.

UnixJunkie:
+1 for having nice defaults...

3Rafal:
@UnixJunkie ,
I made a PR that changes the defaults. Can you give me some nice examples of the behavior so that I can test it?

UnixJunkie:
Just take a significant ocaml source file from any project, and change it so that it does not compile anymore.

UnixJunkie:
And yes, this PR https://github.com/ocaml/merlin/pull/1693 changes the default to something I prefer.
Now, the question is, is this what's preferred by most people or not and hence will the PR be merged...

nojb:
> Can you give me some nice examples of the behavior so that I can test it?

I exaplained a typical example that exhibits the issue in https://github.com/ocaml/merlin/pull/1693#issuecomment-1774635100

voodoos:
> Now, the question is, is this what's preferred by most people or not and hence will the PR be merged...

I personally don't have a strong opinion about this. Knowing the scope of the error is useful, but having the whole buffer turn red is not. Do other language plugins have better way to deal with this issue ? Maybe we could underline everything only when more information is asked about the error ? 

Since we had recently a nice collective discussion about keybindings, I will summon the people who took part in it to gather more feedback 🙂 

@Chris00 @mattiase @bbatsov @erikmd @catern

bbatsov:
Depends on what info we can get from the compiler about the error - in CIDER we try to map the errors to a single expression, even if it's multiline. Can someone post a screenshot of the problem, because I don't recall seeing crazy error highlighting in Merlin.

nojb:
> Can someone post a screenshot of the problem, because I don't recall seeing crazy error highlighting in Merlin.

![image](https://github.com/ocaml/merlin/assets/113560/74c65889-5e55-4954-9523-1559a1281e89)

bbatsov:
[...]

PR #1693, opened 2023-10-22
Emacs: Highlight only first error line by default

3Rafal:
As proposed in #1663


nojb:
I think the change makes sense. The typical example when the current (`nil`) behaviour is unreadable is:
```
module M : sig
 ...
end = struct
 ...
end
```
when the structure does not match the signature (with a "big" module).

Issue #1636, opened 2023-06-26
Merlin's time-space trade-off


pitag-ha:
The Merlin file caches currently stay alive for 5 min. Why 5 min? I'd love to know the impact on both time and space when changing that number.

Action points:
- [ ] Refine the Merlin telemetry to capture memory usage as well.
- [ ] Prepare a Merlin branch (without merging) that lets that number be set by the editor.
- [ ] Ask Jane Street to add the following logic to their Emacs plug-in: Set the number to 60 by default, but let the user configure it back down to 5 again (in case of memory issues).
- [ ] Observe the change in latency and memory.

3Rafal:
I'm wondering if it would be a good idea to add (optional?) data about cache hit/misses to merlin responses. It would be similar performance data as "timing" field. I think it would be helpful for telemetry, and we could use it in merl-an to verify caching and provide statistics of cache usage.

pitag-ha:
Currently, the info about cache hits/misses is dumped into the Merlin log-file. I agree that for `merl-an`, it would be simpler if it was added to the Merlin telemetry directly. Making that optional sounds like a good idea to me.

PR #1698, opened 2023-10-30
Configurable cache retention period

3Rafal:
Second task from #1636

PoC:
The cache retention value was added to Mconfig. Call to `Mocaml.flush_caches` was moved to a place where Mconfig is available. Next step will be to create a flag that allows setting the config value from client. Last step is to use this flag from emacs/vim plugins.

### DIFF
--- a/src/kernel/mconfig.ml
+++ b/src/kernel/mconfig.ml
@@ -92,8 +92,9 @@ type merlin = {
   flags_applied : string list with_workdir list;
 
   failures : string list;
-  extension_to_reader : (string * string) list
+  extension_to_reader : (string * string) list;
 
+  cache_period : int
 }
 
 let dump_merlin x =
@@ -129,7 +130,8 @@ let dump_merlin x =
           "extension", `String suffix;
           "reader", `String reader;
         ]) x.extension_to_reader
-    )
+    );
+    "cache_period"   , Json.string (string_of_int x.cache_period)
   ]
 
 module Verbosity = struct
@@ -357,6 +359,15 @@ let merlin_flags = [
     "-ocamllib-path",
     marg_path (fun path merlin -> {merlin with stdlib = Some path}),
     "<path> Change path of ocaml standard library"
+  );
+  (
+    "-cache-period",
+    Marg.param "int" (fun prot merlin ->
+        try {merlin with cache_period = (int_of_string prot)}
+        with _ -> invalid_arg "Valid value is int";
+      ),
+    "Change file cache retention period. It's measured in minutes. \
+      Default value is 5."
   );
   (
     (* Legacy support for janestreet. Ignored. To be removed soon. *)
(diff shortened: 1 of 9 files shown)


voodoos:
Is that PR ready for review ? Looks like everything is in place :-) 

Maybe we could write a test which show cache hits when the flag has a very high value and no cache hit when it is set to 0.

3Rafal:
@voodoos I was considering extending vim and emacs plugins, so that they can use this new flag. But this PR is actually nice and self-contained, so I'll just write tests and ask you for review when ready :)

src/kernel/mconfig.ml
```diff
@@ -358,6 +360,14 @@ let merlin_flags = [
     marg_path (fun path merlin -> {merlin with stdlib = Some path}),
     "<path> Change path of ocaml standard library"
   );
+  (
+    "-cache-time",
+    Marg.param "int" (fun prot merlin ->
+        try {merlin with cache_time = (int_of_string prot)}
+        with _ -> invalid_arg "Valid value is int";
+      ),
+    "Change file cache retention time. Measured in minutes."
```
voodoos, line 369: Might be useful to document the default value (5) here.

voodoos (approved):
Looks good to me. The test is very nice :-)

src/kernel/mconfig.ml
3Rafal: Done

3Rafal:
@voodoos , I've added emacs handling, because it was super simple. Here is logged result of using it (as a proof that it actually works). It's a simple query run on merlin codebase:
```
command:  ocamlmerlin server type-enclosing -protocol sexp -verbosity 0 -filename /Users/rafal/Projects/Tarides/merlin/src/kernel/mconfig.ml -cache-period 45 -position 73:3 -index 0
- result: ((assoc) (class . "return") (value ((assoc) (start (assoc) (line . 72) (col . 0)) (end (assoc) (line . 98) (col . 1)) (type . "type merlin = {
  build_path : string list;
  source_path : string list;
  cmi_path : string list;
  cmt_path : string list;
  extensions : string list;
  suffixes : (string * string) list;
  stdlib : string option;
  reader : string list;
  protocol : [ `Json | `Sexp ];
  log_file : string option;
  log_sections : string list;
  config_path : string option;
  use_ppx_cache : bool;
  exclude_query_dir : bool;
  flags_to_apply : string list with_workdir list;
  flags_applied : string list with_workdir list;
  failures : string list;
  extension_to_reader : (string * string) list;
  cache_period : int;
}") (tail . "no"))) (notifications) (timing (assoc) (clock . 60) (cpu . 37) (query . 4) (pp . 0) (reader . 4) (ppx . 0) (typer . 29) (error . 0)))
```

voodoos:
Perfect !

Thank you !

PR #1700, opened 2023-10-31
uid_of_path can use the decl_id if shapes fail

### LINKED ISSUE #1699: Missing a source digest when looking for an implementation
To reproduce:

- Download the attached tarball and expand
- `cd buggy`
- `source dune-script`
- `ocamlmerlin single locate -position 4:40 -filename mmc/src/hack.ml -look-for implementation -build-path m_kernel/experimental/typed/.m_kernel_experimental_typed.objs/byte -build-path m_kernel/unix/src/.m_kernel_unix.objs/byte -source-path mmc/src -source-path m_kernel/unix/src -source-path m_kernel/experimental/typed < mmc/src/hack.ml`

Observe that the response refers to `m_kernel/unix/src/event_reader_intf.ml` instead of the correct `m_kernel/experimental/typed/event_reader_intf.ml`.

Looking at the logs, it seems that the problem is that the implementation lookup fails to find a uid, and thus fails to get a source digest to discern between the two files of the same name.

This is admittedly an unusual setup, but it's relevant within Jane Street.

[merlin-bug.tar.gz](https://github.com/ocaml/merlin/files/13210080/merlin-bug.tar.gz)

goldfirere:
Fixes #1699. Written with direction from @lpw25.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ merlin NEXT_VERSION
   + merlin binary
     - Fix a follow-up issue to the preference of non-ghost nodes introduced in #1660 (#1690, fixes #1689)
     - Add `--cache-period` flag, that sets cache invalidation period. (#1698)
+    - Fix Merlin locate not fallbacking on the correct file in case of ambiguity
+      (@goldfirere, #1699)
   + editor modes
     - vim: load merlin when Vim is compiled with +python3/dyn (e.g. MacVim)
 
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -182,7 +182,9 @@ end = struct
   let reset () = state := None
 
   let move_to ~digest file =
-    log ~title:"File_switching.move_to" "%s" file;
+    log ~title:"File_switching.move_to" "file: %s\ndigest: %s" file
+    @@ Digest.to_hex digest;
+
     state := Some { last_file_visited = file ; digest }
 
   let where_am_i () = Option.map !state ~f:last_file_visited
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -361,14 +361,15 @@ let uid_of_path ~config ~env ~ml_or_mli ~decl_uid path namespace =
         ~namespace:Shape.Sig_component_kind.Module env (Pident id)
     end)
   in
-  match ml_or_mli with
-  | `MLI ->
-    let uid = scrape_alias ~fallback_uid:decl_uid ~env ~namespace path in
-    log ~title:"uid_of_path" "Declaration uid: %a"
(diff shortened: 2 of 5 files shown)


goldfirere:
I don't know how best to test this within the merlin testsuite; advice welcome.

voodoos:
Thanks for having a look.

For testing you should be able to reproduce your script in a single-file cram test by cd-ing and calling the compiler manually.
I wonder whether it could be simplified a bit since the `hack.ml` file only references the `_intf` modules?
You could also write a `.merlin` file to make the command line call to Merlin simpler.

goldfirere:
Those new test case failures look (slightly) troublesome. Would you have a chance to look @voodoos? I don't know my way around this code very well and provided the fix in this patch only with @lpw25's guidance.

voodoos:
I will have a look!

voodoos:
I started looking at this and there is two different kinds of fall backing:
- If, looking for a definition, we fail to get an `uid` we fallback to looking the for the `uid` of the declaration. This fallback can also fail, as is the case in the broken test because it also relies on loading an external `cmt`.
- We can also fallback to a `loc` that is found by looking for the item in the environment (which rely only on the cmi file) and should also correspond to the declatation.

I'm wondering whether these two fallbacks lead to the same loc in which case we could only rely only on the second one which doesn't require the cmt files.

I will try to rework that mechanism and make it clearer!

lpw25:
These two fallbacks will lead to the same location but the `uid` one has more information which enables it to distinguish between different files with the same name. The aim of this PR was to make sure we used the `uid` rather than the `loc`. The test shows that sometimes the `uid` one can fail due to missing files, so I guess we should first try the `uid` and then fallback again to the `loc`.

voodoos:
Yes, if finding the definition's uid failed because of a missing `cmt`, looking for information about the declaration's uid will often fail similarly. I agree that we should fallback on the "lookup `loc`" if that happens. I will work on reproducing the issue in a test and fixing it cleanly.

lpw25:
Sorry, to be clear, we want to fall back to the declaration's `uid` when the definitions `uid` could not be found, and only if that `uid` lookup failed should we use the `loc`.

voodoos:
Yes, that's what I meant :-)

voodoos:
I pushed a commit that should fix the issue with all tests passing.
(I still need to add a test illustrating the initial issue)

voodoos:
@goldfirere I added a test and force pushed to illustrate the impact of these changes.

There is still an issue though: when both files have exactly the same content, they have the same digest, we should probably salt the digest...

lpw25:
FWIW the remaining issue a long-standing one that thankfully doesn't occur very often

PR #1706, opened 2023-11-15
[FEATURE] Command to provide information for OCaml syntax

### LINKED ISSUE #1692: Feature: provide information about the syntax
OCaml's syntax can be puzzling for new users and sometime prove to be challenging to search on the internet due to the use of special characters. On the other hand, the manual and compiler's `parsetree.mli` provides useful comments when one knows where to look for: that is the name of the features.

The proposition is to add a command to Merlin that, given a certain OCaml syntax, will display the corresponding Parsetree node name and a documentation string. For example, on `module F (M : T) =` it will give the hints “Pmod_functor” and “Functor declaration: [functor(X : MT1) -> ME]”.

Tasks:
- Boilerplate code to add a new command to the protocol and associated tests.
- Identify the most interesting node around the cursor. (It would be natural to do it on the non-preprocessed AST for this feature, but Merlin only has utilities that work on the Typedtree.)
- Lookup documentation for the interesting nodes, in the sources themselves and in a complementary table.
- Update `ocaml-lsp` to also provide that information as a code action or in addition to the standard `document` request.

PizieDust:
closes #1692 

# FEATURE: New Merlin Command - Syntax Document

## Introduction 
The `Document` command of Merlin returns comments written in code. This PR introduces a new command, the `Syntax Document` command which takes the syntax under the cursor, analyses it's TypedTree nodes and returns specific documentation for the Syntax.

### Code
```ocaml
type point = { x:int; y:int }
```
### Output
```json
{
   "name": "Record types",
   "description": "Allows you to define variants with a fixed set of fields, and all of the constructors for a record variant type must have the same fields",
   "url": "https://v2.ocaml.org/manual/extensiblevariants.html"
}
```
This command can also identify specific 
### Code
```ocaml
type point = private { x:int; y:int }
```
### Output
```json
{
   "name": "Private Record Types",
   "description": "Values of a record type declared private can be de-structured via the expr . field notation. However, values of these types cannot be constructed directly by record construction.",
   "url": "https://v2.ocaml.org/releases/5.1/htmlman/privatetypes.html#ss:private-types-variant"
}
```

## Scope
This command for a start covers `Language Extensions` of OCaml. The covered sections include:
- [x] Recursive definition of values
- [x] Recursive modules
- [x] Recovering the type of a module
- [x] Signature Substitutions (Destructive substitutions, Local substitutions)
- [x] Extensible Variant types
- [x] Variant types
- [x] Abstract types
- [x] Record types
- [x] Private types
- [x] Empty variant types
- [x] Locally abstract data types
- [ ] Bigarray access
- [ ] Type-level module aliases
- [ ] Overriding in open statements
- [ ] Generalized algebraic datatypes
- [ ] Attributes
- [ ] Generative functors
- [ ] Extension operators **
- [ ] Inline records ==== extensible variants
- [ ] Alerts
- [ ] Binding operators

More sections will be covered and updated to the command along with other parts of the OCaml programming language.

cc @voodoos @pitag-ha


src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,16 @@
+type info = {
+  name: string;
+  description: string;
+  example: string;
+  documentation: string;
+}
+
+let node_infos : info list = [
```
voodoos, line 8: The more I look at it, the more I am convinced that we don't need this intermediate database since we have no other usage for it than the syntax-doc command.

Let's simply do the matching here, and return the records directly:

```
match node with
| a_type_extension -> { name="Type extension"; description...}
l ... -> ...
```

src/frontend/ocamlmerlin/query_json.ml
```diff
@@ -112,6 +112,14 @@ let dump (type a) : a t -> json =
         );
       "position", mk_position pos;
     ]
+  | Syntax_document (identifier, pos) ->
+    mk "syntax-document" [
+      "identifier", (match identifier with
```
voodoos, line 117: Is there a specific reason why we need an identifier to perform the query ?

src/frontend/query_commands.ml
```diff
@@ -502,6 +502,47 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
     if path = "" then `Invalid_context else
       Locate.get_doc ~config
         ~env ~local_defs ~comments ~pos (`User_input path)
+  | Syntax_document (patho, pos) ->
+    let typer = Mpipeline.typer_result pipeline in
+    let pos = Mpipeline.get_lexing_pos pipeline pos in
+    let path =
```
voodoos, line 508: You don't use that at all, except for checking if it's empty. We probably don't need it. The only interesting thing is the node under the cursor.

src/frontend/query_commands.ml
```diff
@@ -502,6 +502,47 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+    if path = "" then `Invalid_context else 
+      let node = Mtyper.node_at typer pos in
+      let get_node_parent = Mbrowse.drop_leaf node in
+      let node_parent = Option.get get_node_parent in
+      let raw_node = snd (Mbrowse.leaf_node node) in
+      let node_name = Mbrowse.print_node () raw_node in 
+      Format.eprintf "%s" (Mbrowse.print() node);
+      let info = 
+        begin 
+          match node_parent, node with
```
voodoos, line 526: This matching is part of the analysis, we should move it to the `Syntax_doc` module.

src/frontend/query_commands.ml
```diff
@@ -502,6 +502,47 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+        let path = Mreader_lexer.identifier_suffix path in
+        let path = List.map ~f:(fun {Location. txt; _} -> txt) path in
+        String.concat ~sep:"." path
+    in
+    if path = "" then `Invalid_context else 
+      let node = Mtyper.node_at typer pos in
+      let get_node_parent = Mbrowse.drop_leaf node in
+      let node_parent = Option.get get_node_parent in
+      let raw_node = snd (Mbrowse.leaf_node node) in
+      let node_name = Mbrowse.print_node () raw_node in
```
voodoos, line 522: This printer is here for debugging purposes, I don't think we should rely on it for any program logic.

If you want to keep a "database", you could use a stronger type than `string` for the keys like a variant (or even a polymorphic variant).

But we might not need an intermediary database at all, unless we can think of another usage ?

src/frontend/ocamlmerlin/query_json.ml
PizieDust: I understand an identifier to be like, a user selected/highlighted some code, if this is the wrong idea then I don't think we need an identifier to run the query and will probably remove it alltogether

src/frontend/query_commands.ml
PizieDust: I agree. An intermediary database seems too much for our use case, especially as right now we don't have any other use for it. I'll use the earlier recommendation

src/frontend/ocamlmerlin/query_json.ml
voodoos: I think for the first iteration we can restrict ourselves to the position of the cursor and the node(s) around that 👍

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,43 @@
...
+  name: string;
+  description: string;
+  documentation: string;
+}
+
+let get_syntax_doc node_parent node =
+  let info = 
+    begin 
+      match node_parent, node with
+      | (_, Type_kind _) :: _, (_, Constructor_declaration _) :: _ ->
```
voodoos, line 13: If I am not mistaken, you don't need `node_parent` here. The parent is already in the list `node` so you can simply match on the few first elements of that list:

```
match node with
| (_, Constructor_declaration _) :: (_, Type_kind _) :: _ ->  
```

(or something similar)

voodoos (commented):
That's much better !

Now is probably a good time to decide which features we actually want to document.

There are two possibilities here:
- If the feature is aimed at complete beginner, it might make sense to document as much of the syntax as we can.
- Or we just want to shed a light on weird / rarely use OCaml features that might puzzle even seasoned developers.

@trefis, @pitag-ha, what's your feeling about this ?

It might also depend on how the feature is used: is it called on every on-hover in LSP ? Or only on specific code actions ?

src/analysis/syntax_doc.ml
PizieDust: Yeah that's right

voodoos:
@PizieDust with discussed with Trefis and Sonja and for the scope of the feature we think you can first focus on "language extensions". These are advanced features that were added to OCaml other the time and are often not well known y beginners.

You can find the list of these features here:
https://v2.ocaml.org/releases/5.1/htmlman/extn.html

About the UI, we can start (after the backend part in Merlin is done) with extending `hover` with the new information then make it configurable. (And in fact, rework the hover command in general, it could be more useful than it is now in some situations).

PizieDust:
Amazing. I should be moving this PR from draft soon

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -0,0 +1,82 @@
+// Recursive defition of functions
+  $ $MERLIN single syntax-document -position 1:6 \
+  > -filename ./recursive-def.ml < ./recursive-def.ml
```
voodoos, line 3: Where are the tests results ?  You should run the test and then `dune promote` the results :-) 

Also, it make the tests hard to read to have every case in a different file and all the commads in a single other.
Don't feel obliged to change it now, but it would have been clearer to use the `cat >***.ml <<'EOF'` syntax you can find in many other tests to have the content of the file displayed right before the call to Merlin. Like in this test for example:

https://github.com/ocaml/merlin/blob/master/tests/test-dirs/document/unattached-comment.t#L4-L16

src/frontend/query_protocol.ml
```diff
@@ -133,6 +133,12 @@ type _ t =
        | `Not_found of string * string option
        | `No_documentation
        ] t
+  | Syntax_document
+    : Msource.position
+    ->[ `Found of string
+      | `Invalid_context
+      | `No_documentation
```
voodoos, line 140: What difference do you make between these two return values ?
It looks like we never return `` `Invalid_context``.

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,128 @@
...
+  name: string;
+  description: string;
+  documentation: string;
+}
+
+let get_syntax_doc node =
+  let info =
+    begin
+      match node with
+      | (_, Type_kind _) :: (_, Type_declaration _) :: (_, With_constraint Twith_typesubst _) :: _ ->
```
voodoos, line 13: We try to always keep lines under 80 characters. It's going to be challenging on this file.
I've ran ocamlformat on it, you can reuse the result:

```
open Browse_raw

type info = { name : string; description : string; documentation : string }

let get_syntax_doc node =
  let info =
    match node with
    | (_, Type_kind _)
      :: (_, Type_declaration _)
      :: (_, With_constraint (Twith_typesubst _))
      :: _ ->
        Some
          {
            name = "Signature Substitution - Destructive substitutions";
            description =
              "Behaves essentially like normal signature constraints, but it \
               additionally removes the redefined type or module from the \
               signature.";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/signaturesubstitution.html#ss:destructive-substitution";
          }
    | (_, Type_kind _)
      :: (_, Type_declaration _)
      :: (_, Signature_item ({ sig_desc = Tsig_typesubst _; _ }, _))
      :: _ ->
        Some
          {
            name = "Signature Substitution - Local substitution";
            description =
              "Local substitutions behave like destructive substitutions \
               `(with ... := ...)` but instead of being applied to a whole \
               signature after the fact, they are introduced during the \
               specification of the signature, and will apply to all the items \
               that follow.";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/signaturesubstitution.html#ss:local-substitution";
          }
    | (_, Module_type _)
      :: (_, Module_type _)
      :: ( _,
           Module_type_constraint
             (Tmodtype_explicit
               { mty_desc = Tmty_with (_, [ (_, _, Twith_modtype _) ]); _ }) )
      :: _ ->
        Some
          {
            name = "Signature Substitution - Module substitution";
            description =
              "Module type substitution essentially behaves like type \
               substitutions. They are useful to refine an abstract module \
               type in a signature into a concrete module type,";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/signaturesubstitution.html#ss:module-type-substitution";
          }
    | (_, Constructor_declaration _)
      :: (_, Type_kind (Ttype_variant _))
      :: (_, Type_declaration { typ_private = Public; _ })
      :: _ ->
        Some
          {
            name = "Variant Types";
            description =
              "Lets you represent data that may take on multiple different \
               forms.";
            documentation =
              "https://v2.ocaml.org/manual/coreexamples.html#s:tut-recvariants";
          }
    | (_, Type_kind Ttype_open)
      :: (_, Type_declaration { typ_private = Public; _ })
      :: _ ->
        Some
          {
            name = "Extensible variant types";
            description =
              "Can be extended with new variant constructors using +=.";
            documentation =
              "https://v2.ocaml.org/manual/extensiblevariants.html";
          }
    | (_, Type_kind Ttype_abstract)
      :: (_, Type_declaration { typ_private = Public; _ })
      :: _ ->
        Some
          {
            name = "Abstract variant types";
            description =
              "Allows you to define variants with arbitrary data structures, \
               including other variants, records, and functions";
            documentation =
              "https://v2.ocaml.org/manual/extensiblevariants.html";
          }
    | (_, Type_kind (Ttype_record _))
      :: (_, Type_declaration { typ_private = Public; _ })
      :: _ ->
        Some
          {
            name = "Record variant types";
            description =
              "Allows you to define variants with a fixed set of fields, and \
               all of the constructors for a record variant type must have the \
               same fields";
            documentation =
              "https://v2.ocaml.org/manual/extensiblevariants.html";
          }
    | (_, Type_kind _) :: (_, Type_declaration { typ_private = Public; _ }) :: _
      ->
        Some
          {
            name = "Empty variant types";
            description =
              "This extension allows the user to define empty variants.";
            documentation = "https://v2.ocaml.org/manual/emptyvariants.html";
          }
    | _
      :: (_, Constructor_declaration _)
      :: (_, Type_kind (Ttype_variant _))
      :: (_, Type_declaration { typ_private = Private; _ })
      :: _ ->
        Some
          {
            name = "Private Variant Types";
            description =
              "Values of a variant type declared private can be de-structured \
               normally in pattern-matching. However, values of these types \
               cannot be constructed directly by constructor application.";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/privatetypes.html#ss:private-types-variant";
          }
    | _ :: _
      :: (_, Label_declaration _)
      :: (_, Type_kind (Ttype_record _))
      :: (_, Type_declaration { typ_private = Private; _ })
      :: _ ->
        Some
          {
            name = "Private Record Types";
            description =
              "Values of a record type declared private can be de-structured \
               via the expr . field notation. However, values of these types \
               cannot be constructed directly by record construction.";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/privatetypes.html#ss:private-types-variant";
          }
    | (_, Type_kind Ttype_open)
      :: (_, Type_declaration { typ_private = Private; _ })
      :: _ ->
        Some
          {
            name = "Private Extensible Types";
            description =
              "Enable libraries to reveal some, but not all aspects of the \
               implementation of a type to clients of the library";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/privatetypes.html#start-section";
          }
    | (_, Type_kind Ttype_abstract)
      :: (_, Type_declaration { typ_private = Private; _ })
      :: _ ->
        Some
          {
            name = "Private Type Abbreviations";
            description =
              "A private type abbreviation declares a type that is distinct \
               from its implementation type `typexpr`.";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/privatetypes.html#ss:private-types-abbrev";
          }
    | (_, Expression _)
      :: (_, Expression _)
      :: (_, Value_binding _)
      :: (_, Structure_item ({ str_desc = Tstr_value (Recursive, _); _ }, _))
      :: _ ->
        Some
          {
            name = "Recursive definitions of values";
            description =
              "The `let rec` binding construct, in addition to the definition \
               of recursive functions, also supports a certain class of \
               recursive definitions of non-functional values, such as `let \
               rec name1 = 1 :: name2 and name2 = 2 :: name1 in expr` which \
               binds `name1` to the cyclic list `1::2::1::2::…`, and `name2` \
               to the cyclic list `2::1::2::1::…`";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/letrecvalues.html";
          }
    | [ (_, Structure _) ] ->
        Some
          {
            name = "Documentation comments";
            description =
              "Automatically converted during parsing into attributes to allow \
               tools to process them as documentation";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/doccomments.html";
          }
    | (_, Module_expr _)
      :: (_, Module_type { mty_desc = Tmty_typeof _; _ })
      :: _ ->
        Some
          {
            name = "Recovering the type of a module";
            description =
              "The construction `module type of module-expr` expands to the \
               module type (signature or functor type) inferred for the module \
               expression `module-expr`. ";
            documentation =
              "https://v2.ocaml.org/releases/5.1/htmlman/moduletypeof.html";
          }
    | node -> (
        let n = List.rev node in
        match n with
        | _ :: (_, Structure_item ({ str_desc = Tstr_recmodule _; _ }, _)) :: _
          ->
            Some
              {
                name = "Recursive modules";
                description =
                  "Recursive module definitions, introduced by the `module rec \
                   …and …` construction, generalize regular module definitions \
                   `module module-name = module-expr` and module \
                   specifications `module module-name : module-type` by \
                   allowing the defining `module-expr` and the `module-type` \
                   to refer recursively to the module identifiers being \
                   defined.";
                documentation =
                  "https://v2.ocaml.org/releases/5.1/htmlman/recursivemodules.html";
              }
        | _ :: _
          :: ( _,
               Value_binding
                 { vb_expr = { exp_extra = [ (Texp_newtype' _, _, _) ]; _ }; _ }
             )
          :: _ ->
            Some
              {
                name = "Locally Abstract Type";
                description =
                  "The expression `fun ( type typeconstr-name ) -> expr` \
                   introduces a type constructor named `typeconstr-name` which \
                   is considered abstract in the scope of the sub-expression, \
                   but then replaced by a fresh type variable.";
                documentation =
                  "https://v2.ocaml.org/releases/5.1/htmlman/locallyabstract.html";
              }
        | _ :: _
          :: ( _,
               Value_binding
                 {
                   vb_expr =
                     {
                       exp_desc = Texp_pack _;
                       exp_extra =
                         [
                           ( Texp_constraint { ctyp_desc = Ttyp_package _; _ },
                             _,
                             _ );
                         ];
                       _;
                     };
                   _;
                 } )
          :: _ ->
            Some
              {
                name = "First class modules";
                description =
                  "The expression ( `module module-expr : package-type` ) \
                   converts the module (structure or functor) denoted by \
                   module expression `module-expr` to a value of the core \
                   language that encapsulates this module.";
                documentation =
                  "https://v2.ocaml.org/releases/5.1/htmlman/firstclassmodules.html";
              }
        | _ -> None)
  in
  match info with
  | Some info ->
      `Found
        (Printf.sprintf "%s: \n%s \n%s" info.name info.description
           info.documentation)
  | None -> `No_documentation
```

voodoos (commented):
I made a (very incomplete) round of review: could you reformat the file (I provided the result of ocamlformating it in a comment, and refactor / promote the result of the tests so that they can be used to understand the expected behavior ?

src/frontend/query_protocol.ml
PizieDust: I think I need to remove `Invalid_context`. It could have made sense if I was checking that the position actually contains an interesting node but since I am not currently doing that, I think I'll best remove it.

Great catch

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,128 @@
...
+                  description = "The expression ( `module module-expr : package-type` ) converts the module (structure or functor) denoted by module \
+                                 expression `module-expr` to a value of the core language that encapsulates this module.";
+                  documentation = "https://v2.ocaml.org/releases/5.1/htmlman/firstclassmodules.html";
+          }
+      | _ -> None      
+    end
+  in
+  match info with
+  | Some info -> `Found (Printf.sprintf "%s: \n%s \n%s" info.name info.description info.documentation)
+  | None -> `No_documentation
```
pitag-ha, line 128: For a good separation between what's done in the "backend" (i.e. here) and what's done in the "frontend" (i.e. in `frontend/`), it could be good to return the `info` value as is here. That type can then be managed further by `Query_commands`/`Query_protocol` and be turned into a textual structure (i.e. into a JSON or a string) in `ocamlmerlin/` and in `ocaml-lsp-server`.

Btw, similarly to the other commands, the `ocamlmerlin` binary could turn this into a JSON that reflects your `info` type (`{"name" : (...), "description" : (...), "documentation" : (...)}`). That's easier to process by the plug-ins than this long string here. You can have a look at the tests of other commands such as `type-enclosing` or `locate` to see how the response value is usually structured.

@voodoos , is that also something you had in mind? Haven't you mentioned it yet, because it's quite low-level (i.e. it's a comment mostly on the implementation and only slightly on the behavior)? @PizieDust , let me know if you'd like to focus on polishing the behavior for now and think about the implementation structure later.

pitag-ha (commented):
This is super cool, @PizieDust , thanks a lot!

Are you planning to add an mli-file for `syntax_doc.ml` once you won't change that file that much anymore?

tests/test-dirs/syntax-document/language-extensions.t/run.t
PizieDust: Yes, this was the initial structure I was using. If it's better I'll revert back to it.

src/analysis/syntax_doc.ml
PizieDust: @pitag-ha 
> Are you planning to add an mli-file for syntax_doc.ml once you won't change that file that much anymore?

Yes, thank you. I just pushed a commit with the `.mli` file.

voodoos: We definitely want to keep the structured information as long as possible that is as a return type for the command and with a json output that reflect it.

pitag-ha: > We definitely want to keep the structured information as long as possible that is as a return type for the command and with a json output that reflect it.

:+1: 

@PizieDust, let us know if you have questions about this :)

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,280 @@
...
+            name = "Recovering the type of a module";
+            description =
+              "The construction `module type of module-expr` expands to the \
+               module type (signature or functor type) inferred for the module \
+               expression `module-expr`. ";
+            documentation =
+              "https://v2.ocaml.org/releases/5.1/htmlman/moduletypeof.html";
+          }
+    | node -> (
+        let n = List.rev node in
```
voodoos, line 209: I was intrigued by this `List.rev` and the matching that follows. It has two consequences:

- It always find the recursive module for every cursor position inside it. But we only want to show the documentation when the cursor is on the `rec` keyword (and maybe the `module` keyword)

- It won't match on nested recursive modules: 
```
module B = struct 
  module rec A : sig 
  ...
  end
end
```

I don't know what merlin identifies as the "node under the cursor" when on the `rec` keyword, but if this is not what we need then we should change that :-)

src/analysis/syntax_doc.ml
PizieDust: @pitag-ha @voodoos 
This was a great idea. @jfeser left a comment [the comment](https://github.com/ocaml/ocaml-lsp/pull/1218#issuecomment-1852265222) about using [Go to Documentation]() type links instead of adding the full `url to the UX and returning the results as a `JSON will make this a breeze.

I've made some modifications and the new output should correctly return a `JSON:
```
 // Variuant types
   $ $MERLIN single syntax-document -position 1:1 \
   > -filename ./variant-types.ml < ./variant-types.ml
+  {
+    "class": "return",
+    "value": {
+      "name": "Extensible variant types",
+      "description": "Can be extended with new variant constructors using +=.",
+      "url": "https://v2.ocaml.org/manual/extensiblevariants.html"
+    },
+    "notifications": []
+  }
```
This update will be included in the next commit

src/frontend/query_protocol.ml
```diff
@@ -96,6 +96,13 @@ type error_filter = {
   typing : bool;
 }
 
+type info =
```
voodoos, line 99: Since there is a lot of other things defined in that module that are unrelated to the `Syntax_document` command we should use a more descriptive name like `syntax_info` or `syntax_document_result`.

One way to see that this name is not precise enough is is when you use it elsewhere like in ocaml-lsp: `Query_protocol.info` is really non-informative :-)

tests/test-dirs/syntax-document/language-extensions.t/first-class-modules.ml
```diff
@@ -0,0 +1,2 @@
+let sort (type s) (module Set : Set.S with type elt = s) l =
+  Set.elements (List.fold_right Set.add l Set.empty)
```
voodoos, line 2: Did you forget to move this one to the run.t file ?

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,286 @@
...
+    :: (_, Type_declaration { typ_private = Private; _ })
+    :: _ ->
+      {
+        name = "Private Types";
+        description =
+          "Values of a variant type declared private can be de-structured \
+           normally in pattern-matching. However, values of these types cannot \
+           be constructed directly by constructor application.";
+        documentation =
+          "https://v2.ocaml.org/releases/5.1/htmlman/privatetypes.html#ss:private-types-variant";
```
voodoos, line 123: Some urls start with `https://v2.ocaml.org/releases/5.1/htmlman` and others with `https://v2.ocaml.org/manual/`.
Is that expected ?

I guess we should always point to same version of the manual and `https://v2.ocaml.org/releases/5.1/htmlman/` are the more precise ones ?

You could make a small builder function that takes only the changing part of the url (like `privatetypes.html#ss:private-types-variant
`) and concatenates it to the base url. This way it would easier to change the manual version for other Merlin branches.

And in fact, you are targeting Merlin for OCaml 4.14 here, so you should point to the 4.14's manual 👍

voodoos (commented):
Nice, you're making good progress on this :-) 

I made a few comments, and you should also try to sort out the rebase mess :-)

src/analysis/syntax_doc.ml
PizieDust: Thank you. This is a brilliant idea.

src/frontend/ocamlmerlin/query_json.ml
```diff
@@ -1,442 +1,424 @@
...
+     The Software is provided "as is", without warranty of any kind, express or
+     implied, including but not limited to the warranties of merchantability,
+     fitness for a particular purpose and noninfringement. In no event shall
+     the authors or copyright holders be liable for any claim, damages or other
+     liability, whether in an action of contract, tort or otherwise, arising
+     from, out of or in connection with the software or the use or other dealings
+     in the Software.
 
-)* }}} *)
+   )* }}} *)
```
voodoos, line 27: We do not ocamlformat the codebase yet, and formating changes should generally not be mixed in unrelated PRs.
I only ocamlformated your `syntax_command` new file because it would have been very cumbersome to do it manually, but you should not format other files with it.

PizieDust: Oh my bad. Got it.

tests/test-dirs/syntax-document/language-extensions.t/first-class-modules.ml
PizieDust: I hadn't concluded with it's node tree so I hadn't moved it to the `run.t` file yet.

src/frontend/query_protocol.ml
```diff
@@ -142,8 +142,7 @@ type _ t =
        ] t
   | Syntax_document
     : Msource.position
-    ->[ `Found of syntax_doc_result
-      | `No_documentation
+    ->[ `Found of syntax_doc_result option
```
voodoos, line 145: I don't understand why you made that change ? The previous type was perfectly fine.
`Found None` reads like a contradiction...

src/frontend/ocamlmerlin/query_json.ml
voodoos: This is not resolved yet, the formatting is still wrong :-)

voodoos (commented):
I will need more time to fully review the matching in `syntax_doc.ml`.

Meanwhile you can focus on making the feature configurable in LSP.

src/frontend/ocamlmerlin/query_json.ml
PizieDust: Oh, just pushed it in the new commit. 👍

src/frontend/query_protocol.ml
PizieDust: making the return value `syntax_result option`  allows us to have more freedom in other implementations like `query_json` and `ocaml-lsp`. 
but looking at it now I agree it does look weird . I'll refactor it so it has a more meaningful structure.

voodoos: I don't see why it would give us more freedom in`query_json` or `ocaml-lsp` since we encode the exact same amount of information: we could simply return `syntax_doc_result option` since there is only two cases. Either that or custom variant like `` `Found`` and `` `No_documentation`. All these options encode the same things and allow for the same flexibility. We often encourage more verbose names instead of plain `option`, especially when writing protocols, since it allows future readers to understand better the purpose just by reading the types.

PizieDust: ah I see, I was going for simply returning `syntax_doc_result option` but I like the importance of using verbose names as you mentioned above. I think it makes more sence to use `Found` and `No_documentation`

voodoos:
Hi Pizie, I started reviewing one last time the PR.
I think it would be nice to have tests that show exactly where the feature triggers and where it does not.
I just pushed a commit doing that for the "destructive substitution". I introduced an alias to try to make it less verbose. Could you do the same for the other tests ?

src/frontend/ocamlmerlin/new/new_commands.ml
```diff
@@ -199,6 +199,21 @@ Otherwise, Merlin looks for the documentation for the entity under the cursor (a
     end
   ;
 
+  command "syntax-document"
+    ~doc: "Returns documentation for OCaml syntax and possibly some AST info for the most interesting entity under the cursor"
```
voodoos, line 203:
```suggestion
    ~doc: "Returns documentation for OCaml syntax for the entity under the cursor"
```

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -0,0 +1,227 @@
...
+> -filename ./sig-subs.ml < ./sig-subs.ml
+
+
+// Types
+  $ cat > types.ml << EOF 
+  > type a1 = ..;
+  > type a2 = A;
+  > type a3 = |;
+  > type a4 = {x: int};
+  > type a5 = int;
```
voodoos, line 138: There is a `  > EOF` missing here

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,286 @@
...
+      }
+  | _ :: _
+    :: ( _,
+         Value_binding
+           {
+             vb_expr =
+               {
+                 exp_desc = Texp_pack _;
+                 exp_extra =
+                   [ (Texp_constraint { ctyp_desc = Ttyp_package _; _ }, _, _) ];
```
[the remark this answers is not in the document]
voodoos: @PizieDust there is no test for that case :-)

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,243 @@
...
+  | (_, Module_type _)
+    :: (_, Module_type _)
+    :: ( _,
+         Module_type_constraint
+           (Tmodtype_explicit
+             { mty_desc = Tmty_with (_, [ (_, _, Twith_modtype _) ]); _ }) )
+    :: _ ->
+      Some
+        {
+          name = "Module substitution";
```
voodoos, line 48: Test is missing for that case

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -0,0 +1,228 @@
...
+  $ syn_doc 11:37 -filename ./sig-subs.ml < ./sig-subs.ml | jq '.value'
+  "No documentation found"
+
+// Local substitutions
+  $ syn_doc 16:12 \
+  > -filename ./sig-subs.ml < ./sig-subs.ml | jq '.value.name'
+  "Local substitution"
+// Module type substitutions
+$ syn_doc 26:58 \
+> -filename ./sig-subs.ml < ./sig-subs.ml
```
voodoos, line 129: This test is disabled since the two required white spaces are missing

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -0,0 +1,228 @@
...
+  > EOF 
+
+// Locally abstract data types
+on type t..
+  $ syn_doc 1:17 \
+  > -filename ./locally-abstract-dt.ml < ./locally-abstract-dt.ml | jq '.value.name'
+  "Locally Abstract Type"
+// Locally abstract data types B
+$ syn_doc 3:17 \
+ > -filename ./locally-abstract-dt.ml < ./locally-abstract-dt.ml
```
voodoos, line 226: This test is disabled since the two required white spaces are missing

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -0,0 +1,228 @@
...
+  > -filename ./p-types.ml < ./p-types.ml | jq '.value.name'
+  "Private Type Abbreviation"
+on type t = private int..
+  $ syn_doc 7:14 \
+  > -filename ./p-types.ml < ./p-types.ml | jq '.value.name'
+  "Private Type Abbreviation"
+on type t = int..
+  $ syn_doc 11:7 \
+  > -filename ./p-types.ml < ./p-types.ml | jq '.value.name'
+  "Abstract type"
```
voodoos, line 206: I think it's misleading to hint at `type t = int` being an abstract type, even if it is represented as such in the compiler. An abstract type as commonly understood is really a type with no definition / constructors at all like `type t`. Maybe we should check that there is no `ptype_manifest` in the abstract case of the matching

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -0,0 +1,247 @@
...
+  > let _png = Hashtbl.add devices "PNG" (module PNG : SVG)
+  > EOF
+on type picture
+  $ syn_doc 1:6 \
+  > -filename ./first-class-modules.ml < ./first-class-modules.ml | jq '.value.name'
+  "Record type"
+on (module SVG : DEVICE)
+  $ syn_doc 6:43 \
+  > -filename ./first-class-modules.ml < ./first-class-modules.ml | jq '.value.name'
+  "First class module"
```
voodoos, line 242: This is not testing what the comment says: 
`(module SVG : DEVICE)` is on line 8, not 6.

And it also shows that there is an issue: on line 6, this is not a first class module but a normal module. The output should be "no documentation".

And if I test it on line 8 it doesn't find a first class module, so there is definitely an issue here.

We should keep the two tests: on on line 6 and on on line 8 and fix the behavior.

(same remark applies to the next test)

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -0,0 +1,247 @@
...
+  > let sort_uniq (type s) (cmp : s -> s -> int) =
+  >   let module S = Set.Make(struct type t = s let compare = cmp end) in
+  >   fun l ->
+  >     S.elements (List.fold_right S.add l S.empty)
+  > EOF
+// Locally abstract data types
+on type t..
+  $ syn_doc 1:17 \
+  > -filename ./locally-abstract-dt.ml < ./locally-abstract-dt.ml | jq '.value.name'
+  "Locally Abstract Type"
```
voodoos, line 221: Note that it also triggers when the cursor is on the body of the function (`1:39`) which is not very satisfying.

We should at least show it with a test and add a FIXME comment before. 

We probably won't be able to improve on it without having additional location information from the compiler (it might even be already erased in the parsetree, we need a CST..)

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,231 @@
...
+          description =
+            "Define variants with arbitrary data structures, including other \
+             variants, records, and functions";
+          documentation = syntax_doc_url "typedecl.html#ss:typedefs";
+        }
+  | (_, Core_type _)
+    :: (_, Core_type _)
+    :: (_, Label_declaration _)
+    :: (_, Type_kind (Ttype_record _))
+    :: (_, Type_declaration { typ_private = Public; _ })
```
voodoos, line 97: We have lot of duplication depending on the type being public or private. You might want to refactor it: you could match for both public and private and then you update the texts with privacity information:

```
let name =" Record type" in
let name = match typ_private with
  | Public -> name
  | Private -> Format.sprintf "Private %s" name
in

let description = "Record type description." in
let name = match typ_private with
  | Public -> name
  | Private -> Format.sprintf "%s This type is private, values cannot be constructed directly but can be de-structured as usual." name
in
... 
```

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,231 @@
...
+          documentation = syntax_doc_url "typedecl.html#ss:typedefs";
+        }
+  | (_, Type_kind Ttype_abstract)
+    :: (_, Type_declaration { typ_private = Public; typ_manifest = None; _ })
+    :: _ ->
+      Some
+        {
+          name = "Abstract type";
+          description =
+            "Define variants with arbitrary data structures, including other \
```
voodoos, line 89:
```suggestion
            "Defines variants with arbitrary data structures, including other \
```

src/analysis/syntax_doc.ml
```diff
@@ -0,0 +1,231 @@
...
+  | _
+    :: (_, Constructor_declaration _)
+    :: (_, Type_kind (Ttype_variant _))
+    :: (_, Type_declaration { typ_private = Public; _ })
+    :: _ ->
+      Some
+        {
+          name = "Variant Type";
+          description =
+            "Represent data that may take on multiple different forms.";
```
voodoos, line 79:
```suggestion
            "Represents data that may take on multiple different forms.";
```

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -177,7 +176,7 @@ on type a5..
   >   let to_int n = n
   > end
   > EOF
-on type b1..
+  on type b1..
```
voodoos, line 179: This is a comment, it should start at the beginning of the line.

src/analysis/syntax_doc.ml
PizieDust: This was a fun refactoring to do

src/analysis/syntax_doc.ml
```diff
@@ -76,7 +76,7 @@ let get_syntax_doc node : syntax_info =
         {
           name = "Variant Type";
           description =
-            "Represent data that may take on multiple different forms.";
+            "Represents data that may take on multiple different forms.";
```
voodoos: Just a note: this is not plural (it's a verb), this is the third person singlar's `s` 🙂

src/analysis/syntax_doc.ml
```diff
@@ -219,7 +219,7 @@ let get_syntax_doc node : syntax_info =
              sub-expression and replaced by a fresh type variable.";
           documentation = syntax_doc_url "locallyabstract.html";
         }
-  | (_, Module_expr _) :: (_, Module_binding _) :: _ ->
+  | (_, Module_expr _) :: (_, Module_expr _) :: _ ->
```
voodoos: This is unrelated to the commit message, it should not be here. (No need to change it, we will probably squash most of the commits at the end anyway.)

tests/test-dirs/syntax-document/language-extensions.t/run.t
```diff
@@ -230,18 +233,38 @@ on type t..
...
+  $ syn_doc 6:6 \
+  > -filename ./first-class-modules.ml < ./first-class-modules.ml | jq '.value'
+  "No documentation found"
 on (module SVG : DEVICE)
-  $ syn_doc 6:43 \
+  $ syn_doc 8:43 \
   > -filename ./first-class-modules.ml < ./first-class-modules.ml | jq '.value.name'
   "First class module"
-on (module PNG : SVG)
-  $ syn_doc 7:43 \
```
voodoos: You could have kept the test for line 6 since it was actually showing incorrect behavior. It should be fixed now.

voodoos (commented):
Right, I think we just need to improve the `locally abstract data types` case before merging a first version of this PR.

src/frontend/ocamlmerlin/query_json.ml
```diff
@@ -380,6 +382,17 @@ let json_of_response (type a) (query : a t) (response : a) : json =
...
+    match resp with
+    | `Found info ->
+          `Assoc
+            [
+              ("name", `String info.name);
+              ("description", `String info.description);
+              ("url", `String info.documentation);
+            ]
+    | `No_documentation -> `String "No documentation found"
+      )
```
voodoos, line 395: Indentation is weird here.

src/frontend/query_commands.ml
```diff
@@ -502,6 +502,15 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
         ~env ~local_defs ~comments ~pos (`User_input path)
+  | Syntax_document pos ->
+    let typer = Mpipeline.typer_result pipeline in
+    let pos = Mpipeline.get_lexing_pos pipeline pos in
+    let node = Mtyper.node_at typer pos in
+    let res = Syntax_doc.get_syntax_doc pos node in 
+    (match res with
+    | Some res -> `Found res 
+    | None -> `No_documentation) 
+
```
voodoos, line 514: There is one empty line to many, and according to the rest of this file there should be an empty line before instead.

CHANGES.md
```diff
@@ -1,9 +1,11 @@
-merlin NEXT_VERSION
-===================
+merlin 4.13
+===========
+Fri Dec  1 15:00:42 CET 2023
```
voodoos, line 3: No need to change this header before the release itself.

voodoos (approved):
I think the PR is almost ready.
I just made a few formatting remarks.

Do you want to rebase and squash your changes into meaningful commits ?
If not I can simply squash everything before merging.

Issue #1680, opened 2023-09-21
Enrich Merlin telemetry with GC info


pitag-ha:
For our performance analysis, it would be very useful to enhance the Merlin telemetry.

### Current telemetry

Currently, the Merlin telemetry only contains time information. That timing information is broken down into the various Merlin phases: time spent during the reader phase, the preprocessor phase, the typer phase etc.

### Proposed telemetry additions

It would be extremely useful to add memory and general GC information to that. Concretely, when running an `ocamlmerlin` query, I'd love to see information about
- [ ] New memory allocation, i.e. the sum of memory allocated throughout the query execution.
- [ ] Total memory usage, i.e. how much memory was used in total by `ocamlmerlin`, including the cached memory from before. This can be captured in different ways. I think the total memory usage at its peak would be enough.
- [ ] (Possibly more info about the GC's work, e.g. for how long it has been running during the query.)

This new telemetry should be similarly broken down into the various Merlin phases as the timing telemetry.

### Motivation for telemetry additions

In general, for performance analysis, it's useful to have insight into the memory footprint. Concretely, I have three concrete use cases in mind:
1. I'd like to optimize the time/space trade-off. See https://github.com/ocaml/merlin/issues/1636 . This is the most important one.
2. I'd like to know if the time performance outliers we sometimes observe have some correlation with memory bursts.
3. Adding memory data to our continuous benchmarks might help us catch newly introduced memory leaks/ memory performance regressions.

### Discussion on CLI details

Do we want this enriched telemetry always or only when opted in in some way? As a heads-up: I'm also enriching the telemetry in other ways, so having full telemetry always would be quite verbose.

pitag-ha:
Implemented in https://github.com/ocaml/merlin/pull/1717

PR #1717, opened 2023-12-05
Report current heap usage on Merlin response

Engil:
This PR is a first step toward completing #1680

It enriches Merlin's responses with a new `heap_mbytes` field, that contains the total major heap size (in *megabytes*) after completing the request.

The implementation relies on using the `heap_words` field from  [Gc.stat](https://v2.ocaml.org/api/Gc.html).

This PR should be covering enough ground to satisfy the motivations 1 and 3 of the aforementined issue.
Motivation 2 maybe needs a finer grained solution with better granularity (ie. not only at the end of a request).

Let me know if there any refinement required, or if we need to add other fields, or if this is providing enough data as it is.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+merlin NEXT_VERSION
+===================
+
+  + merlin binary
+    - Add a "heap_mbytes" field to Merlin server responses to report heap usage (#1717)
+
 merlin 4.13
 ===========
 Fri Dec  1 15:00:42 CET 2023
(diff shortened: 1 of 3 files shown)


voodoos (approved):
The changes looks good to me. I guess the call to `Gc.quick_stat` is quite fast ?

Could you add a changelog entry ? Thank you !

Engil:
Done: is it ok?

voodoos:
> Done: is it ok?

It won't be after a rebase 🙂, I made a release yesterday so there is a new section to create in the changelog.

voodoos:
❤️

Issue #1722, opened 2024-01-25
FLG -pp ppx.exe -as-pp does not work on Windows


jonahbeckford:
### Problem

A `dune` stanza:

```scheme
(executable
 (name something)
 (modules something)
 (modes byte)
 (libraries DkSDKCoder_Std csexp)
 (preprocess
  (action
   (run %{lib:DkSDKCoder_PPX:ppx.exe} --as-pp %{input-file}))))
```

The merlin config from `dune ocaml merlin dump-config .`:

```scheme
 (FLG (-open Dune__exe))
 (FLG
  (-pp
   "C:\Users\beckf\AppData\Local\Programs\DkSDK\coder\h\main\lib\DkSDKCoder_PPX\ppx.exe
   --as-pp"))
 (FLG
  (-w
   @1..3@5..28@30..39@43@46..47@49..57@61..62@67@69-40
   -strict-sequence
   -strict-formats
   -short-paths
   -keep-locs
   -g))
```

During LSP session in vscode, I was getting `Uninterpreted extension` errors attached to PPX macros like `[%proc ...]`.

### Root Cause

By adding some custom logging into `ppxlib` I could see the following failures from `ppx.exe`:

```powershell
C:\Users\beckf\AppData\Local\Programs\DkSDK\coder\h\main\lib\DkSDKCoder_PPX\ppx.exe: too many input files. 
    First argument was
       C:\Users\beckf\AppData\Local\Temp\merlinppcc6fceQuickstart.ml.
    Next argument was
       1>C:\Users\beckf\AppData\Local\Temp\merlinppcc6fceQuickstart.ml.out.
    
    All arguments:
       C:\Users\beckf\AppData\Local\Programs\DkSDK\coder\h\main\lib\DkSDKCoder_PPX\ppx.exe --as-pp C:\Users\beckf\AppData\Local\Temp\merlinppcc6fceQuickstart.ml 1>C:\Users\beckf\AppData\Local\Temp\merlinppcc6fceQuickstart.ml.out.
```

Notice the `merlinpp` temporary file. The call flow is:

1. `Pparse.apply_pp` calls `System.run_in_directory` on line 162: https://github.com/ocaml/merlin/blob/857f17a6fdf997ce04227b4c28b88a2edb573782/src/ocaml/driver/pparse.ml#L150-L168
2. `System.run_in_directory` **incorrectly** asks for Command Prompt to redirect stdout with `1>` on line 777:
[...]

PR #1723, opened 2024-01-26
Fix FLG -pp ppx.exe -as-pp/-dump-ast on Windows

jonahbeckford:
Windows is special cased in Merlin to have a direct
 program execution without any shell so shell redirection
 does not work.

This is a narrow, Windows-only fix for
 https://github.com/ocaml/merlin/issues/1722

*I don't particularly like this one since it is based on heuristics.* And it does seem like there is a lot of unfixed code in other packages that is required to get `-as-pp` working on Windows. Perhaps someone more knowledgeable about Merlin can make this cleaner.


nojb (commented):
Exactly what I had in mind, thanks!

nojb (approved):
LGTM

Someone with commit privileges will need to approve/merge (cc @voodoos)

src/utils/std.ml
```diff
@@ -773,18 +773,21 @@ module System = struct
...
-             channel used by Merlin to communicate with the editor *)
-          if Sys.win32 then args else Format.sprintf "%s 1>&2" args
+      (* Runned program should never output on stdout since it is the
+          channel used by Merlin to communicate with the editor *)
+      let args =
+        if Sys.win32 then args
+        else
+          Printf.sprintf "%s 1>%s" args
+            (match stdout with | Some file -> Filename.quote file
+                               | None -> "&2")
```
voodoos, line 783:
```suggestion
          let stdout = match stdout with
            | Some file -> Filename.quote file
            | None -> "&2"
          in
          Printf.sprintf "%s 1>%s" args stdout
```

src/utils/std.ml
```diff
@@ -773,18 +773,21 @@ module System = struct
...
       in
       let cmd = Format.sprintf "%s %s" prog args in
       let exit_code =
         if Sys.win32 then
-          (* Note: the following function will never output to stdout *)
-          windows_merlin_system_command cmd ~cwd
+          (* Note: the following function will never output to stdout
+             when [stdout = None]. Instead stdout is sent to stderr. *)
```
voodoos, line 789:
```suggestion
          (* Note: the following function will never output to stdout.
             When [stdout = None], stdout is sent to stderr. *)
```

voodoos (approved):
I am not expert enough to comment the Windows code, it looks all right and I will trust your judgement on it.

The rest of the PR looks fine to me, I just made a few stylistic propositions.

nojb:
Friendly ping @voodoos: I think the PR can now be merged.

Issue #1724, opened 2024-01-30
Invalid arg from Marshal.from_bytes


jonahbeckford:
> Moved the "Second Attempt" problem from https://github.com/ocaml-ppx/ppxlib/issues/466 to this issue

Context:

```
 (preprocess
  (action
   (run %{lib:TheNameOfTheLibrary:ppx.exe} -as-pp %{input-file})))
```

Problem:

```text
[Error - 4:50:25 PM] Request textDocument/codeAction failed.
  Message: uncaught exception
  Code: -32603 
[object Object]
detached: /-----------------------------------------------------------------------
| Internal error: Uncaught exception.
| Invalid_argument("Marshal.from_bytes")
| Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
| Called from Stdlib__Marshal.from_string in file "marshal.ml" (inlined), line 67, characters 2-46
| Called from Merlin_kernel__Pparse.decode_potential_ast in file "src/ocaml/driver/pparse.ml", line 200, characters 14-47
| Called from Merlin_kernel__Mpipeline.process in file "src/kernel/mpipeline.ml", line 253, characters 10-120
| Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
| Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
| Called from Merlin_kernel__Mpipeline.timed_lazy in file "src/kernel/mpipeline.ml", line 17, characters 10-22
| Re-raised at Merlin_kernel__Mpipeline.timed_lazy in file "src/kernel/mpipeline.ml", line 19, characters 34-49
| Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
| Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
| Called from Merlin_kernel__Mpipeline.input_source in file "src/kernel/mpipeline.ml", line 111, characters 25-46
| Called from Merlin_kernel__Mpipeline.with_pipeline.(fun) in file "src/kernel/mpipeline.ml", line 115, characters 39-55
| Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
| Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
[...]

PR #1725, opened 2024-01-30
Use open_in_bin not open_in for merlinpp

jonahbeckford:
Fixes https://github.com/ocaml/merlin/issues/1724 thanks to @nojb spotting the problem in https://github.com/ocaml-ppx/ppxlib/pull/465.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ merlin NEXT_VERSION
     - Add a query_num field to the `ocamlmerlin` responses to detect server crashes (#1716)
   + editor modes
     - vim: load merlin under the ocamlinterface and ocamllex filetypes (#1340)
+    - Fix merlinpp not using binary file open (#1725, fixes #1724)
 
 merlin 4.13
 ===========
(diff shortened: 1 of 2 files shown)


jonahbeckford:
There are also two problematic other uses of `open_in`.

First is https://github.com/ocaml/merlin/blob/master/src/utils/misc.ml#L520-L537

Second is https://github.com/ocaml/merlin/blob/master/src/utils/std.ml#L822-LL839

Both of these are the same and are not used in the Merlin code base. I assume they are helpers for consumers of Merlin (but why is it duplicated?). Regardless, I think those should be changed as well, but that could break Merlin consumers.

nojb (approved):
LGTM

cc @voodoos who has been approving and merging these PRs lately :)

voodoos:
> There are also two problematic other uses of `open_in`.
> 
> First is https://github.com/ocaml/merlin/blob/master/src/utils/misc.ml#L520-L537
> 
> Second is https://github.com/ocaml/merlin/blob/master/src/utils/std.ml#L822-LL839
> 
> Both of these are the same and are not used in the Merlin code base. I assume they are helpers for consumers of Merlin (but why is it duplicated?). Regardless, I think those should be changed as well, but that could break Merlin consumers.

They might be leftovers since it's not even part of the upstream `Misc` module.

voodoos (approved):
Looks good to me, thank you @jonahbeckford

PR #1726, opened 2024-01-31
Jump to cases within a Match statement

### LINKED ISSUE #1721: Feature: jump to next/previous case
(This is an idea for a feature.)

If you're editing a large match expression, it could be nice to jump to the previous/next case, e.g.:

```ocaml
let f x =
  match x with
  | A ->
    let x1 = ... in
    let x2 = ... in
    ^
    (* many dozens of lines of code *)
  | B -> 3
```
If my cursor is as indicated in the `A` case, it would be nice to jump to the `B` case.

A sensible place to put this feature might be in the `jump` rpc. Right now, that supports jumping to the top of a `let`, `match`, or `function`; it might be easy to extend it to support `next-case` and `prev-case`.

(I started to look at this but it wasn't trivial so I thought I'd log this issue.)

PizieDust:
fixes #1721 
This PR allows for one to navigate the source code, particularly `match` constructs. From any point within the match statement, the command can move to the next case.

cc @voodoos


PizieDust:
will refactor to group similar functionality as currently there's a lot of code duplication

src/frontend/ocamlmerlin/new/new_commands.ml
```diff
@@ -341,6 +341,40 @@ match expression that contains the cursor
     end
   ;
 
+  command "jump-match-next-case"
```
voodoos, line 344: I think it would be much easier to adapt editors plugins by adding a new "target" to the existing ` jump` command instead of adding to very specific new commands.

PizieDust: Thank you for the review. I think the new commits have merged the logic with the pre-existing `jump` command

src/analysis/jump.ml
```diff
@@ -127,43 +127,101 @@ let rec skip_non_moving pos = function
...
+  get_match_case_pos typed_tree pos (fun pos case_loc -> pos.Lexing.pos_cnum < case_loc.pos_cnum) target
+
+let get_match_prev_case typed_tree pos target =
+  get_match_case_pos typed_tree pos (fun pos case_loc -> pos.Lexing.pos_cnum > case_loc.pos_cnum) target
+
+let get typed_tree pos target =
+  match target with
+  | "match-next-case" -> get_match_next_case typed_tree pos target
+  | "match-prev-case" -> get_match_prev_case typed_tree pos target
+  | _ ->
```
voodoos, line 188: This early match prevent you from reusing more of the existing code: You could use the match predicate the same as for the match target and only when a node has been found switch to your new logic.

src/analysis/jump.ml
```diff
@@ -127,43 +127,101 @@ let rec skip_non_moving pos = function
...
+          | Some (_, f) -> f
+          | None -> raise (No_predicate target)
+        )
+      in
+      if String.length target = 0 then
+        `Error "Specify target"
+      else begin
+        let nodes = skip_non_moving pos enclosings in
+        let node = find_node preds nodes in
+        let node_loc = Browse_raw.node_real_loc Location.none node in
```
voodoos, line 216: Here, if we found a match node and the target was a case, we switch to the new logic.

It's a shame that we have to break the elegant implementation. I wonder if Merlin already has a command to move to siblings and not only ancestors.

src/analysis/jump.ml
```diff
@@ -127,19 +127,54 @@ let rec skip_non_moving pos = function
...
+    match match_pred [node] with
+    | Some (Expression { exp_desc = Texp_match (_, cases, _); _ }) ->
+      let case = find_case_pos pos cases cmp_fn in      
+      (match case with
+      | Some location ->
+        `Found location
+      | None -> (match match_target with 
+            | "match-next-case" -> `Error "No next case found"
+            | "match-prev-case" -> `Error "No previous case found"
+            | _ -> raise (No_matching_target)))
```
voodoos, line 151: This last case is not reachable. This highlight the fact that passing [match_target] to that function is not a good fit. Instead you can make a new type `type direction = Prev | Next` and pass this more precise information.

src/analysis/jump.ml
```diff
@@ -127,19 +127,54 @@ let rec skip_non_moving pos = function
   | [] -> []
 ;;
 
+let rec find_case_pos pos cases cmp_fn =
+  match cases with
+  | [] -> None
+  | { c_lhs = { pat_loc; _ }; _ } :: tail ->
+    let case_loc = pat_loc.loc_start in
+    if cmp_fn pos case_loc then
+      Some case_loc
```
voodoos, line 136: Correct me if I am wrong, but I am under the impression that since you compare the cases from the beginning every time when trying to jump to the previous case you will always jump to the first one. This is not part of your tests:: try to jump from case ` | 'O' ->` to the previous one.

src/analysis/jump.ml
```diff
@@ -127,19 +127,54 @@ let rec skip_non_moving pos = function
...
+      raise (No_matching_target)
+  with
+  | No_matching_target ->
+    `Error "Cursor is not in a match predicate"
+
+let get_match_next_case node pos target =
+  get_match_case_pos node pos (fun pos case_loc -> pos.Lexing.pos_cnum < case_loc.pos_cnum) target
+
+let get_match_prev_case node pos target =
+  get_match_case_pos node pos (fun pos case_loc -> pos.Lexing.pos_cnum > case_loc.pos_cnum) target
```
voodoos, line 162: Location are hard. Having the location comparison logic passed as a higher order function to the `find_case_pos` function make it hard to reason about. (and it is indeed probably broken right now).

We should simply pass the direction (`Prev` or `Next`) information to the `find_case_pos` and have all the complex, error-prone, location comparison done there.

tests/test-dirs/motion/jump_match.t
```diff
@@ -0,0 +1,132 @@
...
+  >   true
+  > | 'I' -> 
+  >   true
+  > | 'O' -> 
+  >   true 
+  > | 'U' -> 
+  >   true 
+  > | _ -> 
+  >   false
+  > EOF
```
voodoos, line 73: Since the program is always the same, there is no reason to repeat it for each test, you can write it to a file once and for all at the beginning:
```
  $ cat >test.ml << EOF
  > let find_vowel x = 
  > match x with 
  > | 'A' -> 
  >   true
  > | 'E' -> 
  >   true
  > | 'I' -> 
  >   true
  > | 'O' -> 
  >   true 
  > | 'U' -> 
  >   true 
  > | _ -> 
  >   false
  > EOF
  
  $MERLIN single jump -target match-prev-case -position 1:2 -filename test.ml <test.ml
  ...
```

src/analysis/jump.ml
PizieDust: If we try to jump from the first case to a previous case it should return an exception saying there's no previous case.

voodoos: You asked for another review but this comment does not appear to be addressed, did you forgot to push some changes?

PizieDust: Thanks for catching that.

PizieDust: I think all commits are now pushed.

src/analysis/jump.ml
```diff
@@ -155,6 +188,10 @@ let get typed_tree pos target =
     else begin
       let nodes = skip_non_moving pos enclosings in
       let node = find_node preds nodes in
+      match target with 
+      | "match-next-case" -> find_case_pos (get_cases_from_match node) pos Next
+      | "match-prev-case" -> find_case_pos (List.rev (get_cases_from_match node)) pos Prev
```
voodoos, line 193: Nice trick 😃
I was puzzled for a moment as to why the comparison was working and then I saw that `List.rev`.

src/analysis/jump.ml
```diff
@@ -127,19 +131,48 @@ let rec skip_non_moving pos = function
...
+        let check = 
+          match direction with 
+          | Prev -> pos.Lexing.pos_cnum > pat_loc.loc_start.pos_cnum
+          | Next -> pos.Lexing.pos_cnum < pat_loc.loc_start.pos_cnum
+        in
+        if check then
+          Some pat_loc.loc_start
+        else 
+          find_pos pos tail direction
+    in
```
voodoos, line 153: Indentation is wrong here

src/analysis/jump.ml
```diff
@@ -127,19 +131,48 @@ let rec skip_non_moving pos = function
...
+    let case = find_pos pos cases direction in 
+    (match case with
```
voodoos, line 155: This parenthesis is useless I think.

src/analysis/jump.ml
```diff
@@ -127,19 +131,48 @@ let rec skip_non_moving pos = function
...
+        else 
+          find_pos pos tail direction
+    in
+    let case = find_pos pos cases direction in 
+    (match case with 
+    | Some location -> `Found location
+    | None -> 
+      (match direction with 
+      | Next -> raise (No_next_match_case)
+      | Prev -> raise (No_prev_match_case)))
```
voodoos, line 160:
```suggestion
      | Next -> raise No_next_match_case
      | Prev -> raise No_prev_match_case))
```

src/analysis/jump.mli
```diff
@@ -31,7 +31,6 @@ val get :
   Mtyper.typedtree ->
   Std.Lexing.position ->
   string -> [> `Error of string | `Found of Lexing.position ]
-
```
voodoos, line 34: Unrelated refactoring

tests/test-dirs/motion/jump_match.t
```diff
@@ -0,0 +1,86 @@
...
+Test when there's no previous case
+  $ $MERLIN single jump -target match-prev-case -position 3:2 -filename test.ml < test.ml
+  {
+    "class": "return",
+    "value": "No previous case found",
+    "notifications": []
+  }
+
+Test jump from case 'O' to the previous case
+  $ $MERLIN single jump -target match-prev-case -position 9:2 -filename test.ml < test.ml
```
voodoos, line 70: We should always try to test as many "really" different cases as possible. Here we miss "go to next case from somewhere in the middle (like O)".

voodoos (approved):
I think it mostly looks good. We should have a look at how the editors call the jump command now :-)

src/analysis/jump.ml
PizieDust: Thank you 😃

src/analysis/jump.mli
```diff
@@ -31,7 +31,7 @@ val get :
...
+
```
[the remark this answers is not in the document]
voodoos: You marked the comment resolved but the extra whitespace appears to still be there ?

src/analysis/jump.ml
```diff
@@ -155,6 +190,10 @@ let get typed_tree pos target =
     else begin
       let nodes = skip_non_moving pos enclosings in
       let node = find_node preds nodes in
+      match target with 
+      | "match-next-case" -> find_case_pos (get_cases_from_match node) pos Next
+      | "match-prev-case" -> find_case_pos (List.rev (get_cases_from_match node)) pos Prev
+      | _ ->
       let node_loc = Browse_raw.node_real_loc Location.none node in
       `Found node_loc.Location.loc_start
```
voodoos, line 198: Here too the indentation is wrong.

voodoos (approved):
Thank again for this, do you want to rebase into meaningful commits before I merge ?

Also there is still some unrelated format changes (added white space), can you get rid of them ? It's very useful for reviewers and to keep the history clean to only unclude meaningful changes into PRs.

PizieDust:
> Thank again for this, do you want to rebase into meaningful commits before I merge ?
> 
> Also there is still some unrelated format changes (added white space), can you get rid of them ? It's very useful for reviewers and to keep the history clean to only unclude meaningful changes into PRs.

Thank you. I think I've fixed the whitespace and indentation issues in the newest commit.  Is it possible to rebase all into one commit and merge? I

src/analysis/jump.ml
PizieDust: I think it should all be good now

Issue #1656, opened 2023-07-24
Allow jump to an enclosing module type


goldfirere:
The current `jump` command allows jumps to an enclosing `fun`, `let`, `module`, or `match`. It would be great to also add support for module types, perhaps by adding the `sig` target.

voodoos:
Do you mean jumping to the enclosing `sig ...` from inside the signature ?

goldfirere:
Yes, exactly. How is jumping to a module binding the same?

voodoos:
There are definitely some cases that could be added to the `jump` command like `sig` and `struct`. It also lacks the module type binding case.

xvw:
Hi @goldfirere!
PR #1728 implements the behaviour proposed by @voodoos [here](https://github.com/ocaml/merlin/issues/1656#issuecomment-1650001766), is the fix enough or do you really want to jump to `sig` enclosing specifically?

goldfirere:
Hello @xvw! Welcome! Yes, that looks like the fix. Thanks.

PR #1728, opened 2024-02-19
Allow jump to module type

xvw:
This patch proposes to add the `module-type` parameter to the `jump` command.
Address partially #1656


voodoos (approved):
That was fast :-) 

The changes look good to me, could you add an entry to the changelog ?

xvw:
> That was fast :-)
> 
> The changes look good to me, could you add an entry to the changelog ?

Done.

PR #1734, opened 2024-02-22
destruct: destructing punned record field breaks syntax

### LINKED ISSUE #1661: Destructing punned record field breaks syntax
Consider the following code:

```
type t =
  { a : int * int
  ; b : string
  }

let f ({ a; b } : t) = assert false
```

With the cursor at record field `a` on the last line of the snippet, executing `destruct` gives:

```
let f ({ (_, _); b } : t) = assert false
```

whereas the expected result should be something like:

```
let f ({ a = (_, _); b } : t) = assert false
```

xvw:
Destructing a record punning field produced invalid OCaml syntax, as described in #1661, the case where we deconstruct on `a`, the `a` is substituted by the result of the deconstruction, whereas we would like to produce `a = destruct_result`. 

fix #1661

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@ merlin NEXT_VERSION
 
   + merlin binary
     - destruct: Removal of residual patterns (#1737, fixes #1560)
+    - Do not erase fields' names when destructing punned record fields (#1734, 
+      fixes #1661)
 
 merlin 4.14
 ===========
--- a/src/analysis/destruct.ml
+++ b/src/analysis/destruct.ml
@@ -481,6 +481,25 @@ let find_branch patterns sub =
   in
   aux [] patterns
 
+(* In the presence of record punning fields, the definition must be reconstructed
+   with the label. ie: [{a; b}] with destruction on [a] becomes *)
+(*  [{a = destruct_result; b}]. *)
+let find_field_name_for_punned_field patt = function
+  | Pattern {pat_desc = Tpat_record (fields, _); _} :: _ ->
+    List.find_opt ~f:(fun (_, _, opat) ->
+        let ppat_loc = patt.Typedtree.pat_loc
+        and opat_loc = opat.Typedtree.pat_loc in
+        Int.equal (Location_aux.compare ppat_loc opat_loc) 0
+      ) fields |> Option.map ~f:(fun (_, label, _) -> label)
+  | _ -> None
+
+let print_pretty ?punned_field config source subject =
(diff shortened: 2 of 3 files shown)


src/analysis/destruct.ml
```diff
@@ -571,9 +590,9 @@ let rec node config source selected_node parents =
             | [ more_precise ] ->
               (* If only one pattern is generated, then we're only refining the
                 current pattern, not generating new branches. *)
+              let punning_field = find_field_name_for_punning_field patt parents in
               let ppat = filter_pat_attr (Untypeast.untype_pattern more_precise) in
-              let str = Mreader.print_pretty
-                  config source (Pretty_pattern ppat) in
+              let str = print_pretty ?punning_field config source (Pretty_pattern ppat) in
               patt.Typedtree.pat_loc, str
             | sub_patterns ->
```
voodoos, line 597: Does the issue happens only when there is only one pattern generated ?

It might be worth adding a test case for when there are several, maybe:

```
 $ $MERLIN single case-analysis -start 2:9 -end 2:9 \
  > -filename main.ml <<EOF
  > type t = {a: int option; b: string}
  > let f ({a; b} : t) = assert false
  > EOF
```

xvw: I have add your test and it seems that this the right behaviour. WDYT?

voodoos: I think the result is what I was expecting :-)

src/analysis/destruct.ml
```diff
@@ -571,9 +590,9 @@ let rec node config source selected_node parents =
             | [ more_precise ] ->
               (* If only one pattern is generated, then we're only refining the
                 current pattern, not generating new branches. *)
+              let punning_field = find_field_name_for_punning_field patt parents in
```
voodoos, line 593: I think I would prefer something like `punned_field` here, but that's not really important.

voodoos (approved):
Thanks @xvw  the changes looks good to me.

CHANGES.md
```diff
@@ -3,6 +3,8 @@ merlin NEXT_VERSION
 
   + merlin binary
     - destruct: Removal of residual patterns (#1737, fixes #1560)
+    - Reconstruction of the record field in the event of destruction on a record punning field
+      (#1734, fixes #1661)
```
voodoos, line 7:
```suggestion
    - Do not erase fields' names when destructing punned record fields (#1734, 
      fixes #1661)
```

PR #1737, opened 2024-02-23
destruct: Removal of residual patterns

xvw:
The fix is actually ... very simple as pattern generation is assumed to be this way:

```ocaml
type t = A of int | B | C
let f = function A x -> x | _ -> 2 (* on _ we want to produces B | C *)
let g = function A x -> x | _ -> 1 | C -> 2 (* I assume that we want only generate B *)
```

So the idea is just to build the list of pattern in this order `after, before, new_branches`.
Understanding the problem was a bit complex (for such a small fix) as the names rev_before and after were confusing (Perhaps because the logic of pattern matching has changed significantly since the original implementation?). I can probably take advantage of the branch to do some renaming.


voodoos (approved):
The test looks nice, that's a useful fix :-)

PR #1743, opened 2024-03-29
add get/set contexts commands

jchavarri:
Implements new additions to Dune protocol. The goal is to allow users to dynamically choose the Dune context to use for merlin information from their editor, rather than having to edit `dune-workspace` file. See https://github.com/ocamllabs/vscode-ocaml-platform/issues/1432 for added context.

Depends on https://github.com/ocaml/dune/pull/10324.


jchavarri:
This won't be necessary with the most recent approach, see https://github.com/ocamllabs/vscode-ocaml-platform/issues/1432#issuecomment-2034676788.

PR #1745, opened 2024-04-02
[New Command] Expand PPX nodes

### LINKED ISSUE #1744: Implement Enhanced Hover Feature for PPXs in Merlin
The objective of this issue is to enhance the hover feature in Merlin to provide context-specific information for PPX annotations (derivers and extension nodes). 
Currently, the hover feature in `ocaml-lsp` provides information based on the `type-enclosing` command in Merlin, which is not context-specific. By implementing this enhancement, users will receive relevant information when hovering over PPX annotations, particularly what the PPX output should be.

PizieDust:
closes #1744 
This implementation has support for:
- Ppx derivers written using ppxlib
- Ppx derivers written using ppx_deriving
- Ppx rewriters


src/frontend/ocamlmerlin/query_json.ml
```diff
@@ -392,6 +394,10 @@ let json_of_response (type a) (query : a t) (response : a) : json =
         ("url", `String info.documentation);
       ]
     | `No_documentation -> `String "No documentation found")
+  | Expand_node _, resp -> 
+    (match resp with
+    | `Found nodes -> `String nodes
+    | `No_deriver -> `String "No PPX deriver/extension node found on this position")
```
xvw, line 400: A very small nitpick that do not require parenthesis:

```ocaml
| Expand_node _, resp -> 
   let str = match resp with 
   | `Found nodes -> nodes
   | `No_deriver -> "No PPX deriver/extension node found on this position"
   in `String str
```

src/frontend/query_commands.ml
```diff
@@ -512,6 +512,29 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
     | Some res -> `Found res 
     | None -> `No_documentation)
 
+  | Expand_node pos -> 
+    let pos = Mpipeline.get_lexing_pos pipeline pos in
+    let typer = Mpipeline.typer_result pipeline in 
+    let node = Mtyper.node_at typer pos in
+    let check_ppx (node: Env.t * Mbrowse.node) : bool =
```
xvw, line 519: Is the type anotation really mandatory? 
`let check_ppx node = ...` seems sufficient no?

src/frontend/query_protocol.ml
```diff
@@ -145,6 +145,11 @@ type _ t =
     -> [ `Found of syntax_doc_result
        | `No_documentation
        ] t
+  | Expand_node
+  : Msource.position
+  -> [ `Found of string
+    | `No_deriver
+    ] t
```
xvw, line 152: Formatting seems a little bit weird here.

src/kernel/mbrowse.ml
```diff
@@ -221,6 +221,22 @@ let of_signature sg =
   in
   [env, Browse_raw.Signature sg]
 
+let rec of_structure_items lst =
+  match lst with
+  | [] -> []
+  | (env,head) :: tl -> 
+    Browse_raw.Structure_item (head,env) :: (of_structure_items tl)
```
xvw, line 228: I think that the function could be tail-recursive?  (but in fact it seems that it is `map` and it will probably do not require a `List.rev` on the previous file?): 

```ocaml
let of_structure_items = 
  List.map (fun (head, env) -> 
       Browse_raw.structure_item (head, env)) 
```

src/kernel/mbrowse.ml
```diff
@@ -221,6 +221,22 @@ let of_signature sg =
...
+    
+let rec of_signature_items lst =
+  match lst with
+  | [] -> []
+  | (env,head) :: tl -> 
+    Browse_raw.Signature_item (head,env) :: (of_signature_items tl)
+
```
xvw, line 235: Same here. 
But if you want to keep the list reversed. I suggest (and also for `of_structure_item`): 

```ocaml
let of_structure_item = 
  let rec aux acc = function 
  | [] -> acc
  | (env, head) :: tl -> 
     aux (Browse_raw.Signature_item (head, env) :: acc)
  in aux []
```

src/kernel/mbrowse.ml
```diff
@@ -261,3 +277,39 @@ let print_node () node =
...
+  let children = 
+    match root with
+    | (_,Structure str) -> 
+        of_structure_items (List.filter ~f:(fun (_,x) ->
+            check_node pos (Structure_item (x, str.str_final_env))
+        ) (List.map ~f:(fun x -> (str.str_final_env, x)) str.str_items))
+    | (_,Signature str) -> 
+      of_signature_items (List.filter ~f:(fun (_,x) ->
+        check_node pos (Signature_item (x, str.sig_final_env))
+      ) (List.map ~f:(fun x -> (str.sig_final_env, x)) str.sig_items))
```
xvw, line 296: I guess that we can simplify `get_children` using `filter_map` or moving some logic in `of_(signature|structure)_items`. WDYT?

src/kernel/mbrowse.ml
```diff
@@ -261,3 +277,39 @@ let print_node () node =
...
+      | Browse_raw.Signature_item (n,_) ->
+        Pprintast.signature_item ppf
+        (Untypeast.untype_signature_item n)
+      | _ -> raise (Invalid_argument "Wrong nodes")
+    end;
+  Format.pp_print_newline ppf ();
+  to_string ()
+
+let pprint_deriver_nodes () nodes = 
+  List.print (fun () node -> pprint_deriver_node () node) () nodes
```
xvw, line 315: Should works:

```ocaml
let pprint_deriver_nodes = List.print pprint_deriver_node
```

src/frontend/query_commands.ml
```diff
@@ -512,6 +512,29 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+      let attrs = Browse_raw.node_attributes n in
+      List.exists ~f:(fun (a:Typedtree.attribute) ->
+        let loc = a.attr_loc in
+            loc.loc_start.pos_cnum <= pos.pos_cnum
+            && loc.loc_end.pos_cnum >= pos.pos_cnum
+      ) attrs && has_a_deriver
+    in 
+    let has_deriver = List.exists ~f:check_ppx node in
+    if has_deriver then 
+      let root = (List.hd (List.rev node)) in
```
pitag-ha, line 532: You're going all the way to the top-level structure/signature here. Instead, you'll want to take the structure/signature that the structure/signature item belongs to that the deriver is attached to. Let me explain with an example:
```
module M = struct
    module N = struct
       type t = int [@@deriving foo]
       let a = 1
    end
    let b = 2
end
let c = 3
```

In that example, `root` will contain the module binding of `M` and the value binding of `c`. And you'll retrieve all items from `root` that contain the cursor position, i.e. you'll retreive the whole module binding of `M`. Instead, you'll want `root` to be the structure of the module expression of `N`, i.e. the type declaration of `t`, the derived structure items, and the value binding of `a`. That way, when you retrieve all items from `root` that contain the cursor position, you'll get the type definition of `t` and the derived code from it.

Let me know if it's clear what I mean!

src/frontend/query_commands.ml
```diff
@@ -512,6 +512,29 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+      let derived_nodes = Mbrowse.get_children pos root in
```
pitag-ha, line 533: You're using the cursor position here to decide whether an item is a derived item or not. Concretely, your logic is: Check whether the cursor position is contained in the item or not. If so, consider it a derived item (or the original item). If not, not.

When a deriver is defined with `ppxlib` that works! It works for the following reason: `ppxlib` always wraps all derived items into an
```
include struct <derived items> end <some attributes for odoc and merlin>
```
and it gives that `include` structure item the location of the whole item the deriver is attached to. So the `include` item will always contain the cursor position, when the cursor is somewhere on the deriving annotation and will therefore be detected correctly.

However, `ppx_deriving` doesn't wrap the derived items into such an `include`. So for a deriver registered via `ppx_deriving`, the situation is a bit strange. Example: If the deriver author has given the location of the deriver's name to the derived item, then the derived item will be detected as derived if the cursor position is on the deriver's name, whereas a cursor location on the `[@@deriving` won't detect it. Or worse: If the author has given the derived item the location of the type the deriver is attached to, then the derived item will never be detected.

The problem here: I don't know if we want to add `ppx_deriving` as a test dependency to `merlin-lib`. If not, we can't test the `ppx_deriving` behavior. So in that case, it might be better to add a note to the feature's doc saying that it works reliably for `ppxlib` derivers, but not for `ppx_deriving` derivers. @voodoos , what do you think about this?

src/kernel/mbrowse.ml
```diff
@@ -261,3 +277,39 @@ let print_node () node =
...
+    match root with
+    | (_,Structure str) -> 
+        of_structure_items (List.filter ~f:(fun (_,x) ->
+            check_node pos (Structure_item (x, str.str_final_env))
+        ) (List.map ~f:(fun x -> (str.str_final_env, x)) str.str_items))
+    | (_,Signature str) -> 
+      of_signature_items (List.filter ~f:(fun (_,x) ->
+        check_node pos (Signature_item (x, str.sig_final_env))
+      ) (List.map ~f:(fun x -> (str.sig_final_env, x)) str.sig_items))
+    | _ -> raise (Invalid_argument "Not a valid root node")
```
pitag-ha, line 297: It's a strongly good practice to avoid raising. The reason: If this case is hit, then the user will see a merlin answer along the lines of "exception: Invalid_argument "Not a valid root node". That's not very helpful. (side-note: Luckily merlin catches exceptions at the end, so at least the server doesn't crash, and wraps the exception into an answer instead).

Instead, try to use the OCaml type system. Why don't you directly pass a structure or a signature to `get_children`? If you do so, that case can never be hit, and instead, we're protected by the type system to pass something to the function that it doesn't know to handle.

pitag-ha (commented):
This looks very nice and is a very useful feature. Thanks a lot, @PizieDust! two general comments:

As a suggestion for an addition to the tests: It would be interesting to test what happens when you use the feature on a deriver that's not registered via `dune`. Or when you use it on a deriver that triggers some other error, e.g. you call it with a `flag`, when it doesn't support flags, or it expects some value to be in scope that's not in scope. In those cases, do the errors appear as "derived code" or not?

And as a nit-pick: `merlin-lib` doesn't know about hover, so the title of the PR is a little misleading. The part of your work about hover is the related work you've done on ocaml-lsp.

src/frontend/query_commands.ml
PizieDust: Yes the issue with `ppx_deriving` derivers as a whole is the fact that Merlin (currently) doesn't know how to traverse them, especially given that most don't respect location invariants [ppxlib location invariants](https://ocaml.org/p/ppxlib/latest/doc/good-practices.html#the-invariants) and so this functionality completely falls apart when used with any deriver written with `ppx_deriving`.

src/frontend/query_commands.ml
PizieDust: Now, it will only return the exact ppx node which has been added. I am adding a conditional `loc_ghost = true` to make sure we get only the `include ...` part of it.

src/commands/query_json.ml
```diff
@@ -392,6 +394,19 @@ let json_of_response (type a) (query : a t) (response : a) : json =
...
+    | `Found ppx_info -> 
+      `Assoc
+      [
+        ("code", `String ppx_info.code);
+        ("deriver", `Assoc [
+          ("start", Lexing.json_of_position ppx_info.deriver.a_start);
+          ("end", Lexing.json_of_position ppx_info.deriver.a_end);
+        ])
+      ]
+    | `No_deriver -> `String "No PPX deriver/extension node found on this position"
```
voodoos, line 408: ``` `No_deriver``` is not a good name if it means "no deriver or extension node"

src/frontend/query_protocol.ml
```diff
@@ -103,6 +103,18 @@ type syntax_doc_result =
...
+type ppx_deriver_pos =
+{ 
+  a_start : Lexing.position;
+  a_end : Lexing.position;
+}
+
+type ppx_expand_result = 
+{
+  code : string;
+  deriver : ppx_deriver_pos
```
voodoos, line 115: Do you use the `ppx_deriver_pos` anywhere else ? If not you can just inline it. Also this does not only concern `derivers` anymore isn't it ?

Since these are actually the position of the attribute, you could call it `attr_start` / `attr_end` or something like that.

src/ocaml/merlin_specific/browse_raw.ml
```diff
@@ -671,6 +674,8 @@ let of_node = function
   | Type_extension { tyext_params; tyext_constructors } ->
     list_fold of_typ_param tyext_params **
     list_fold (fun ec -> app (Extension_constructor ec)) tyext_constructors
+  | Type_exception {tyexn_constructor} ->
```
voodoos, line 677:
```suggestion
  | Type_exception { tyexn_constructor } ->
```

(this should go in a separate pr anyway)

tests/test-dirs/expand_node/ppx-tests.t/rewriter/dune
```diff
@@ -0,0 +1,4 @@
+(library
+ (name my_ppx)
+ (kind ppx_rewriter)
+ (libraries ppxlib))
```
voodoos, line 4: Be careful to always have a newline at the end of the files. It is an old tradition and that's the reason why github is displaying this ugly ⛔️ sign. You can configure your editor to always add a last line.

(also happens in other files :-) )

tests/test-dirs/outline.t/run.t
```diff
@@ -110,21 +110,6 @@
...
-        "end": {
-          "line": 18,
-          "col": 20
-        },
-        "name": "Ex",
-        "kind": "Exn",
-        "type": null,
-        "children": [],
-        "deprecated": false
-      },
```
voodoos, line 127: Looks like this is related to your Type_exception changes, so we can discuss it in the dedicated PR, but are you sure this diff is not a regression ?

src/ocaml/typing/untypeast.mli
```diff
@@ -81,6 +81,8 @@ val default_mapper : mapper
 
 val untype_structure : ?mapper:mapper -> Typedtree.structure -> structure
 val untype_signature : ?mapper:mapper -> Typedtree.signature -> signature
+val untype_structure_item : ?mapper:mapper -> Typedtree.structure_item -> structure_item
+val untype_signature_item : ?mapper:mapper -> Typedtree.signature_item -> signature_item
```
voodoos, line 85: These changes are not used anymore, you can revert them.

src/frontend/query_commands.ml
```diff
@@ -511,6 +511,164 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+            Pprintast.expression Format.str_formatter exp)
+            (List.rev expression);
+            Format.flush_str_formatter ()
+        in
+        ppx_expansion 
+        ~ppx:exp 
+        ~a_start:(Option.get !atr).loc_start
+        ~a_end:(Option.get !atr).loc_end
+      | _ -> `No_deriver
+    end
```
voodoos, line 670: I think this is long and complex enough to be moved in a specific module in `src/analysis/`

tests/test-dirs/expand_node/ppx-tests.t/run.t
```diff
@@ -0,0 +1,732 @@
+Dune setup
```
voodoos, line 1: You added a lot of tests to show that the feature only triggers on attributes, maybe a bit too much ? It makes the test very inconvenient to read.

I think it is less important than for the "syntax doc" query because here we don't rely on fragile "node matching" but on the iterator and the exact attribute nodes are found. It's nice to have a couple of cases showing that it doesn't over-trigger, but dozens is a bit too much.

src/frontend/query_commands.ml
```diff
@@ -511,6 +511,164 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+      match !signature,!structure,!expression with
+      | signature,[],[] when signature <> [] ->
+        let exp =
+          Pprintast.signature Format.str_formatter (List.rev signature);
+          Format.flush_str_formatter ()
+        in
+        ppx_expansion 
+        ~ppx:exp 
+        ~a_start:(Option.get !atr).loc_start
+        ~a_end:(Option.get !atr).loc_end
```
voodoos, line 648: These should be indented.

src/frontend/query_commands.ml
```diff
@@ -511,6 +511,164 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+        let iterator = 
+          {Ast_iterator.default_iterator with signature_item}
+        in 
+        (match ppxed_parsetree with
+        | `Interface si -> iterator.signature iterator si
+        | `Implementation str -> iterator.structure iterator str)
+      | false -> Ast_iterator.default_iterator.signature_item self item_1
+    in
+    let structure_item (self : Ast_iterator.iterator) 
+    (item_1 : Parsetree.structure_item) =
```
voodoos, line 611: There is something wrong with the formatting / indentation here.

src/frontend/query_commands.ml
```diff
@@ -511,6 +511,164 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+      (match check_structures item_1.pstr_desc with
```
voodoos, line 612: I don't think this parenthesis is needed

src/frontend/query_commands.ml
```diff
@@ -511,6 +511,164 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a =
...
+      | signature,[],[] when signature <> [] ->
+        let exp =
+          Pprintast.signature Format.str_formatter (List.rev signature);
+          Format.flush_str_formatter ()
+        in
+        ppx_expansion 
+        ~ppx:exp 
+        ~a_start:(Option.get !atr).loc_start
+        ~a_end:(Option.get !atr).loc_end
+      | [],structure,[] when structure <> [] ->
```
voodoos, line 649: You could put a `| [], [], [] -> ...` case first so that you don't need the guarding clauses `when ...`

voodoos (commented):
I dit a round of review. I did not tackle the heart of the PR yet

src/analysis/ppx_expand.ml
```diff
@@ -0,0 +1,134 @@
...
+        let iterator = { Ast_iterator.default_iterator with structure_item } in
+        match ppxed_parsetree with
+        | `Interface si -> iterator.signature iterator si
+        | `Implementation str -> iterator.structure iterator str)
+    | false -> Ast_iterator.default_iterator.structure_item self item_1
+  in
+  let iterator =
+    { Ast_iterator.default_iterator with signature_item; structure_item; expr }
+  in
+  let _ =
```
voodoos, line 101:
```suggestion
  let () =
```

src/analysis/ppx_expand.ml
```diff
@@ -0,0 +1,134 @@
...
+    match parsetree with
+    | `Interface si -> iterator.signature iterator si
+    | `Implementation str -> iterator.structure iterator str
+  in
+  match (!signature, !structure, !expression) with
```
voodoos, line 106: This matching makes me sad :-) 
Lets see if other refactoring might lead us to a nicer code organization...

src/analysis/ppx_expand.ml
```diff
@@ -0,0 +1,134 @@
...
+        let structure_item (self : Ast_iterator.iterator)
+            (item_2 : Parsetree.structure_item) =
+          match check_at_pos item_2.pstr_loc && item_1 <> item_2 with
+          | true -> structure := item_2 :: !structure
+          | false -> Ast_iterator.default_iterator.structure_item self item_2
+        in
+        let iterator = { Ast_iterator.default_iterator with structure_item } in
+        match ppxed_parsetree with
+        | `Interface si -> iterator.signature iterator si
+        | `Implementation str -> iterator.structure iterator str)
```
voodoos, line 95: These nested iterators make the logic unclear. 

Would it be possible to separate the two steps of the process in a more explicit way ?

First step: check if the pos is a ppx attribute.
Second step: get all the items with that location.


The result of the first step could be stored in a variant type such as: 
`type kind = Expr of Parsetree.expression | Sig_item of Parsetree.signature_item | Str_item ...`

Then you match on this, apply the correct iterator to gather results and return them. 
This way you can make the accumulating ref specific to each case.

src/frontend/query_protocol.ml
```diff
@@ -103,6 +103,13 @@ type syntax_doc_result =
     documentation : string
 }
 
+type ppx_expand_result = 
+{
+  code : string;
```
voodoos, line 108: Sorry, that really neatpicky, but I would prefer another name like `source` or `ppxed_source`, code is a bit vague.

src/analysis/ppx_expand.ml
```diff
@@ -0,0 +1,163 @@
...
+type ppx_kind =
+  | Expr of Parsetree.expression
+  | Sig_item of Parsetree.signature_item
+  | Str_item of Parsetree.structure_item
+
+let attr = ref None
+let expression = ref []
+let signature = ref []
+let structure = ref []
+let kind = ref None
```
voodoos, line 12: You're refactoring led to move these reference at the global scope which is something we really don't want to do :-)

I think what I meant was:

1. Have a first iterator which detects the kind of ppx extension we are on: 
```
check_extension : ~parsetree -> pos -> ppx_kind
```
This function can use a reference or a locally defined exception to extract the info from the iterator.

2. And then another function which will have the local references such as `signature` and `structure` but doesn't expose them: 
```
get_ppxed_source : ~ppxed_parsetree pos ppx_kind -> [`Signature of Parsetree.signature | `Structure of Parsetree.structure```

That way the references you're using, and that are implementation details used to extract information from the iterators will be as constrained as possible.


If it's not clear, we can have a call about it !

File: src/commands/new_commands.ml [hunk 28]
```diff
@@ -242,6 +242,21 @@ Otherwise, Merlin looks for the documentation for the entity under the cursor (a
     end
   ;
 
+  command "expand-ppx"
+    ~doc: "Returns the generated code of a PPX."
+    ~spec: [
+      arg "-position" "<position> Position to complete"
```
voodoos, line 248:
```suggestion
      arg "-position" "<position> Position to expand"
```

src/frontend/query_protocol.ml
```diff
@@ -152,6 +159,12 @@ type _ t =
     -> [ `Found of syntax_doc_result
        | `No_documentation
        ] t
+  | Expand_ppx
+    : Msource.position
+    -> 
+      [ `Found of ppx_expand_result
+      | `No_ppx
+      ] t
```
voodoos, line 167:
```suggestion
  | Expand_ppx
    : Msource.position
    -> [ `Found of ppx_expand_result
       | `No_ppx
       ] t
```

src/analysis/ppx_expand.ml
PizieDust: This was a great review. Thanks. 
I've pushed commits with the changes. 

In other to have correct attribute locations for highlighting purposes in VSCode, `check_extension` returns the `ppx_kind * Warnings.loc` and `get_ppxed_source` takes in the `ppx_kind * Warnings.loc`

src/analysis/ppx_expand.ml
```diff
@@ -0,0 +1,155 @@
...
+  let () =
+    match ppx_kind_with_attr with
+    | Expr original_expr, _ -> (
+        let expr (self : Ast_iterator.iterator)
+            (new_expr : Parsetree.expression) =
+          match
+            original_expr.pexp_loc.loc_start.pos_cnum
+            <= new_expr.pexp_loc.loc_start.pos_cnum
+            && new_expr.pexp_loc.loc_end.pos_cnum
+               <= original_expr.pexp_loc.loc_end.pos_cnum
```
voodoos, line 93: Are you sure that works with multi-line expressions ?
You probably want to use `Location_aux.included`

src/analysis/ppx_expand.ml
```diff
@@ -0,0 +1,155 @@
+type ppx_kind =
+  | Expr of Parsetree.expression
+  | Sig_item of Parsetree.signature_item
+  | Str_item of Parsetree.structure_item
```
voodoos, line 4: In the LSP implementation they handle only two cases `Ppx_expr` and `Ppx_typedef_attr`. Could you explain to me the difference ?

voodoos:
Thanks, that looks much better !

src/analysis/ppx_expand.ml
PizieDust: So in the LSP implementation, `Ppx_expr` is to handle PPX extension nodes `[%foo]` and `Ppx_typedef_attr` is for PPX derivers on `type_declarations`. So this was a bit shallow to me as it will only work for `Type declarations`, but we need the expand feature to work on all PPX nodes. and this includes:
- type declaration in structure and signature
- type extension in structure and signature
- exception in structure and signature
- module type declaration in structure and signature

So this PR handles all the different nodes, and not just `type_declarations`

src/analysis/ppx_expand.ml
PizieDust: This is a very cool function.

src/analysis/ppx_expand.ml
PizieDust: I'll update the `ocaml-lsp` PR to handle all these nodes as well because currently it will only trigger `ppx-expand` on Merlin if it's a `type_declaration` or `expression`.

tests/test-dirs/expand_node/ppx-tests.t/run.t
```diff
@@ -0,0 +1,587 @@
...
+  >   val push : t -> stack -> stack 
+  >   val pop : stack -> stack
+  >   val peek : stack -> t
+  > end [@@deriving rename]
+  > EOF
+
+  $ dune build
+
+on attribute name deriving of type t
+TODO: a cursor here should only output the derived t and now the whole module
```
voodoos, line 384: What about this todo, is there something we can do ? Was this behavior already the one of the current LSP implementation ?

PizieDust: The current LSP implementation will not give an output as it only works on `type_declarations`.

src/analysis/ppx_expand.ml
```diff
@@ -99,9 +99,13 @@ let get_ppxed_source ~ppxed_parsetree ~pos ppx_kind_with_attr :
...
         let signature_item (self : Ast_iterator.iterator)
             (new_sg : Parsetree.signature_item) =
-          match check_at_pos pos new_sg.psig_loc && original_sg <> new_sg with
-          | true -> signature := new_sg :: !signature
-          | false -> Ast_iterator.default_iterator.signature_item self new_sg
+          let included =
+            Location_aux.included new_sg.psig_loc ~into:original_sg.psig_loc
+          in
+          match included && original_sg <> new_sg, new_sg.psig_loc.loc_ghost with
+          | true, _ -> signature := new_sg :: !signature
```
PizieDust, line 106: This works great for `signature_items` but for `structure_items` something weird happens, the result looks like this:
```
;;rename
 include
    struct
      let _ = fun (_ : point) -> ()
      type point_renamed = {
        x: int ;
        y: int }
    end[@@ocaml.doc "@inline"][@@merlin.hide ]
```
There is an extra `;;rename` added which can be removed if we enforce `ghost_locs` for the `structure_items`

tests/test-dirs/expand_node/ppx-tests.t/run.t
```diff
@@ -0,0 +1,579 @@
...
+    "value": {
+      "code": "include sig [@@@ocaml.warning \"-32\"] type t_renamed end[@@ocaml.doc
+                                                           \"@inline\"][@@merlin.hide
+                                                                      ]
+  include sig [@@@ocaml.warning \"-32\"] type t_renamed end[@@ocaml.doc
+                                                           \"@inline\"][@@merlin.hide
+                                                                      ]
+  include sig [@@@ocaml.warning \"-32\"] type t_renamed end[@@ocaml.doc
+                                                           \"@inline\"][@@merlin.hide
+                                                                      ]",
```
voodoos, line 338: Are the duplications in this test and the following one expected ?

PizieDust: Yes, but it's unrelated to Merlin or the functionality itself. The problem is with the example ppx I have written. Somewhere in the implementation for module type declarations, it takes the original type, and while processing, processes the ppx again for that type.

voodoos: Right, that seems like a non-trivial fix...

Maybe we should just rely on additional dependencies for testing, would `ppx_deriving` allows you to simplify the tests and use "real" ppxes ?

PizieDust: yes we could, but it may be difficult to find a ppx that acts on nodes like `module_type_declarations` or `exceptions`

voodoos: Right, let's keep it like that then. It would be great if you could add a comment in the test explaining why there are duplicated results. Then I will do a last review before merging :-)

PizieDust: I just added the comment

voodoos (approved):
Thanks @PizieDust ! This looks like a great addition :-)

PR #1746, opened 2024-04-04
Ignore SIGPIPE in the Merlin server process

liam923:
Bringing in changes from [merlin-jst](https://github.com/janestreet/merlin-jst/pull/49).

From @catern:

The merlin server writes to pipes received from the merlin client process. If the read end of those pipes is closed, e.g. because the merlin client process was killed, then when the server writes to the pipe it will generate a SIGPIPE. This kills the Merlin server. Ignore the SIGPIPE instead of unnecessarily dying.

The fact that you get SIGPIPE when you write to a closed pipe is a classic Unix footgun, and this is the usual resolution. Another solution is to avoid pipes (in favor of e.g. socketpairs), but we can't do that because we write directly to the file descriptors received from the client.

### DIFF
--- a/src/frontend/ocamlmerlin/ocamlmerlin_server.ml
+++ b/src/frontend/ocamlmerlin/ocamlmerlin_server.ml
@@ -65,7 +65,7 @@ module Server = struct
       Logger.log ~section:"server" ~title:"cannot setup listener" ""
     | Some server ->
       (* If the client closes its connection, don't let it kill us with a SIGPIPE. *)
-      let (_ : Sys.signal_behavior) = Sys.signal Sys.sigpipe Sys.Signal_ignore in
+      if Sys.unix then ignore @@ Sys.signal Sys.sigpipe Sys.Signal_ignore;
       loop (File_id.get Sys.executable_name) server;
       Os_ipc.server_close server
 end
--- a/src/frontend/ocamlmerlin/ocamlmerlin_server.ml
+++ b/src/frontend/ocamlmerlin/ocamlmerlin_server.ml
@@ -64,6 +64,8 @@ module Server = struct
     | None ->
       Logger.log ~section:"server" ~title:"cannot setup listener" ""
     | Some server ->
+      (* If the client closes its connection, don't let it kill us with a SIGPIPE. *)
+      let (_ : Sys.signal_behavior) = Sys.signal Sys.sigpipe Sys.Signal_ignore in
       loop (File_id.get Sys.executable_name) server;
       Os_ipc.server_close server
 end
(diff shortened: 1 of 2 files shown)


src/frontend/ocamlmerlin/ocamlmerlin_server.ml
```diff
@@ -64,6 +64,8 @@ module Server = struct
     | None ->
       Logger.log ~section:"server" ~title:"cannot setup listener" ""
     | Some server ->
+      (* If the client closes its connection, don't let it kill us with a SIGPIPE. *)
+      let (_ : Sys.signal_behavior) = Sys.signal Sys.sigpipe Sys.Signal_ignore in
```
nojb, line 68: This should be guarded under `Sys.unix` (no signals on Win32).

src/frontend/ocamlmerlin/ocamlmerlin_server.ml
```diff
@@ -64,6 +64,8 @@ module Server = struct
     | None ->
       Logger.log ~section:"server" ~title:"cannot setup listener" ""
     | Some server ->
+      (* If the client closes its connection, don't let it kill us with a SIGPIPE. *)
+      if Sys.unix then ignore @@ Sys.signal Sys.sigpipe Sys.Signal_ignore;
```
voodoos, line 68: I think that you can use `set_signal` here to remove the need for ignore.

voodoos (commented):
The changes look reasonnable to me.

Just a few thougts to understand better why this change is useful:

Merlin's "client" processes are usually short lived and not subject to frequent crashes. (Crashes due to issues with analysis crash the server, not the client.)

So I guess this is most impactful when Merlin gets stuck into a slow computation and the user kills the client process to bring back control to the editor  ? (This doesn't stop the server from finishing the work, and then crash due to sigpipe.)

Am I correct ? Is there another sequence of events that lead to these undesirable sigpipes ?

catern:
>So I guess this is most impactful when Merlin gets stuck into a slow computation and the user kills the client process to bring back control to the editor ? (This doesn't stop the server from finishing the work, and then crash due to sigpipe.)

Yes, that's exactly right.

BTW, since killing the client process doesn't actually cancel the operation, subsequent merlin clients still have to wait for the first operation to complete.  So control can be returned to the user, but subsequent commands may still have slow results.  It would be nice if killing the client actually also cancelled the operation.

src/frontend/ocamlmerlin/ocamlmerlin_server.ml
voodoos: @liam923 gentle ping :-)

liam923: Sorry about that, just fixed it

voodoos: Thanks !

voodoos (approved):

voodoos:
Thanks @catern !

PR #1749, opened 2024-04-11
Adopt OCaml's code of conduct

xvw:
As mentioned, adopting the OCaml code of conduct seems to be a good practice!
Who should be listed as "moderators"? (cc @voodoos, @let-def, @trefis, @PizieDust)?


voodoos:
You can add me !

PizieDust:
I'll be happy to participate in moderating as well.

Issue #1748, opened 2024-04-09
Jump to incorrect module


liam923:
Hi, I've recently joined the Jane Street compiler team. We've had a user report a bug, which I've included reproduction steps for below. Essentially, there is an `import.ml` file, which includes another module named `Import`. This leads to a `Import.Import` module existing in scope. When another file contains `open Import`, Merlin traces the location of `Import` to the correct location when the file is a `.ml` file but incorrectly traces to `Import.Import` when the file is a `.mli file`.

### Reproduction Steps

1. Download and extract [jump_to_wrong_import.tar.gz](https://github.com/ocaml/merlin/files/14923287/jump_to_wrong_import.tar.gz)
2. `cd` into the `jump_to_wrong_import` directory
3. `dune build`
4. `ocamlmerlin single locate -position 1:10 -look-for implementation -filename lib/foo.mli < lib/foo.mli`

This gives:
```
{"class":"return","value":{"file":"/path/to/jump_to_wrong_import/lib/bar.ml","pos":{"line":1,"col":0}},"notifications":[],"timing":{"clock":20,"cpu":4,"query":2,"pp":0,"reader":0,"ppx":0,"typer":2,"error":0},"heap_mbytes":1,"cache":{"reader_phase":"miss","ppx_phase":"miss","typer":"miss","cmt":{"hit":1,"miss":3},"cmi":{"hit":0,"miss":4}},"query_num":0}
```

Compare this to `ocamlmerlin single locate -position 1:10 -look-for implementation -filename lib/foo.ml < lib/foo.ml`, which gives the expected:
```
{"class":"return","value":{"file":"/path/to/jump_to_wrong_import/lib/import.ml","pos":{"line":1,"col":0}},"notifications":[],"timing":{"clock":18,"cpu":4,"query":2,"pp":0,"reader":0,"ppx":0,"typer":2,"error":0},"heap_mbytes":1,"cache":{"reader_phase":"miss","ppx_phase":"miss","typer":"miss","cmt":{"hit":1,"miss":2},"cmi":{"hit":0,"miss":4}},"query_num":0}
```

#### Version info:
- `ocamlmerlin` built from [d5dd40e70eceabe7f9c7c413141430e57e44ded4](https://github.com/ocaml/merlin/commit/d5dd40e70eceabe7f9c7c413141430e57e44ded4) (up-to-date master as of the time of posting this)
- `dune` 3.14.2
- `ocaml` 4.14.2

Full opam env:
```
[...]

PR #1750, opened 2024-04-12
test: reproduce issue #1748

voodoos:


voodoos:
It looks like Merlin is not performing the lookup in the correct environment:
```
  # 0.01 locate - from_string
  looking for the source of 'Import' (prioritizing .ml files)
  # 0.01 locate - lookup
  lookup in module namespace
  # 0.01 locate - env_lookup
  found: 'Import!.Import' in namespace module with uid Bar.1
  ```

liam923:
This is great, it seems to fix the bug I originally reported. Thanks for the quick turnaround.

I did find a new issue that it causes, however. Once some wrapping modules are added, jumping from an implementation now causes a `Not in environment` error. See the test I attached below that demonstrates this:

[homonyms.t.tar.gz](https://github.com/ocaml/merlin/files/14962655/homonyms.t.tar.gz)

voodoos:
> This is great, it seems to fix the bug I originally reported. Thanks for the quick turnaround.
> 
> I did find a new issue that it causes, however. Once some wrapping modules are added, jumping from an implementation now causes a `Not in environment` error. See the test I attached below that demonstrates this:
> 
> [homonyms.t.tar.gz](https://github.com/ocaml/merlin/files/14962655/homonyms.t.tar.gz)

Right, I think we will need a different approach. I use an empty env which explains that error.

PR #1756, opened 2024-05-02
Improve context detection for infix operators

voodoos:
Context detection for longidents relies on some britle "where is the cursor in the lid" logic.
This PR add a special case to improve the handling of operators that requires parenthesis.
More precise locations in lid is required to actually get a clean and correct implementation, but at least this PR improves the situation a bit.

### DIFF
--- a/src/analysis/context.ml
+++ b/src/analysis/context.ml
@@ -64,8 +64,17 @@ let cursor_on_longident_end
   match lid with
   | Longident.Lident _ -> true
   | _ ->
-    let end_offset = loc.loc_end.pos_cnum in
-    let cstr_name_size = String.length name in
+    let end_offset =
+      loc.loc_end.pos_cnum in
+    let cstr_name_size =
+      (* FIXME: this is britle, but lids don't have precise enough location
+         information to handle these cases correctly. *)
+      let name_lenght = String.length name in
+      if Pprintast.needs_parens name then
+        name_lenght + 2
+      else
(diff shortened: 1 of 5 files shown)


src/analysis/context.ml
```diff
@@ -64,8 +64,17 @@ let cursor_on_longident_end
   match lid with
   | Longident.Lident _ -> true
   | _ ->
-    let end_offset = loc.loc_end.pos_cnum in
-    let cstr_name_size = String.length name in
+    let end_offset =
+      loc.loc_end.pos_cnum in
+    let cstr_name_size =
+      (* FIXME: this is britle, but lids don't have precise enough location
+         information to handle these cases correctly. *)
```
xvw, line 71: I think it is `brittle`. But in fact, is that really brittle?

xvw (approved):

src/analysis/context.ml
voodoos: Well, the whole thing is a bit dodgy. Merlin reconstructs a string, that it then parses again, to try to guess on what component the cursor here since longident don't have precise enough location information. Mixed with weird things like ident that require parenthesis or others that require parenthesis plus spaces, it makes the whole process very hacky.

Hopefully having better location information in longidents will allow us to replace all of this soon :-)

xvw: Make sense. Thanks for the explanation!

PR #1757, opened 2024-05-02
Move branch master to 5.2 support

voodoos:
We kept master on OCaml 4.14 for an unusually long time due to the experimental nature of OCaml 5.0.
Branch `502` is the branch `502-preview` rebased onto master. 
Support for OCaml 4.14 will continue in it's own `414` branch.

PR #1758, opened 2024-05-05
Extract commands from new protocol inside a dedicated library

xvw:
**This PR is based on `voodoos/502` (https://github.com/ocaml/merlin/pull/1757)**

In order to easily enable tunneling, via a custom query from `ocaml-lsp`, a large part of the argument processing logic (argv) can be used. This branch separates the command invocation logic (and the dispatching of CLI arguments to useful `merlin-lib` functionality) into a dedicated library, which can be used from within `ocaml-lsp`.


src/commands/dune
```diff
@@ -0,0 +1,14 @@
+(library
+  (name merlin_commands)
+  (public_name merlin-lib-commands)
```
voodoos, line 3:
```
(public_name merlin-lib.commands)
```
If it does not actually introduce new dependencies, we might simply ship it as a part of `merlin-lib`

voodoos (commented):
Thanks @xvw !

It's nice to realize that `Query_json` actually doesn't depends on Yojson :-)

Did you also check that other important parsing function such as the one parsing the flags to build the configuration are already exposed by merlin lib ?

src/commands/dune
xvw: Done in the last commit https://github.com/ocaml/merlin/pull/1758/commits/24e97948e1d3f3774484f26e6c18338ac266d02a

voodoos (approved):
These changes LGTM.

Were you able to confirm that this is actually enough to implement the "tunneling" on the LSP side ?

xvw:
> Thanks @xvw !
> 
> It's nice to realize that `Query_json` actually doesn't depends on Yojson :-)
> 
> Did you also check that other important parsing function such as the one parsing the flags to build the configuration are already exposed by merlin lib ?

Yes. I am integrating test into my `ocaml-lsp` branch. I'll do the PR today.

xvw:
Done!

PR #1766, opened 2024-05-13
Add support for project-wide occurrences

voodoos:


src/analysis/occurrences.ml
```diff
@@ -87,12 +95,13 @@ let index_buffer =
...
-  fun ~current_buffer_path ~stamp ~local_defs () ->
+  fun ~scope ~current_buffer_path ~stamp ~local_defs () ->
     let {Logger. log} = Logger.for_section "index" in
     match !cache with
-    | Some (path, stamp', value) when
+    | Some (path, stamp', scope', value) when
         String.equal path current_buffer_path
-        && Int.equal stamp' stamp ->
+        && Int.equal stamp' stamp
+        && scope' = scope ->
```
xvw, line 104: Could we benefit to add a `Scope.equal` (probably written as `let equal = (=)` in order to try to get rid of `=` and providing, in a future, an implementation of scopes equality?

src/index-format/dune
```diff
@@ -0,0 +1,10 @@
+(library
+ (name merlin_index_format)
```
xvw, line 2: Why the name `index_format`?

xvw (approved):
👍🏾

src/index-format/dune
voodoos: This was answered offline: it uses the same convention as compiler artifacts: `cmt_format`, `cmi_format` etc.

PR #1768, opened 2024-05-16
Do not index occurrences from ppxes 

voodoos:
Current version of the indexing iterator actually works on the ppxed-typedtree. This means that we should detect part of the tree that are introduced by PPXes and not index them. 

This PR tries to do that in the usual way: it looks for the `merlin.hide` attribute on some strategic nodes.

@pitag-ha I am still a bit uncertain of which nodes should be considered to look for the `merlin.hide` attributes ?
Once again, that will only work with ppxes that actually use this attribute... I guess we can have that certainty only with ppxes written using `ppxlib` ?

We should try to devise a most robust way to distinguished ppxed part of the tree...

cc @Octachron since similar changes should be upstreamed at some point to prevent project-wide occurrences to return location that are actually coming from ppxes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@ merlin NEXT_VERSION
     - Addition of a `merlin-lib.commands` library which disassociates the
       execution of commands from the `new_protocol`, from the binary, allowing
       it to be invoked from other projects (#1758)
+    - New occurrences backend: Don't index occurrences when `merlin.hide`
+      attribute is present. (#1768)
 
 merlin 4.14
 ===========
--- a/src/analysis/ast_iterators.ml
+++ b/src/analysis/ast_iterators.ml
@@ -3,6 +3,95 @@ open Typedtree
 
 let {Logger. log} = Logger.for_section "iterators"
 
+(* Sometimes we do not want to iterate on nodes that do not correspond to actual
+    syntax such as the ones introduced by PPXes with the `merlin.hide`
+    attribute. *)
+let iter_only_visible iter =
+  let has_attribute ~name attrs =
+    List.exists ~f:(fun a ->
+      let (str,_) = Ast_helper.Attr.as_tuple a in
+      str.Location.txt = name
+    ) attrs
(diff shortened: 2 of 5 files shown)


pitag-ha:
> which nodes should be considered to look for the merlin.hide attributes ?

@voodoos and I have already discussed this in private, but in case it's interesting to know what we've discussed: Let's distinguish derivers from other PPXs such as extension node rewriters.

For derivers, we'd only need to look for `merlin.hide` attrs attached to strucutre/signature items, because Ppxlib wraps the derived code into a structure/signature item with `merlin.hide` attr. The reason for that is to preserve the location invariants: Without `merlin.hide` attr, the derived nodes would duplicate the location of a sibling node. All this is only true for Ppxlib derivers. Ppx_deriving doesn't add a `merlin.hide` attribute, which is why Ppx_deriving derivers have a tendency to mess up Merlin in general.

For other PPXs, e.g. extension node rewriters, it's not Ppxlib itself who adds the `merlin.hide` attributes, but the PPX author who can do so for certain nodes via Ppxlib's `Merlin_helpers.hide_attribute`. While it's more likely to add that attribute on expressions or patterns than on other nodes, it can be done on any node allowing an attribute. E.g. `ppx_compare` does so [on a core_type](https://github.com/janestreet/ppx_compare/blob/master/expander/ppx_compare_expander.ml#L199).

So, it's safest here to look for the `merlin.hide` attribute on any node allowing attributes.

pitag-ha:
To my not-very-familiar-with-the-typedtree eyes, this looks good to me up to one doubt I have: The iteration through a node is not necessary to update the typing environment / the scope to what happens in the node, right? I'm asking because, with the PR, the iteration is fully skipped if there's a `merlin.hide` attribute. So if the iteration was needed to update the scope, then there might be the following problem: Whenever derived code (i.e. code with `merlin.hide` attribute) shadows the value whose occurrences are being analyzed, there might be false positives of those occurrences in the rest of the file.

PR #1769, opened 2024-05-16
nix: Build with OCaml 5.2

Julow:
This changes flake.nix to build Merlin with OCaml 5.2 in order to support 5.2-only features.

Nixpkgs needed to be updated too.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -39,11 +39,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1688587528,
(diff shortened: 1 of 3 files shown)


Julow:
I changed `nix develop` to install the current merlin instead of ocaml-lsp (which pulled nixpkgs' version of merlin). I changed the definition so that adding ocaml-lsp back to the devShells would use the current merlin but the last release do not build that way.
It's possible to pin a branch of ocaml-lsp but I think this would add too much maintenance.
Are there any users of `nix develop` ? (other than the CI, which doesn't need ocaml-lsp)

voodoos (approved):
Thanks a lot @Julow !

PR #1788, opened 2024-06-19
`WRAPPING_PREFIX` directive

liam923:
Add a `WRAPPING_PREFIX` directive. When used, the prefix will be appended to the name of the current unit. This PR also makes two other changes:
1. Add a test for project-wide occurrences on a directory that has prefixed module names, using the `UNIT_NAME` and `WRAPPING_PREFIX` directives.
2. Fix a bug with `UNIT_NAME`.


voodoos (approved):
Looks good to me, thanks @liam923 

src/dot-merlin/dot_merlin_reader.ml
```diff
@@ -425,6 +431,10 @@ let postprocess cfg =
         (dirs :> Merlin_dot_protocol.directive list)
       )
     ; (cfg.pass_forward :> Merlin_dot_protocol.directive list)
+    ; cfg.unit_name |> Option.map ~f:(fun name -> `UNIT_NAME name) |> Option.to_list
+    ; cfg.wrapping_prefix
+        |> Option.map ~f:(fun prefix -> `WRAPPING_PREFIX prefix)
+        |> Option.to_list
```
voodoos, line 437: Actually, since these variant are part of the `no_processing_required` ones we should simply forward them:
We should remove both `unit_name` and `wrapping_prefix` from `Dot_merlin_reader.config` and simply push them to the `pass_forward` list.

voodoos (changes requested):

merlin.opam
```diff
@@ -72,5 +72,5 @@ See https://github.com/OCamlPro/opam-user-setup
   {success & !user-setup:installed}
 ]
 pin-depends: [
-  ["ocaml-index.1.0"  "git+https://github.com/voodoos/ocaml-index#82b08987921884daeeb5dccc345a2dcb667fe113"]
+  ["ocaml-index.1.0"  "/usr/local/home/lstevenson/github/ocaml-index-upstream"]
```
voodoos, line 75: This looks like a mistake :-) 
(the release is imminent, I will get rid of the pin as soon as possible)

liam923: Oops yes I committed that by mistake, thanks for the catch. I've fixed it.

PR #1800, opened 2024-07-16
Refinement in the presence of optional arguments

### LINKED ISSUE #1770: Optional argument include by `case-analysis`
With the file `test.ml`:
```
let foo ?bar x = x

let () = foo ()
```
running
```
$ ocamlmerlin single case-analysis -start 3:10 -end 3:15 -filename test.ml < test.ml | jq ".value[1]"
```

**Expected:**
```
"match foo () with | () -> _"
```

**Actual:**
```
"match foo ?bar:None () with | () -> _"
```

I believe the root of this issue is in `Untypeast`. Looking at `Destruct.destruct_expression`, it logs the expression being destructed after passing it to `Untypeast.untype_expression`. For this program, it is:
```
# 0.01 destruct - node_expression
expression (test.ml[3,20+9]..[3,20+15])
  Pexp_apply
  expression (test.ml[3,20+9]..[3,20+12])
    Pexp_ident "foo" (test.ml[3,20+9]..[3,20+12])
  [
    <arg>
    Optional "bar"
      expression (_none_[0,0+-1]..[0,0+-1]) ghost
        Pexp_construct "None" (_none_[0,0+-1]..[0,0+-1]) ghost
        None
    <arg>
    Nolabel
      expression (test.ml[3,20+13]..[3,20+15])
        Pexp_construct "()" (test.ml[3,20+13]..[3,20+15])
        None
  ]
```
We can see here that the `~bar` argument has snuck into our expression. I think solving this might involve some Merlin-specifc logic in `Untypeast` that detects optional arguments that users didn't supply and strips them out.

### Version
On branch `master` (commit `fa9b391529d1b01f091bd6d565efa2cee545d58c`)

xvw:
~I don't know whether the fixed system might break some of the parameters generated by PPX. (Perhaps if there is a ghost on the parent expression, you can check that it is the `None` constructor?).~

After a conversation with @pitag-ha , we came to the conclusion that the introduction of ghost location in this context (which, in PPX terms, should be reserved for derivers) is a problem of semantics ‘on the PPX side’, so that **this fix**, which is based on ghost location, is not problematic (for PPX).

Fix #1770

### DIFF
--- a/src/analysis/destruct.ml
+++ b/src/analysis/destruct.ml
@@ -588,9 +588,44 @@ module Conv = struct
     (ps, constrs, labels)
 end
 
+let need_recover_labeled_args = function
+  | Parsetree.Pexp_construct ({loc; txt = Longident.Lident ctor}, Some e) ->
+    (* If the internal construction is ghosted, then the expression must be
+       re-labelled. *)
+    if String.equal "Some" ctor && loc.loc_ghost then Some e else None
+  | _ -> None
+
+let remove_non_applied_optional_args (Parsetree.{ pexp_desc; _} as base_expr) =
+  (* Fix the behaviour described here
+     https://github.com/ocaml/merlin/issues/1770 *)
+  match pexp_desc with
+  | Parsetree.Pexp_apply (expr, args) ->
+    let args = List.concat_map ~f:(fun (label, expr) ->
+      match label with
+      | Asttypes.Optional str ->
+        (* If an optional parameter is not applied, its location is assumed to
+           be ghost, and the parameter should not be generated. *)
+        let loc = expr.Parsetree.pexp_loc in
+        if loc.loc_ghost
+        then []
+        else begin
+          match need_recover_labeled_args expr.pexp_desc with
+          | Some e ->  [(Asttypes.Labelled str, e)]
+          | None ->  [(label, expr)]
+        end
+      | _ -> [(label, expr)]
+    ) args
+    in
+    let pexp_desc = Parsetree.Pexp_apply (expr, args) in
+    { base_expr with pexp_desc }
+  | _ -> base_expr
+
 let destruct_expression loc config source parents expr =
   let ty = expr.Typedtree.exp_type in
-  let pexp = filter_expr_attr (Untypeast.untype_expression expr) in
+  let pexp =
+    filter_expr_attr (Untypeast.untype_expression expr)
+    |> remove_non_applied_optional_args
+  in
   let () =
     log ~title:"node_expression" "%a"
       Logger.fmt (fun fmt -> Printast.expression 0 fmt pexp)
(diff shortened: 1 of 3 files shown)


src/analysis/destruct.ml
```diff
@@ -588,9 +588,42 @@ module Conv = struct
     (ps, constrs, labels)
 end
 
+let need_recover_labeled_args = function
+  | Parsetree.Pexp_construct ({loc; txt = Longident.Lident ctor}, Some e) ->
+    (* If the internal construction is ghosted, then the expression must be
+       re-labelled. *)
+    if String.equal "Some" ctor && loc.loc_ghost then Some e else None
+  | _ -> None
```
voodoos, line 596: I don't understand why this is needed :-) 
What would happen in the testsuite examples without it ?

src/analysis/destruct.ml
```diff
@@ -588,9 +588,42 @@ module Conv = struct
...
+        else begin
+          match need_recover_labeled_args expr.pexp_desc with
+          | Some e ->  [(Asttypes.Labelled str, e)]
+          | None ->  [(label, expr)]
+        end
+      | _ -> [(label, expr)]
+    ) args
+    in
+    let pexp_desc = Parsetree.Pexp_apply (expr, args) in
+    { expr with pexp_desc }
```
voodoos, line 618: There seems to be a looping use of `expr` here, is that on purpose ?

voodoos (commented):
Looks good, thanks @xvw !

I just have a few understanding questions :-) 
Maybe you could mention the issue in the comments for future reference

src/analysis/destruct.ml
xvw:
```
$ $MERLIN single case-analysis -start 2:10 -end 2:15 \
  > -filename main.ml <<EOF
  > let foo ?(bar = 10) x = x
  > let () = foo ~bar:15 ()
  > EOF
  {
    "class": "return",
    "value": [
      {
        "start": {
          "line": 2,
          "col": 9
        },
        "end": {
          "line": 2,
          "col": 23
        }
      },
      "match foo ?bar:(Some 15) () with | () -> _"
    ],
    "notifications": []
  }
```

and 

```
$ $MERLIN single case-analysis -start 2:10 -end 2:15 \

  $ $MERLIN single case-analysis -start 2:10 -end 2:15 \
  > -filename main.ml <<EOF
  > let foo ?bar x = x
  > let () = foo ~bar:10 ()
  > EOF
  {
    "class": "return",
    "value": [
      {
        "start": {
          "line": 2,
          "col": 9
        },
        "end": {
          "line": 2,
          "col": 23
        }
      },
      "match foo ?bar:(Some 10) () with | () -> _"
    ],
    "notifications": []
  }

```

xvw:
(Comment referencing the issue was added)

voodoos (approved):
Thanks !

liam923:
Thank you!

ncik-roberts:
In https://github.com/janestreet/merlin-jst/pull/73#issuecomment-2253475246 I suggested only omitting a ghost-location argument when it's `None`. Maybe that's preferable here too?

xvw:
> In [janestreet/merlin-jst#73 (comment)](https://github.com/janestreet/merlin-jst/pull/73#issuecomment-2253475246) I suggested only omitting a ghost-location argument when it's `None`. Maybe that's preferable here too?

Yep, I'll do a patch today! Thanks for the feedback.

PR #1801, opened 2024-07-17
Fix deletion of preprocessor output file after usage

### LINKED ISSUE #1802: Permissions and deletion of temporary files created by preprocessors (pp)
Hello,
When a preprocessor (pp) is used, the input file is passed as an argument to the preprocessor executable and then redirected to a temporary file before being read again and deleted.
https://github.com/ocaml/merlin/blob/ce924954140eb382955f946fac5cea86130b49bf/src/ocaml/driver/pparse.ml#L162

The first problem I have is that, when correctly read, the temporary file is never deleted which leads to an accumulation of `merlinpp****.out` in the `/tmp/` directory. I propose a fix of this issue in PR #1801.

The second problem is that the temporary file is created, at least on Linux, via a redirection of the stdout of the preprocessor program to a temporary file.
https://github.com/ocaml/merlin/blob/ce924954140eb382955f946fac5cea86130b49bf/src/utils/std.ml#L787
So, the permissions of this file are by default, at least on my computer, `0o644` which means everybody can read its content. Note that in the case of the input file, it is copied via a call to the function `Filename.temp_file` that uses permissions `0o600` (only owner can read and write).
I know this file only exists briefly (at least after the patch), and reading the source files requires being able to execute commands, but I'm not comfortable having source files accessible to everyone.
A possible fix could be to first create the source file via `Filename.temp_file` and then write into it with the redirection? I think this is the only place where the function `System.run_in_directory` is used with the `stdout` argument set. Should I include this in the PR? Should the creation of such a file be done in `System.run_in_directory` or in pparse.ml`?

Have a good day!

jboillot:
When correctly read, the temporary file used to store the result of the preprocessor on the input file is never deleted, which leads to accumulation of files in the `/tmp/` directory.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ unreleased
     - Add `-unboxed-types` and `-no-unboxed-types` as ocaml ignored flags (#1795, fixes #1794)
   + editor modes
     - vim: fix python-3.12 syntax warnings in merlin.py (#1798)
+    - Fix merlinpp not removing temporary files (#1801, partly fixes #1802)
 
 merlin 5.1
 ==========
--- a/src/ocaml/driver/pparse.ml
+++ b/src/ocaml/driver/pparse.ml
@@ -180,6 +180,7 @@ let apply_pp ~workdir ~filename ~source ~pp =
     let ic = open_in_bin fn_out in
     let result = Misc.string_of_file ic in
     close_in ic;
+    Misc.remove_file fn_out;
     Ok result
 
 let decode_potential_ast source =


panglesd (approved):
Thanks, looks good to me. A file that we never use anymore (even in the cache) should be deleted.

The second part of #1802 can be in another PR.

(And this needs rebase for the changelog conflict, which I cannot do.)

voodoos:
Thanks for reviewing @panglesd I trust that you checked that the file in the tmp directory is not used as a kind of cold cache by merlin ?

panglesd:
Yes, I did check that (the file name is created in the function as `<tmpfilename>.out`, `<tmpfilename>` is generated unique and not returned by the function)

PR #1803, opened 2024-07-17
Fix ignorance of STDLIB in .merlin

liam923:
Currently, the directive `STDLIB` inside a `.merlin` file is ignored. This PR fixes this bug.

### DIFF
--- a/src/dot-merlin/dot_merlin_reader.ml
+++ b/src/dot-merlin/dot_merlin_reader.ml
@@ -356,8 +356,7 @@ let prepend_config ~cwd ~cfg =
       | Some p ->
         log ~title:"conflicting paths for stdlib" "%s\n%s" p canon_path
       end;
-      { cfg with stdlib = Some canon_path;
-                 pass_forward = `STDLIB canon_path :: cfg.pass_forward }
+      { cfg with stdlib = Some canon_path }
     | `SOURCE_ROOT path ->
       let canon_path = canonicalize_filename ~cwd path in
       { cfg with source_root = Some canon_path }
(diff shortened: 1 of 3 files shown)


src/dot-merlin/dot_merlin_reader.ml
```diff
@@ -356,7 +356,8 @@ let prepend_config ~cwd ~cfg =
       | Some p ->
         log ~title:"conflicting paths for stdlib" "%s\n%s" p canon_path
       end;
-      { cfg with stdlib = Some canon_path }
+      { cfg with stdlib = Some canon_path;
+                 pass_forward = `STDLIB canon_path :: cfg.pass_forward }
```
voodoos, line 360: Is this the correct place to fix the issue ? Why do we need to both update the threaded config and add it to the pass_foward list ?

I would expect the issue being related to some configuration merging that ignores the correct stdlib field in config.

liam923: Hm I think you're right that this isn't the best way to do it - I've updated the PR.

The issue doesn't seem to be with any configuration merging. I think it is instead that `dot_merlin_reader` never passes on the `STDLIB` directive, so `mconfig_dot` never sees it.

src/dot-merlin/dot_merlin_reader.ml
```diff
@@ -427,6 +427,7 @@ let postprocess cfg =
         (dirs :> Merlin_dot_protocol.directive list)
       )
     ; (cfg.pass_forward :> Merlin_dot_protocol.directive list)
+    ; cfg.stdlib |> Option.map ~f:(fun stdlib -> `STDLIB stdlib) |> Option.to_list
```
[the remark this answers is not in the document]
liam923: I think this results in incorrect behavior, as the `-ocamllib-path` flag will be ignored. With this change, it will appear to `mconfig_dot.ml` as if the `.merlin` file contained the `STDLIB` directive, which it will then prioritize over the `-ocmallib-path` flag. Indeed, I added a test for this case and confirmed that accepting this change would break the test.

liam923:
> Can you rebase the PR on `main` ? It should help fixing the CI issues.

Done!

Issue #1755, opened 2024-04-30
Hover of a mutually recursive type can cause a stack overflow 


Chaarlesss:
Hi,

Merlin produces a stack overflow which causes Emacs to freeze when hovering the type `b` on line 1 of the following code:
```ocaml
type a = A | B of int * b
and b = a list
```
Strangely, the problem does not occur when hovering the type `a` on line 2.

I am using Merlin 4.14-501 with OCaml 5.1.1.

The end of the stack trace is:
![stack trace](https://github.com/ocaml/merlin/assets/12775522/dd896f64-6ef8-4902-b5a8-2d5025a65741)

voodoos:
Are you using merlin mode or lsp ?

Could it be that your "hover" query has a high verbosity by default ?

Chaarlesss:
I am using merlin mode (with [merlin-eldoc](https://github.com/Khady/merlin-eldoc)).

To clarify, I found this error in the log's buffer and does not appear when the problem occurs.

I guess the freeze may come from my own configuration of Emacs (which is roughly Doom Emacs with the OCaml module), but I think that the underlying problem may come from Merlin.

I also tried to reproduce this in VSCode (standard setup with OCamlPlatform), and I also get a strange result (but I am not sure Merlin the one in cause here):
![image](https://github.com/ocaml/merlin/assets/12775522/2680d939-dfd3-4748-83cc-fc4f09ca6f00)

voodoos:
Looking into eglot, it does seems that there is a verbosity setting that can be `min` or `max`. You might want to set that on `min`.

Verbosity was not meant to be always set on max and it can lead to strange issues with recursive types.

While setting verbosity to min might be a workaround I agree that the behavior of Merlin is not satisfying here, it would be worth to have a look.

Chaarlesss:
Changing the verbosity fixed indeed the freezing. Thanks for pointing me to this solution !

raphael-proust:
possibly the same bug as https://github.com/ocaml/merlin/issues/822

panglesd:
Closed by #1805!

PR #1805, opened 2024-07-24
Type enclosing from inside recursive type def

### LINKED ISSUE #1335: Stack overflow when verbosity = 1

voodoos:
Merlin used to stack-overflow when querying type-enclosing from inside a recursive type definition with verbosity >= 1. This is no longer the case as shown in these two tests. However the results are unsatisfying because the environment around the cursor is not aware of the recursive definition.

Closes #1335 and #1755


ncik-roberts:
Do you know when or how this was fixed in merlin?

voodoos:
Unfortunately we don't know which change stopped Merlin from crashing in these cases. My guess would be that it is indeed the result of merging upstream changes to the `Printtyp` module.

PR #1807, opened 2024-07-29
Check always that default args are option

xvw:
A follow up of #1800, based on https://github.com/ocaml/merlin/pull/1800#issuecomment-2253532309


voodoos (approved):
Will merge once the CI pass.

Could you add this PR number to the changelog entry for #1800 ?

PR #1810, opened 2024-08-15
Produce a better error message when a flag spec appears multiple times

ncik-roberts:
As title: just a small change to make a better error message for devs that want to see the name of an incorrectly-duplicated flag spec.


voodoos:
Thanks @ncik-roberts and @xvw !

PR #1811, opened 2024-08-28
Exposes some helper for reducing direct typedtree usage in Lsp

xvw:
Introduction of a module `Merlin_analysis.Typedtree_utils` that can be used in `lsp`. Here, it introduces two function to refactor inferInterface and actionInline.

### DIFF
--- a/src/analysis/ast_iterators.ml
+++ b/src/analysis/ast_iterators.ml
@@ -99,7 +99,7 @@ these cases. *)
 let iter_on_defs ~uid_to_locs_tbl =
   let log = log ~title:"iter_on_defs" in
   let register_uid uid fragment =
-    let loc = Misc_utils.loc_of_decl ~uid fragment in
+    let loc = Typedtree_utils.location_of_declaration ~uid fragment in
     Option.iter loc ~f:(fun loc ->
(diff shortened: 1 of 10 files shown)


voodoos (commented):
Thanks Xavier. We have a few other typedtree utils laying around, notably in `Mist_utils`.  I think it would be worth it to organize them all in the same, better named, `Typedtree_utils` module.

Could you also make sure to provide doc comments for every of these utils ?

xvw:
Taking account of your feedbacks @voodoos, thanks!
Two question: 
- I don't think that kind of refactoring deserve a CHANGE entry
- `Misc_utils.Compat` (now `Typedtree_utils.Compat`) seems not used, even in `ocaml-lsp`, should we keep it?

voodoos:
>     * `Misc_utils.Compat` (now `Typedtree_utils.Compat`) seems not used, even in `ocaml-lsp`, should we keep it?

It is not used because we eventually did not make a multi-version release of ocaml-lsp. I would keep it for now, it should be useful when we try again to to it. (No need to keep the Compat module however, you can extract the functions that are inside.)

xvw:
Done!

voodoos (approved):
Thanks !

PR #1812, opened 2024-08-28
Inlay hint upstreaming

xvw:
Currently, the command returns a list of `postion * label` and the label is _computed_ in the sense of `lsp` : `: type_expr` but I guess we can just return the type without any formatting and let's the LSP server prefixing it? WDYT @voodoos?

### DIFF
--- a/src/analysis/inlay_hints.ml
+++ b/src/analysis/inlay_hints.ml
@@ -135,18 +135,11 @@ let structure_iterator
   in iterator.structure iterator typedtree
 
 type hint = Lexing.position * string
-
-let outline_type env typ =
-  let is_word_char = function
-    | ' ' | '\t' | '\n' -> false
-    | _ -> true
-  in
-  Ocaml_typing.Printtyp.wrap_printing_env env (fun () ->
-      Format.asprintf "@[<h>: %a@]" Ocaml_typing.Printtyp.type_scheme typ
-    ) |> String.extract_words ~is_word_char |> String.concat ~sep:" "
     
 let create_hint env typ loc =
-  let label = outline_type env typ in
+  let label =  Printtyp.wrap_printing_env env (fun () ->
+    Format.asprintf "%a" Printtyp.type_scheme typ)
+  in
   let position = loc.Location.loc_end in
   (position, label)
 
--- a/src/analysis/inlay_hints.ml
+++ b/src/analysis/inlay_hints.ml
@@ -4,11 +4,6 @@ let {Logger.log} = Logger.for_section "inlay-hints"
 
 module Iterator = Ocaml_typing.Tast_iterator
 
-let overlap_with_loc (start, stop) loc =
-  let a = Lexing.compare_pos start loc.Location.loc_end
-  and b = Lexing.compare_pos stop loc.Location.loc_start in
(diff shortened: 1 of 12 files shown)


voodoos:
Yes, I agree that we should return the raw type.

xvw:
> Yes, I agree that we should return the raw type.

Fixed in b785dcfbabcf0f6472bccc87e2517d011df44aef

src/analysis/inlay_hints.ml
```diff
@@ -0,0 +1,183 @@
+open Std
+
+let {Logger.log} = Logger.for_section "inlay-hints"
+
+module Iterator = Ocaml_typing.Tast_iterator
+
+let overlap_with_loc (start, stop) loc =
+  let a = Lexing.compare_pos start loc.Location.loc_end
+  and b = Lexing.compare_pos stop loc.Location.loc_start in
+  a <= 0 && b >= 0 || a >= 0 && b <= 0
```
voodoos, line 10: I have the usual question: doesn't this helper function already exists elsewhere ? If not, maybe it could :-)

src/analysis/inlay_hints.ml
```diff
@@ -0,0 +1,183 @@
...
+    pipeline =
+  let () = log ~title:"start" "%a" Logger.fmt (fun fmt ->
+      Format.fprintf fmt "Start on %s to %s with : let: %b, pat: %b, ghost: %b"
+        (Lexing.print_position () start)
+        (Lexing.print_position () stop)
+        hint_let_binding
+        hint_pattern_binding
+        avoid_ghost_location)
+  in
+  let typer_result = Mpipeline.typer_result pipeline in
```
voodoos, line 161: This should move to `Query_command`. The usual separation is that pipeline manipulation is done in query command and tools in analysis are then passed the resulting typedtree (or config, parsetree, etc). That way they are independent of the kernel.

src/analysis/inlay_hints.mli
```diff
@@ -0,0 +1,12 @@
...
+
+type hint = Lexing.position * string
+
+val run :
+  hint_let_binding:bool
+  -> hint_pattern_binding:bool
+  -> avoid_ghost_location:bool
+  -> start:Lexing.position
+  -> stop:Lexing.position
+  -> Mpipeline.t
```
voodoos, line 11:
```suggestion
  -> Mtyper.typedtree
```

src/analysis/inlay_hints.ml
xvw: Right, moved into `Location_aux`.

src/analysis/inlay_hints.ml
xvw: Done! You are right, it is better now!

src/analysis/inlay_hints.mli
xvw: I go for `Typedtree.structure` instead of the `Mtyper.typedtree`, is that ok for you?

voodoos: It makes sens if the feature only works for implementations.

xvw: Yes

PR #1814, opened 2024-08-29
Some UI improvement for `emacs/merlin-search`

xvw:
Since polarity research output a result in a popup that take the focus, the user experience was a little bit complicated. So the patch returns the result inside a dedicated buffer:

image

(I was not able to take a screenshot about the "before", because the popup was taking the focus...)

### DIFF
--- a/emacs/merlin.el
+++ b/emacs/merlin.el
@@ -137,6 +137,10 @@ a call to `merlin-occurrences'."
 See `merlin-debug'."
   :group 'merlin :type 'string)
 
+(defcustom merlin-polarity-search-buffer-name "*merlin-polarity-search-result*"
+  "The name of the buffer displaying result of polarity search."
+  :group 'merlin :type 'string)
+
 (defcustom merlin-favourite-caml-mode nil
   "The OCaml mode to use for the *merlin-types* buffer."
   :group 'merlin :type 'symbol)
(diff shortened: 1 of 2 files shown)


emacs/merlin.el
```diff
@@ -1099,17 +1103,44 @@ An ocaml atom is any string containing [a-z_0-9A-Z`.]."
...
-                      `(lambda () (insert ,text)))))))
-    (popup-menu (easy-menu-create-menu "Results" (mapcar transform entries)))))
+         (previous-buff (current-buffer)))
+    (let ((pol-buff (merlin--get-polarity-buff))
+          (inhibit-read-only t))
+      (with-current-buffer pol-buff
+        (switch-to-buffer-other-window pol-buff)
+        (goto-char 1)
+        (tabulated-list-mode)
+        (setq tabulated-list-format [("functions" 100 t)])
```
voodoos, line 1135: Not sure what's your mean here: in your screenshots some results are not functions...

voodoos (commented):
The formatting changes make it inconvenient to review the final diff but the changes look reasonable to me.

xvw:
> The formatting changes make it inconvenient to review the final diff but the changes look reasonable to me.

I can try to remove the formatting.

PR #1759, opened 2024-05-06
Rewrite merlin-completion-at-point integration to be faster and better

catern:
Let me know how you'd like me to approach this PR.  This is what we use internally at Jane Street, and it's been written to be relatively upstreamable.  At the same time, the old version of merlin-cap was in need of a lot of improvement and modernization to how the Emacs completion API is supposed to be used, and essentially none of the original code survived.  So it's kind of just a giant code drop, which I know is pretty hard to review.

The merlin-completion-at-point integration is completely rewritten.  The version before was simple and slow; the new version is much faster and much more featureful:

- Completions are requested for an entire module at once and filtered locally in Emacs rather than doing the filtering on the Merlin side.

- Completions are cached based on the OCaml atom being completed, so they are re-used instead of re-requested when a new character is typed.

- Those completions are cached for a given position inside an OCaml atom, so that if a user types Li TAB ma TAB to complete "List.map" and then decides they actually want the module "Labels", when they delete the "ist.map" part and hit TAB, they'll use the previously-requested completions.

- We avoid updating Merlin with the new buffer contents as completion proceeds, so that Merlin doesn't need to re-parse and re-type-check, substantially improving performance in expensive
files.  (merlin-cap--omit-bounds)

- Completion requests are handled asynchronously and reused, so that if completion is interrupted and then resumed, we're able to use the results of the previous completion request.  This makes completion UIs which use while-no-input (like corfu-mode) much more performant.

- Completion is wrapped in while-no-input when non-essential is set; this makes completion UIs which don't use while-no-input (like company-mode) much more responsive.

- Completions are sorted more intelligently: if they're a constructor or variant or label, they're likely to be more relevant to the user, so they're sorted first.

- Module names in completions are suffixed with a ., matching Emacs behavior for file name completion (where directories are suffixed with a /); this makes completion of module paths much more fluent, since there's no need to hit . after every module name.

- We use completion boundaries, so the built-in Emacs partial-completion feature now works: if the user types "Li.ma TAB", it will complete to "List.map".

- Likewise, partial-completion will expand * as a glob, so if the user types "Deferred.*.map TAB" they will be presented with every module in "Deferred." which contains the method "map"

There are also several tests now, testing the new functionality.

PR #1819, opened 2024-09-13
Move ert tests into separate file and run tests in CI.

tmcgilchrist:
The improvements in https://github.com/ocaml/merlin/pull/1759 introduced a dependency on `ert` tests which makes `merlin-mode` in ELPA depend on test code. It would be better to not require test code when using `merlin-mode`.
 
This change puts the test code into its own file and introduced a CI step to run the tests. 

How does the ELPA release process work? The new `merlin-cap-test.el` should be excluded from the package sent to ELPA.


tmcgilchrist:
https://github.com/tmcgilchrist/merlin/pull/2 shows test failures on Emacs 28 and below.

Related to this `merlin-mode` now uses functions only available in 29.1. Should older there be support for versions before this?

```
$  opam exec -- ./check.sh
....
merlin-cap.el:1:86: warning: You should depend on (emacs "24.1") if you need lexical-binding.
merlin-cap.el:19:10: error: You should depend on (emacs "24.4") if you need `subr-x'.
merlin-cap.el:21:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:30:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:50:1: error: You should depend on (emacs "24.4") if you need `define-error'.
merlin-cap.el:51:38: error: You should depend on (emacs "29.1") if you need `position-symbol'.
merlin-cap.el:56:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:67:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:73:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:75:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:77:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:79:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:81:1: error: You should depend on (emacs "24.3") if you need `defvar-local'.
merlin-cap.el:112:20: error: You should depend on (emacs "25.1") or the compat package if you need `when-let'.
merlin-cap.el:126:23: error: You should depend on (emacs "25.1") if you need `make-pipe-process'.
merlin-cap.el:134:23: error: You should depend on (emacs "25.1") if you need `make-process'.
merlin-cap.el:170:5: error: You should depend on (emacs "24.3") or the cl-lib package if you need `cl-assert'.
merlin-cap.el:171:5: error: You should depend on (emacs "25.1") or the compat package if you need `when-let'.
merlin-cap.el:224:20: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:225:19: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:230:22: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:231:61: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:251:22: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:251:56: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:252:21: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:253:21: error: You should depend on (emacs "25.1") or the compat package if you need `alist-get'.
merlin-cap.el:259:31: error: You should depend on (emacs "24.4") if you need `string-remove-suffix'.
merlin-cap.el:339:13: error: You should depend on (emacs "28.1") or the compat package if you need `string-search'.
merlin-cap.el:354:3: error: You should depend on (emacs "25.1") or the compat package if you need `when-let'.
merlin-cap.el:354:19: error: You should depend on (emacs "28.1") or the compat package if you need `string-search'.
merlin-cap.el:374:18: error: You should depend on (emacs "28.1") or the compat package if you need `string-search'.
merlin-cap.el:485:0: error: Aliases should start with the package's prefix "merlin-cap".
```

xvw:
Hi! Thanks a lot! 
What do you think about moving test files in a dedicated subfolder inside `emacs/` (ie: `emacs/test`)? And could you add in the readme a way to run tests locally?

Aside, I think that installing `merlin-mode` via ELPA/MELPA is not the right way to do since MELPA use `main` as a source. So we can lead to have a feature implemented in `merlin-mode`, still not available in `ocamlmerlin`. We can probably, at the mode level, relay on merlin version to avoid that, but FMPOV, just relaying on the files installed into `opam var share` is more robuts.

voodoos:
cc @purcell who wrote melpa's receipe

purcell:
> What do you think about moving test files in a dedicated subfolder inside emacs/ (ie: emacs/test)?

MELPA already excludes `*-test.el`, unsure how ELPA handles things. A subfolder would work well for MELPA too — I can easily update the package recipe there.

> Aside, I think that installing merlin-mode via ELPA/MELPA is not the right way to do since MELPA use main as a source. So we can lead to have a feature implemented in merlin-mode, still not available in ocamlmerlin.

Regarding MELPA using `main`, that's not a huge issue since there's also the [MELPA Stable package version](https://stable.melpa.org/#/merlin) based on the latest version tag here — users can use that instead if they have issues. Shipping merlin-mode via opam just means the elisp here can't rely on any proper Emacs package dependencies, and it's not a great install/set-up experience for Emacs users, though I get your point about it helping the mutual compatibility.

xvw:
@purcell Thanks a lot for the explanation (and for the clarification about `-test` files.
Just out of curiosity, how to deal, via MELPA (even in stable version) to deal with specific version of `merlin-mode`. For example, let's imagine that I use, in my switch `merlin xxx.xxx` but the latest stable version of `elpa/melpa` is `merlin yyy.yyy`? The only to "fix that" seems to add a way to check the version of `merlin` inside the `el file`, since we will take some time to write a advanced extension mode for LSP via Emacs, it is something interesting to keep in mind.

xvw:
Note that "partially answer to one of my previous question"

It seems that we can probably deal with version difference using a mechanism like "capabilities" (already present in LSP).

purcell:
Yes, generally you'd indeed have to do some version checking of `merlin` within the elisp and handle any difference in capabilities there. With `package.el` you can't explicitly install older versions of packages, so it's all a little primitive. :)

tmcgilchrist:
@xvw I'm running the tests as `opam exec -- ./check.sh` which expects a local Emacs installation and an opam switch for `merlin`.  

@purcell What's the recommended way to deal with backwards compatible elisp? I thought about using `boundp` and `fboundp` to detect new functions like `string-search` and providing a fall back compatibility function. Is there a better option than that? Maybe using the Advising Functions https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html

purcell:
> What's the recommended way to deal with backwards compatible elisp? I thought about using `boundp` and `fboundp` to detect new functions like `string-search` and providing a fall back compatibility function. Is there a better option than that? Maybe using the Advising Functions https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html

Depending on [the `compat` package](https://elpa.gnu.org/packages/compat.html) solves a lot of that. Otherwise, you'd normally make your own wrapper, e.g.
```
(if (fboundp 'string-search)
    (defalias 'merlin--string-search 'string-search)
  (defun merlin--string-search (...)
     ...))
```
and use that wrapper in the code. It would be frowned upon to simply define `string-search` if it is found to be missing (as it could be a sentinel for other packages, or the local reimplementation could be wrong), and advising functions generally implies changing someone else's function, which is also not often a good thing for a package to do.

voodoos (changes requested):
Thanks a lot @tmcgilchrist !

Would you have the time to implement the suggestions of @xvw  and @purcell ?

- [ ] Move the tests to a separate folder and document how they can be run.
- [ ] Depend on the compat package to see if it solves the backward compatibility issue

voodoos:
So indeed tests are falling on `emacs <= 28.2`, which is very unfortunate to say the list. 

I will try to test if completion does indeed work or not with emacs 28.

voodoos:
@xvw do you think you could check that the new completion at point work on Emacs 28.2 ? I am having difficulties installing it on my machine...

voodoos:
@purcell if it turns out that the new completion does not work on emacs <= 28.2 and that the fix is not-trivial, would it be an option to split merlin in two recipes, a legacy one which ships the old `merlin-cap` module and another one that ships the new completion module for emacs >= 29 ?

voodoos:
There is another issue with this: merlin mode now require the `compat` package and that breaks the installation with `opam user-setup` which does not install that package automatically.

To summarize the current issues:

- The new CAP requires the ` compat` package for emacs < 29.1 this plays badly with the opam-based setup
- Even when `compat` is used tests are not passing on emacs <= 28.2

I guess we could use our own shims for the unavailable functions, but I am not familiar enough with emacs, elisp or melpa to fix all these issues in a timely manner, if anyone want to step in and have a look this will be much welcome. Meanwhile I think reverting the new completion changes is the best thing to do. I will open a draft PR with all these changes so that anyone willing to tackle the task will have a fresh base to build from.

tmcgilchrist:
@voodoos Ideally we can choose the correct implementation for completion by detecting the Emacs version and loading that. I'll see if I can make that work. 

What does the `opam-based setup` actually do? Can we change the instructions or Elisp it adds to explicitly `(require 'compat)`?

bbatsov:
> @voodoos Ideally we can choose the correct implementation for completion by detecting the Emacs version and loading that. I'll see if I can make that work.

I think that'd be an overkill, given the it will likely be easier to fix the updated completion on older Emacsen. 

I agree with @voodoos that it would probably be best to revert the updated completion for a bit, so users are not impacted by the current breakage. In the mean time we can setup a proper test matrix, etc - if we can one the issues wouldn't have been caught so late. 

One more note - probably it doesn't make sense to support anything older than Emacs 27 at this point, as Emacs 26 was released a long time ago and pretty much no one uses it at this point.

voodoos:
> What does the `opam-based setup` actually do? Can we change the instructions or Elisp it adds to explicitly `(require 'compat)`?

The opam-based setup modifies the user's `.emacs` file to load an elisp file that will discover opam packages that provide emacs plugins like merlin and tuareg and load the appropriate files. It will do minimal modifications in an existing `.emacs` file or create a new `.emacs` file from a template if there is none.

It expects that the plugins are "standalone" and does not configure any package manager in Emacs such as Elpa. It won't pull additional packages required by the plugins (such as compat). We might be able to vendor them as an alternative, or replace it by our own compatibility functions.

> One more note - probably it doesn't make sense to support anything older than Emacs 27 at this point, as Emacs 26 was released a long time ago and pretty much no one uses it at this point.

Noted!

I would like to do a release soonish, so I will revert the changes and create a branch with the merged changes plus the content of this PR so that we can continue working on this. If it's ready before I release, we can re-merge it.

purcell:
>  it doesn't make sense to support anything older than Emacs 27 at this point

Yep, agree with this. And then perhaps you can get away without needing _any_ compatibility functions. `package-lint` will mostly tell you if the code is using functions/features that aren't available in the minimum Emacs version the package claims to support.

What's the issue with completion in Emacs < 28.2 though? `completion-at-point` has been around for a long time so it should be possible to write a backend that works with all Emacsen >= 27.

bbatsov:
@purcell One of the new tests fails on Emacs 28, but I'm not sure why. Likely the code is using something that behaves differently on Emacs 29, but I doubt it's something hard to fix.

voodoos:
Let's move further discussions to the new PR: https://github.com/ocaml/merlin/pull/1827

PR #1827, opened 2024-09-18
Re: Rewrite merlin-completion-at-point integration to be faster and better

voodoos:
The first merge tentative of @catern rewrite of completion-at-point (CAP) lead us to discover a few issues that need fixing before we release it. I reverted the commits and this PR reintroduces changes made by @tmcgilchrist and myself to move the tests in a separate file, fix linting issues and enable testing on version 27 to 29.

There are two issues with the current status of this PR:
- Event when using the compat module tests do not pass on Emacs 27 and 28.
- Using the `compat` module breaks the assumption that packages are standalone made by the `opam user-setup` method of installation.

This last issue could be fixed by: 1. shipping the compat package with Merlin or 2. Rewriting our own replacement for the missing features.


.github/workflows/emacs-lint.yml
```diff
@@ -15,20 +15,40 @@ jobs:
...
           - snapshot
       fail-fast: false  # don't stop jobs if one fails
     env:
       EMACS_PACKAGE_LINT_IGNORE: ${{ matrix.package_lint_ignore }}
       EMACS_BYTECOMP_WARN_IGNORE: ${{ matrix.bytecomp_warn_ignore }}
     steps:
-    - uses: purcell/setup-emacs@v6.0
-      with:
-        version: ${{ matrix.emacs_version }}
+      - uses: purcell/setup-emacs@v6.0
```
purcell, line 32:
```suggestion
      - uses: purcell/setup-emacs@master
```
I usually recommend that people not pin this to a specific version, as the underlying Emacs (and Nix installer) versions change frequently.

emacs/check.sh
```diff
@@ -3,12 +3,12 @@
...
 EMACS="${EMACS:=emacs}"
 
-NEEDED_PACKAGES="package-lint company iedit auto-complete"
+NEEDED_PACKAGES="package-lint company iedit auto-complete compat"
 
 ELS_TO_CHECK=*.el
 # To reduce the amount of false positives we only package-lint files
 # that are actual installable packages.
-PKGS_TO_CHECK="merlin.el merlin-ac.el merlin-company.el merlin-iedit.el"
+PKGS_TO_CHECK="merlin.el merlin-ac.el merlin-company.el merlin-iedit.el merlin-cap.el tests/merlin-cap-test.el"
```
purcell, line 11: I don't think this is right any more. From a packaging point of view, `merlin`, `merlin-ac`, `merlin-company` and `merlin-iedit` are separate packages, because their dependencies are different. Each is built into a separate MELPA package, for granular installation.

However, presumably the intention is that `merlin-cap` will be part of the base `merlin` package, since it doesn't depend on anything opinionated that ships separately from Emacs. So it wouldn't be appropriate to run `package-lint` on `merlin-cap` separately: instead, run `package-lint` on both `merlin-cap.el` and `merlin.el` together, with the `package-lint-main-file` variable set to `merlin.el`.

And `merlin-cap-test.el` doesn't need package-linting, because it's not shipped in the package.

emacs/merlin-xref.el
```diff
@@ -1,4 +1,22 @@
...
+
+;; Licensed under the MIT license.
+
+;; Author: Simon Castellan <simon.castellan(_)iuwt.fr>
+;;         Frédéric Bour <frederic.bour(_)lakaban.net>
+;;         Thomas Refis <thomas.refis(_)gmail.com>
+;; Created: 15 May 2015
+;; Version: 0.1
+;; Keywords: ocaml languages
+;; Package-Requires: ((emacs "25.1") (xref))
```
purcell, line 11:
```suggestion
```
Don't put package metadata (e.g. `Package-Requires`) in this file: it's going to be part of the `merlin` base package, so only the `Package-Requires` header in `merlin.el` is relevant.

emacs/tests/merlin-cap-test.el
```diff
@@ -0,0 +1,252 @@
...
+;; Licensed under the MIT license.
+
+;; Author: Simon Castellan <simon.castellan(_)iuwt.fr>
+;;         Frédéric Bour <frederic.bour(_)lakaban.net>
+;;         Thomas Refis <thomas.refis(_)gmail.com>
+;;         Tim McGilchrist <timmcgil@gmail.com>
+;; Created: 13 Sep 2024
+;; Version: 0.1
+;; Keywords: ocaml languages
+;; Package-Requires: ((emacs "25.1"))
```
purcell, line 11:
```suggestion
```

purcell (commented):
Couple of comments re. the packaging aspects

emacs/merlin-cap.el
```diff
@@ -7,67 +7,484 @@
...
+  "The IGNORE-REGION argument to `merlin-cap--complete-prefix' for this request.")
+
+(defun merlin-cap--pending-request (position ignore-region)
+  "Return a pre-existing request for POSITION and IGNORE-REGION.
+
+Specifically, return a cons whose car is the prefix that the
+request was for and whose cdr is the request buffer.  Return nil
+if no pending request exists, or its inputs do not match."
+  (let ((buf (current-buffer)))
+    (with-current-buffer (get-buffer-create " *merlin-async-proc-stdout*" t)
```
tmcgilchrist, line 93: `get-buffer-create` changed the number of arguments between 27 [*]() and 28 [**]() which is causing linting issues.

[1]: https://github.com/emacs-mirror/emacs/blob/emacs-27.2/src/buffer.c#L509-L518
[2]: https://github.com/emacs-mirror/emacs/blob/emacs-28.2/src/buffer.c#L532-L546

emacs/tests/merlin-cap-test.el
```diff
@@ -0,0 +1,252 @@
...
+                 '("aa." "aa.Bb.Cc." "x" ""))))
+
+(defvar-local messages-buffer-name "*Messages*")
+
+(defun merlin-cap--current-message ()
+  "Like `current-message' but work in batch mode and use `messages-buffer-name'."
+  (with-current-buffer messages-buffer-name
+    (save-excursion
+      (forward-line -1)
+      (buffer-substring (point) (line-end-position)))))
```
tmcgilchrist, line 81: I changed `pos-eol` to `line-end-position` this might not be necessary with `compat` package? 
```suggestion
      (buffer-substring (point) (pos-eol)))))
```

emacs/check.sh
voodoos: Thanks, I did that in [a72fd55](https://github.com/ocaml/merlin/pull/1827/commits/a72fd55988ba1c5a00915668e247238adc469295).

There a still spurious `merlin-cap.el: error: You should depend on (emacs "28.1") if you need `string-search'.` even if we use the compat module.


```
merlin.el:477:0: error: `merlin/unmake-point' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
merlin.el:608:0: error: `merlin/call' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
merlin.el:1194:0: error: `merlin/display-in-type-buffer' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
merlin-cap.el:10:0: error: Package-Requires outside the main file have no effect.
merlin-cap.el:341:13: error: You should depend on (emacs "28.1") if you need `string-search'.
merlin-cap.el:356:19: error: You should depend on (emacs "28.1") if you need `string-search'.
merlin-cap.el:376:18: error: You should depend on (emacs "28.1") if you need `string-search'.
```

.github/workflows/emacs-lint.yml
```diff
@@ -15,20 +15,40 @@ jobs:
...
       matrix:
+        os:
+          - ubuntu-latest
+        ocaml-compiler:
+          - 5.2.x
         emacs_version:
           - '27.2'
           - '28.2'
-          - '29.1'
+          - '29.3'
```
tmcgilchrist, line 25:
```suggestion
          - '29.4'
```

emacs/check.sh
purcell: Ah, I was overdue to tag a new `package-lint` release, so this spurious message was due to the build picking up an old version. Once MELPA has rebuilt the new `package-lint` version this complaint should go away.

PR #1828, opened 2024-09-18
Search by type feature, a kind of sherlodoc in Merlin

xvw:
This patch introduces a kind of [Sherlodoc](https://doc.sherlocode.com/) in Merlin!

## Feature added

Adds to merlin's protocol the `search-by-type` command, which takes a query (ie: `'a list list -> ’a list`) and returns a list of possible candidates in the current environment. The structure of the result is as follows: 

```json
{
  "name": "string (the path of the candidate function)",
  "type": "string (the type of value)",
  "cost": "int (the computed cost of the result, for additional sorting on client side)",
  "doc": "string (the docstring of the value)",
  ...
}
```
The result also include the location (using `with_location`) of the `value` (useful to, from a search list, quickly jump to definition or look for usages using `project-wide-occurences`).

### Comparison with `polarity-search`

For the moment, I find that 100% of experiments give more convincing results with type search (which supports type parameters and functions, unlike polarity search) and in many case, I think that the query is more natural to write. For example:

| desired value | Polarity query | By Type query | Comment |
| --- | --- | --- | --- |
| `int_of_string` | `-string +int` | `string -> int` | Almost the same query (and almost the same result) |
| `int_of_string_opt` | `-string +option` | `string -> int option` | Search by type is more precise |
| `List.flatten` | `-list +list` | `'a list list -> 'a list` | Search by type is more precise |

There is also a test-suite that make a comparison for the same expectation between `by type` and `by polarity`.

Here is some direct examples:
 
#### Search by polarity

polarity-search-1

#### Search by type

search-by-type-1

### Difference with Sherlodoc

At present, Sherlodoc is much more intelligent when it comes to indexing, which is not at all within the scope of this PR, which uses the same environment as that used for polarity search. In addition, Sherlodoc is better at comparing names and this PR distinguishes polymorphic variables (but both approach deal relatively properly with equivalence modulo isomorphism). In practice, this distinction is expressed as follows: `'foo list -> ('foo -> 'bar) -> 'bar list` is treated by sherlodoc where every variables are `Poly` and the following PR use indexes (so we can distinguish `fst` and `snd` for example).

## Client support

Currently, **only the suppport for emacs** has been added and uses the same output as the polarity search result. The emacs client introduces 2 new commands:

- `merlin-search-by-polarity` (that replace `merlin-search`)
- `merlin-search-by-type` (that perform a searchby type)

And `merlin-search` was changed to switch between `search-by-type` or `search-by-polarity` (depending on the query, so if the query starts by `-` or `+`, it arbitrary decide that this is a polarity query).

## About the code base

The core of the feature is isolated inside a dedicated sub library : `sherlodoc` and has no dependencies and just deal with "_computing distances between queries and values_". The PR also introduce `unit-test` (using Alcotest) and `ocamlformat`, currently only available for `src/sherldoc` and `tests/test-units`.

## Further work

Currently, I think that the PR is an huge improvement, because it leads (IMHO) in every case to a better result set than the polarity search but we can have improvement:
- Support for first-class modules
- dealing with row-polymorphism (object and polymorphic variant)
- dealing with labels
- better client support (that should be part of the implementation of a decent support of OCaml LSP from Emacs and Vim)

But I think we can improve progressively the feature according to user reviews.

## More visual examples

search-by-type-for-map

Here we can see that `'foo list -> ('foo -> 'bar) -> 'bar list` is valid for `map: ('a -> 'b) -> 'a list -> 'b list`!

search-fst-snd

Here, we can see that `fst` and `snd` are properly ordered!


xvw:
cc @art-w

xvw:
The behaviour of the feature is very strange on Windows :/

xvw:
@voodoos :
- https://github.com/ocaml/merlin/pull/1828/commits/d7094d6ffe48f06746ccce267807bedbd8000724 generated the value with holes (to be constructed)
- https://github.com/ocaml/merlin/pull/1828/commits/5d4125ccbe70e4dec8e7298233bb8e5926e6b9a7 make the command `search-by-type` relay on polarity query if the query looks like a polarity-one
- https://github.com/ocaml/merlin/pull/1828/commits/8deb0fccbec3de1a15e72da3be23aa56382a078e use a better buffer to display results (allowing search inside the result and generating the expression with holes)

aa-final-search

src/analysis/type_search.ml
```diff
@@ -0,0 +1,155 @@
...
+            |> doc_to_option
+          in
+          let constructible = make_constructible path d in
+          (cost, path, desc, doc, constructible) :: acc
+      ) dir env acc
+  in
+  let rec walk acc (T (_, dir, children)) =
+    let force () =
+      let _ = Env.find_module_by_name dir env in
+      Lazy.force children
```
art-w, line 136: I've been asked to review and could only find this nitpick: Perhaps the side-effecting `Env.find_module_by_name` function call could be moved to `make_trie`'s `walk` to avoid incorrect access of the lazy `children`? It's not entirely clear to me why a lazy trie needs to be built since it will be fully traversed? The code also looks inspired by `polarity_search.ml` similar trie, so perhaps this way of traversing the environment could be shared?

(Sorry this turned into a chain of thoughts as I'm still discovering Merlin's codebase, feel free to ignore as there could be excellent reasons to do things that way!)

art-w:
In case it helps others review this PR, the general algorithm implemented is to:
- Traverse the environment to score every value against the query, and filter out values that don't match
- Sort the matching values by their score to output the best results first

The scoring uses an algorithm similar to sherlodoc to compare types, with enhancements by @xvw that I'm going to steal.
- The type ASTs of the query and the tested value are broken into a set of paths from root to leaves: for example `int list -> string * 'a` becomes the three paths `{ [Left_arrow; list; int], [Right_arrow; Product (0,2); string], [Right_arrow; Product (1,2); Poly 0 }`.
- Then each of the N value type-paths is scored against each of the M query type-paths with levenshtein distance, using rules specific to types to e.g. check that terms have the same polarity to avoid matching an argument with an output. We get an NxM integer matrix out with some cells containing `0` for exact matches, while others may have `max_distance` to signal strong incompatibility (e.g. if one path is `[string]` and the other `[char]`).
- A stable marriage algorithm is then applied to the matrix to select exactly one cell per column and row such that the sum of selected cells is minimal, which correspond to finding how to make the value and query types most similar up to reordering or extra things in the value type (but not in the query type, as that would indicate the value type is missing stuff requested by the user).

While this doesn't attempt to solve the much harder type isomorphism problem, it does rank types by similarity by preferring types that requires the less modifications to match the query, while tolerating user mistakes. (If the previous explanations are incomprehensible, there's an even worst description in French with drawings at around 21:00-23:00 in [this presentation](https://www.irill.org/videos/OUPS/2023-03/wendling.html))

@xvw added string levenshtein distance to compare type names while sherlodoc only checks for substring match. His normalization of the type ASTs is a really smart and simple way to handle polymorphic variables! He's also much more principled in his scoring, while sherlodoc had to resort to magic constants to produce results in a nice order... I have yet to test this PR to evaluate if the ordering feels right, but tuning the heuristic constants is highly subjective anyway :) 

The type scoring algorithm can be very expensive. If you run into performance issues on large environments, sherlodoc has an additional trick that might be useful in a follow-up PR: The values are pre-filtered to remove obviously bad candidates using the simpler `polarity_search` matching, and pre-sorted with a heuristic ~= name length + type length, as visually short results are preferable. The costly type scoring is then performed on those best candidates, with a cut-off once enough have been scored such that we can assume the remaining untested candidates will have a terrible score and won't make it into the top results shown to users.

Anyway, I'm super hyped by this PR, it will be excellent to not leave my editor while also getting better results specific to my codebase!

src/analysis/type_search.ml
xvw: Hi @art-w, on https://github.com/ocaml/merlin/pull/1828/commits/87eeea0154bd722bd8b4e35bdff50fc43bafdbf1 I've rewritten the collection of environment values without using the Lazy trie. I haven't noticed any slowdown in use! Thanks for the suggestion!

I haven't changed the polarity search code (because it seems to me that the PR is already big enough ;) )

xvw:
Thank you very much for your feedback and for helping us with future reviews!

> The scoring uses an algorithm similar to sherlodoc to compare types, with enhancements by @xvw that I'm going to steal.

Of course! When I've got a bit more time, I'll try to make a small pass at at least integrating the polymorphic variable distinction (and potentially try to implement what I've got in mind for unsupported forms of type).

> @xvw added string levenshtein distance to compare type names while sherlodoc only checks for substring match. His normalization of the type ASTs is a really smart and simple way to handle polymorphic variables! He's also much more principled in his scoring, while sherlodoc had to resort to magic constants to produce results in a nice order... I have yet to test this PR to evaluate if the ordering feels right, but tuning the heuristic constants is highly subjective anyway :)

I don't know what the best approach is yet, but if I ever get any user feedback complaining about the search by term, I'll try out your approach.

> The type scoring algorithm can be very expensive. If you run into performance issues on large environments, sherlodoc has an additional trick that might be useful in a follow-up PR

As this patch is not based on indexing and does not allow a suffix tree to be expressed, as is the case with Sherlodoc, no pre-optimisation phase has been carried out for the time being. If users ever complain about performance, I'll do some experiments with filters, in the environment, by polarity! (Unlike Sherlodoc, which looks at **all OPAM packets**, here the lookup is only run on the environment, so the number of values is much smaller.)

---

On the whole, I agree with the points you raise! And I'll keep all that in mind when we give our user feedback. For the moment, I think we need to move forward and collect user feedback to refine the heuristics and optimise (avoiding premature optimisations).

art-w (approved):

voodoos:
I guess the Ci faisl on Windows because the standard libraries are slightly different. 
You could try to make the tests compatible, but don't worry too much about it, we can also simply disable them.

xvw:
> I guess the Ci faisl on Windows because the standard libraries are slightly different. You could try to make the tests compatible, but don't worry too much about it, we can also simply disable them.

Is there a straightforward way to do it (make tests compatible)?

voodoos:
> > I guess the Ci faisl on Windows because the standard libraries are slightly different. You could try to make the tests compatible, but don't worry too much about it, we can also simply disable them.
> 
> Is there a straightforward way to do it (make tests compatible)?

No. It depends on the test :-) Here I would imagine that you could filter out results that are not in both sides ? But maybe not :-)

voodoos:
Also, could you update `https://github.com/ocaml/merlin/blob/main/doc/dev/PROTOCOL.md` and maybe describe the query language here (or in a dedicated wiki page) ?

xvw:
> Also, could you update `https://github.com/ocaml/merlin/blob/main/doc/dev/PROTOCOL.md` and maybe describe the query language here (or in a dedicated wiki page) ?

Done in https://github.com/ocaml/merlin/pull/1828/commits/e47a4880cb2266965e3ae5c480fb06352ada191e

voodoos (commented):
I still have to look at the distance calculations but this looks good so far 👍

src/analysis/type_search.ml
```diff
@@ -0,0 +1,148 @@
...
+  the authors or copyright holders be liable for any claim, damages or other
+  liability, whether in an action of contract, tort or otherwise, arising
+  from, out of or in connection with the software or the use or other dealings
+  in the Software.
+
+   )* }}} *)
+
+open Std
+
+let type_of env typ =
```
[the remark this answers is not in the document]
xvw: Moving it in `type_parsed` will make the library depending on OCaml stuff (which is a little bit sad no?)

src/analysis/type_search.ml
```diff
@@ -0,0 +1,148 @@
...
+      and with_label l =
+        match l with
+        | Ocaml_parsing.Asttypes.Nolabel -> " _"
+        | Labelled s -> " ~" ^ s ^ ":_"
+        | Optional _ -> ""
+      in
+      aux (with_label l) b
+    | _ -> ""
+  in
+  path ^ holes
```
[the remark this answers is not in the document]
xvw: Construct is to smart for that but I can add a `todo` for investigating the usage of `construct`?

xvw:
@voodoos 
Comments are adressed here https://github.com/ocaml/merlin/pull/1828/commits/7cf0b85a6fd44faddb100b9d25ee822e025be846

Thanks for the first round of the review!

src/analysis/polarity_search.ml
```diff
@@ -138,3 +151,57 @@ let execute_query query env dirs =
...
+          let name = Format.asprintf "%a" Printtyp.path path in
+          let doc =
+            Locate.get_doc
+              ~config
+              ~env
+              ~local_defs
+              ~comments
+              ~pos
+              (`User_input name)
+            |> Type_search.doc_to_option
```
voodoos, line 179: I am wondering about the cost of fetching the doc of all the returned items. It's not a trivial operation. Maybe we should provide an option to the command so that clients can opt-out from that. (I put the comment here but of course it is also valable for the new type-search results

src/sherlodoc/type_distance.ml
```diff
@@ -0,0 +1,188 @@
...
+     implied, including but not limited to the warranties of merchantability,
+     fitness for a particular purpose and noninfringement. In no event shall
+     the authors or copyright holders be liable for any claim, damages or other
+     liability, whether in an action of contract, tort or otherwise, arising
+     from, out of or in connection with the software or the use or other dealings
+     in the Software.
+
+   )* }}} *)
+
+type step =
```
voodoos, line 29: Maybe this module would be a good place to put the relevant parts of Arthur's comment https://github.com/ocaml/merlin/pull/1828 ?

voodoos (approved):
This looks very good. My only worries are related to performances, I think we should make documentation fetching optional.

src/analysis/polarity_search.ml
xvw: What do you think about the approach proposed here https://github.com/ocaml/merlin/pull/1828/commits/a8688aaacb24b15b4cec612915d4a018d918c33c

CHANGES.md
```diff
@@ -8,10 +8,14 @@ unreleased
     - destruct: Refinement in the presence of optional arguments (#1800 #1807, fixes #1770)
     - Implement new expand-node command for expanding PPX annotations (#1745)
     - Implement new inlay-hints command for adding hints on a sourcetree (#1812)
+	- Implement new search-by-type command for searching values by types (#1828)
```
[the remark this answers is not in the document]
xvw: Fixed in https://github.com/ocaml/merlin/pull/1828/commits/a8688aaacb24b15b4cec612915d4a018d918c33c

.ocamlformat
```diff
@@ -0,0 +1,3 @@
+disable=true
+parse-docstrings
+break-cases=fit-or-vertical
```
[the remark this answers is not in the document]
xvw: Removed in https://github.com/ocaml/merlin/pull/1828/commits/a8688aaacb24b15b4cec612915d4a018d918c33c

tests/test-dirs/search-by-type-comparison-to-polarity-search.t
```diff
@@ -132,17 +132,17 @@ map).
...
   }
   {
-    "name": "ListLabels.concat_map",
-    "type": "f:('a -> 'b list) -> 'a list -> 'b list"
+    "name": "List.concat_map",
+    "type": "('a -> 'b list) -> 'a list -> 'b list"
+  }
+  {
+    "name": "ListLabels.filter_map",
+    "type": "f:('a -> 'b option) -> 'a list -> 'b list"
```
voodoos, line 145: Why did this test result change ?

voodoos:
La CI macos a aussi un problème:

```diff
# File "tests/test-dirs/search-by-type-comparison-to-polarity-search.t", line 1, characters 0-0:
# /opt/homebrew/bin/git --no-pager diff --no-index --color=always -u _build/.sandbox/1535e15bb85fec87f6431da9c6ded6e4/default/tests/test-dirs/search-by-type-comparison-to-polarity-search.t _build/.sandbox/1535e15bb85fec87f6431da9c6ded6e4/default/tests/test-dirs/search-by-type-comparison-to-polarity-search.t.corrected
# diff --git a/_build/.sandbox/1535e15bb85fec87f6431da9c6ded6e4/default/tests/test-dirs/search-by-type-comparison-to-polarity-search.t b/_build/.sandbox/1535e15bb85fec87f6431da9c6ded6e4/default/tests/test-dirs/search-by-type-comparison-to-polarity-search.t.corrected
# index dc5aa5f..0ac9c7d 100644
# --- a/_build/.sandbox/1535e15bb85fec87f6431da9c6ded6e4/default/tests/test-dirs/search-by-type-comparison-to-polarity-search.t
# +++ b/_build/.sandbox/1535e15bb85fec87f6431da9c6ded6e4/default/tests/test-dirs/search-by-type-comparison-to-polarity-search.t.corrected
# @@ -37,11 +37,11 @@ potential failures, so lifting the result in an int option).
#      "type": "string -> bool option"
#    }
#    {
# -    "name": "Float.of_string_opt",
# +    "name": "float_of_string_opt",
#      "type": "string -> float option"
#    }
#    {
# -    "name": "float_of_string_opt",
# +    "name": "Float.of_string_opt",
#      "type": "string -> float option"
#    }
```

tests/test-dirs/search-by-type-comparison-to-polarity-search.t
xvw: Because doc was removed (and it change the result)

xvw:
Rebased and formatted

voodoos (approved):
Great work @xvw !

As we discussed, what this feature needs now is user feedback. Let's merge it 🙂

PR #1830, opened 2024-09-19
Interaction between Search by types and Construct

xvw:
This patch (on top of #1828) introduce an interactive way to search directly from a buffer.
- from the Merlin protocol, search introduces a new entry in the resultset : `constructible` which is a string that represent the function filled with holes (so it can be used in every other client)
- A reimplementation of `merlin-search` that ask for a query if nothing is selected but can use the region as a query and replace the region with the function and holes. Demonstration below

construct-by-type


xvw:
(After a small chat with @voodoos, the UI seems a little bit controversial so I close it and improve the previous one!)

PR #1838, opened 2024-09-25
Fix ignorace of `SOURCE_ROOT` directive

liam923:
At the moment, the `SOURCE_ROOT` directive is not properly read from `.merlin` files. This PR fixes the issue by ensuring the `dot_merlin_reader.ml` passes on the directive.


liam923:
It seems I was looking at a stale version of `main` - this has already been fixed.

liam923:
It seems I got confused by a merge conflict and this was not actually fixed. I'll open another PR.

liam923:
Re-opening

PR #1839, opened 2024-09-25
Fix ignorace of SOURCE_ROOT directive

liam923:
At the moment, the `SOURCE_ROOT` directive is not properly read from .merlin files. This PR fixes the issue by ensuring the `dot_merlin_reader.ml` passes on the directive.


voodoos (approved):
Thank you for the fix @liam923 . We really need to test dot-merlin-reader more extensively in the future since most users use the dune configuration reader.

PR #1840, opened 2024-09-25
Canonicalize paths when finding occurrences

liam923:
This fixes a minor bug in occurrences. See the test in the first commit (5fb1560f522617ef4d80f30166dda5ae84561a90), which demonstrates the issue.


xvw (approved):
(And thanks for the test suite, it was illustrative!)

voodoos (approved):
Thanks, looks good to me

voodoos (approved):

PR #1846, opened 2024-09-27
vim: Expose the search-by-type feature

Julow:
This adds the `:MerlinSearchType` command to query for a value with a matching type. It works like the previous `:MerlinSearch` but uses the new search feature added in https://github.com/ocaml/merlin/pull/1828, the search results are shown in the completion menu.

`:MerlinSearch` switches between polarity search and search by type depending on the first character of the query, like is done in the emacs plugin.

Documentation about the three search functions is added.

### DIFF
--- a/vim/merlin/autoload/merlin.py
+++ b/vim/merlin/autoload/merlin.py
@@ -90,7 +90,7 @@ def vim_value(v):
     if isinstance(v, int):
         return str(v)
     if isinstance(v, str):
-        return "'%s'" % v.replace("'", "''")
+        return "'%s'" % v.replace("'", "''").replace("\n", "'.\"\\n\".'")
     raise Exception("Failed to convert into a vim value: %s" % str(v))
 
 # Format a dictionnary containing integer and string values into a Vim record.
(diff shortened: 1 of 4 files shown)


xvw (approved):
Thanks a lot!

vim/merlin/autoload/merlin.py
```diff
@@ -365,15 +365,15 @@ def command_holes():
...
-    return re.sub(re_spaces_around_nl, "\n", re.sub(re_spaces, " ", str)).replace("'", "''").strip()
-
+# Turns Merlin's search-by-polarity or complete-prefix entries into a Vim's completion-item list
 def vim_fillentries(entries, vimvar):
-    prep = vim_complete_prepare
-    prep_nl = vim_complete_prepare_preserve_newlines
     for prop in entries:
-        vim.command("let tmp = {'word':'%s','menu':'%s','info':'%s','kind':'%s'}" %
-                (prep(prop['name']),prep(prop['desc']),prep_nl(prop['info']),prep(prop['kind'][:1])))
+        vim.command("let tmp = " + vim_record({
```
voodoos, line 371: It's unclear to me that vim_record behave similarly as the previous  use of `prep` and `prep_nl`. Does it "preserve newlines" ? And only for the "info" field ?

vim/merlin/autoload/merlin.vim
```diff
@@ -284,15 +284,33 @@ function! merlin#PolaritySearch(debug,query)
...
+
+" Do a polarity search or a search by type depending on the first character of
+" the query.
+function! merlin#Search(debug, query)
+  if a:query =~ "^[-+]"
+    call merlin#PolaritySearch(a:debug, a:query)
+  else
+    call merlin#SearchByType(a:debug, a:query)
+  endif
+endfunction
```
voodoos, line 312: @xvw this detection is also made on Merlin side isn't it ?
Maybe we could have a simpler interface in both emacs and vim:

- Keep the existing PolaritySearch
- Add a new SearchByType or Search, that does indeed also support the polarity search syntax.

voodoos (commented):
Thanks @Julow !

vim/merlin/autoload/merlin.vim
xvw: Yes, in Emacs I've moved the original `search` behaviour into a `search-by-polarity` command and then the `search` command relies on the merlin `search-by-type` command (which classifies the query)

vim/merlin/autoload/merlin.py
Julow: You are right, there was a problem here!
I've fixed `vim_record` to properly encode newlines and added back the `prep` and `prep_nl` functions. I'm not sure in which cases these are useful.
I've just learnt from this "info" feature and I find it awesome. I tested that completion works fine with these settings:
```
let g:merlin_completion_with_doc = 1
set completeopt+=popup
```

voodoos:
Oh, I completely forgot about this one, sorry. @Julow is it ready ? 🙂

voodoos:
Thanks !

PR #1850, opened 2024-10-03
Add support for 5.3.0

xvw:


src/ocaml/typing/printtyp.ml
```diff
@@ -15,2664 +15,66 @@
...
 
-let shorten_class_type_path env p =
-  wrap_printing_env env
-    (fun () -> best_class_type_path_simple p)
 
-let () =
-  Env.shorten_module_path := shorten_module_path
+let () = Env.print_longident := Doc.longident
+let () = Env.print_path := Doc.path
+let () = Btype.print_raw := Doc.raw_type_expr
```
Octachron, line 79: This part looks weird: the debugging printers now lives in Rawprinttyp which avoids the cyclic dependency on Btype. Normally, it should be doable to remove all raw functions in Printtyp and Btype.

voodoos: @xvw could you explain again what was your strategy with merging the printing changes ?

voodoos: @Octachron the print_raw forward declaration in btype was actually unused, and removed in upstream so I reflected that change in [13407ca](https://github.com/ocaml/merlin/pull/1850/commits/13407cad0047e4458c7efc5681877cffa05dc663). 

Was your comment also referring to the others forward declarations in Env ?

PR #1854, opened 2024-10-07
Fix `EXCLUDE_QUERY_DIR` for cmt files

liam923:
When looking for `.cmt` files, Merlin doesn't currently respect the `EXCLUDE_QUERY_DIR` directive (see the first commit for a test case demonstrating the issue). This PR resolves this issue.

This PR corresponds to [PR 104 on Merlin-JST](https://github.com/janestreet/merlin-jst/pull/104)

### DIFF
--- a/src/kernel/mconfig.ml
+++ b/src/kernel/mconfig.ml
@@ -814,7 +814,9 @@ let cmt_path config =
   let stdlib = stdlib config in
   let exp_dirs = List.map ~f:(Misc.expand_directory stdlib) dirs in
   let stdlib = if config.ocaml.no_std_include then [] else [ stdlib ] in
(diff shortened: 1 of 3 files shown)


voodoos (approved):
Thanks Liam the changes look good to me. I wonder whether we whould de-duplicate the result similarly ar what is done for the  `build_path` ?

Could you add a changelog entry and filter the test output so that the path is not printed ?

liam923:
> Thanks Liam the changes look good to me. I wonder whether we whould de-duplicate the result similarly ar what is done for the `build_path` ?

Good point. Actually at this point `build_path` and `cmt_path` are nearly identical, so I pulled their functionality out into a common function.

voodoos (approved):

Issue #1113, opened 2020-03-09
`Not_found` during polarity search and missing documentation


arvidj:
I'm a OCaml beginner. I'm used to tools like Hoogle or Yahoo. I'm looking for a function of some type `t`. I'm hoping merlin can help me. I'm using Emacs. I try the command `merlin-search`. It is undocumented. I try any string, the response is:

```
Not_found
Raised at file \"src/ocaml/typing/407/ident.ml\", line 171, characters 6-21
Called from file \"src/ocaml/typing/407/env.ml\", line 281, characters 25-57
Re-raised at file \"src/ocaml/typing/407/env.ml\", line 301, characters 12-21
Called from file \"src/ocaml/typing/407/env.ml\" (inlined), line 1261, characters 2-57
Called from file \"src/ocaml/typing/407/env.ml\", line 1319, characters 26-56
Called from file \"src/analysis/polarity_search.ml\", line 83, characters 29-52
Called from file \"src/utils/std.ml\", line 109, characters 12-15
Called from file \"src/analysis/polarity_search.ml\", line 86, characters 17-62
Called from file \"src/frontend/query_commands.ml\", line 417, characters 6-125
Called from file \"src/frontend/ocamlmerlin/new/new_commands.ml\", line 64, characters 15-53
Called from file \"src/utils/std.ml\", line 679, characters 8-12
Re-raised at file \"src/utils/std.ml\", line 681, characters 30-39
Called from file \"src/ocaml/utils/misc.ml\", line 62, characters 8-15
Re-raised at file \"src/ocaml/utils/misc.ml\", line 79, characters 10-24
Called from file \"src/utils/local_store.ml\", line 42, characters 8-12
Re-raised at file \"src/utils/local_store.ml\", line 52, characters 4-15
Called from file \"src/kernel/mocaml.ml\", line 18, characters 8-38
Re-raised at file \"src/kernel/mocaml.ml\", line 20, characters 42-53
Called from file \"src/frontend/ocamlmerlin/new/new_merlin.ml\", line 100, characters 14-110
```

Okay. I'm trying to figure what `merlin-search` is should do. I see that it does something called "polarity search". I find the documenation : 

```
### `search-by-polarity -position <position> -query <string>`

    -position <position>  Position to complete
         -query <string>  Query of the form TODO

search-by-polarity -position pos -query ident
        TODO

```

I'm none the wiser. 

My Merlin version 3.3.3 : 

```
ocamlmerlin -version
[...]

PR #1858, opened 2024-10-24
Fix 1113

xvw:
The patch 
- fix the exception related in #1113 
- add clarification about polarity queries


voodoos (approved):
Thanks for the quick fix Xavier.

PR #1861, opened 2024-10-30
Upgrade to OCamlformat 0.27.0

Julow:
The aim of this preview is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,4 @@
 # git config blame.ignoreRevsFile .git-blame-ignore-revs
 40e0cdb95ff31b81212109fada6e0ff93b6eefee
+# Upgrade to OCamlformat 0.27.0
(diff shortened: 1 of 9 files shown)


voodoos:
It looks like previously unformatted files are formatted. Is the due to the way you run the preview ?

Julow:
I do `dune build @fmt --auto-promote` using dune 3.15.0 and I see these files being reformatted. Am I missing something ?

voodoos:
Ah I see. It's an issue with Dune (or our config) that rebuilds the parser unexpectedly. Since your installed version of Menhir is not the same as the one the committed files are, there is a diff (which is unrelated to ocamlformat). We can simply ignore it for the purpose of the preview.

voodoos:
Done directly on main, thanks Jules !

PR #1885, opened 2025-01-10
Report stale occurrences

liam923:
This PR does two things:
1. Fix a bug with Merlin's detection of stale occurrences. Currently, depending on the layout of a repository, Merlin may not line up file paths correctly.
2. Currently, Merlin discards any occurrences it thinks might be stale. This PR changes this so that the stale occurrences are reported, but flagged as being stale.

The net effect for an end user should be nothing - the stale occurrences should still be reported by their editor. But this enables future LSP/editor work.

The reason these two items are in the same PR is:
- Doing just (1) would result in stale occurrences no longer being reported at all.
- Doing just (2) would be hard to test.

I encourage reviewing commit-by-commit. The tests pass after each commit, and changes (1) and (2) are separated into separate commits.

### DIFF
--- a/.github/workflows/ocaml-lsp-compat.yml
+++ b/.github/workflows/ocaml-lsp-compat.yml
@@ -51,7 +51,6 @@ jobs:
 
       - name: Check that Merlin and OCaml-LSP are co-installable
         run: |
-          opam --cli=2.1 pin --with-version=dev --no-action https://github.com/liam923/ocaml-lsp.git#rename-holes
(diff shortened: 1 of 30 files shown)


voodoos (approved):
Thanks Liam, these changes look good to me.

There might be a bit of work to rebase it on main with the types changes in `Index_format`.

voodoos:
Oh I see that you already merged origin/main yesterday. I will have another look.

PR #1888, opened 2025-01-16
Rename `Texp_hole` and `Tmod_hole`

liam923:
In ocaml-flambda, we've added a new constructor to `Typedtree.expression_desc` named `Texp_hole`: https://github.com/ocaml-flambda/flambda-backend/pull/3310/files#diff-ed47e32f450beb29b2f22ff8073a5592865c82845687a224124858a6b6060fc3R504. This unfortunately conflicts with the `Texp_hole` constructor that Merlin adds to the same type in order to represent typed holes. To solve this, we are renaming what was Merlin's `Texp_hole` to `Texp_typed_hole`. For consistency, we are also renaming `Tmod_hole` to `Tmod_typed_hole`, even though there is no `Tmod_hole` in the compiler. This PR includes the same renames in order to minimize the diff with merlin-jst.

### DIFF
--- a/.github/workflows/ocaml-lsp-compat.yml
+++ b/.github/workflows/ocaml-lsp-compat.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Check that Merlin and OCaml-LSP are co-installable
         run: |
-          opam --cli=2.1 pin --with-version=dev --no-action https://github.com/voodoos/ocaml-lsp.git#merlin-503-compat
+          opam --cli=2.1 pin --with-version=dev --no-action https://github.com/liam923/ocaml-lsp.git#rename-holes
           opam --cli=2.1 pin --with-version=5.4-503 --no-action .
           opam install ocaml-lsp-server --ignore-constraints-on=ocamlformat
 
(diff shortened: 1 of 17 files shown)


liam923:
I see that the ocaml-lsp compatibility check is failing. Does that block merging this PR? If so, how would I atomically make this change to both repos?

voodoos:
> I see that the ocaml-lsp compatibility check is failing. Does that block merging this PR? If so, how would I atomically make this change to both repos?

What we usually do is open a PR on both side and update the CIs to match the other-side branche. Then we merge both when they are ready.

We try to progressively move all Typedtree usage to `merlin-lib` to prevent further these breakage which are the most common.

In that case, I think the best way forward would be to use that PR to move the `Typed_holes` compatibility module to `merlin-lib`, in the `Typedtree_utils`.

liam923:
> In that case, I think the best way forward would be to use that PR to move the `Typed_holes` compatibility module to `merlin-lib`, in the `Typedtree_utils`.

See https://github.com/ocaml/ocaml-lsp/pull/1451

voodoos (approved):
Thanks Liam, looks good to me

PR #1889, opened 2025-01-17
 Replace Marshal by Granular_marshal in ocaml-index

voodoos:
Initial PR presentation by @lucccyo:

The current implementation of ocaml-index uses Marshal to store on the disk the data.
Searching for occurrences on massive projects is time-consuming because the search loads all the data structures from the disk to perform the search.

This Pull Request aims to replace Marshal with a granular version to make the ocaml-index more efficient in reading.
It comes with two granular implementations of the data structures set and map, based on the Stdlib implementation.
During a search operation, the program lazily loads only the required part of the ocaml-index.
It works because the heavy nodes of the granular_map and granular_set have link indirections,
introducing serialization boundaries, which allows Marshal to delay the deserialization of their children.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@ unreleased
       (#1888)
     - `locate` can now disambiguate between files with identical names and contents
       (#1882)
+  + ocaml-index
+    - Improve the granularity of index reading by segmenting the marshalization
+      of the involved data-structures. (#1889)
 
 merlin 5.4.1
 ============
--- a/src/analysis/occurrences.ml
+++ b/src/analysis/occurrences.ml
@@ -113,7 +113,9 @@ end
 let get_buffer_locs result uid =
   Stamped_hashtable.fold
     (fun (uid', loc) () acc ->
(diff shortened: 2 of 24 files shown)


voodoos:
On my own testing the indexing time is a slight (5%) increase of the indexing time and no increase in the file size. Fetches are now close to instantaneous and Merlin's memory usage vastly reduced since the index are not fully loaded into memory anymore.
```
Benchmark 1: Main
  Time (mean ± σ):      1.116 s ±  0.011 s    [User: 0.941 s, System: 0.159 s]
  Range (min … max):    1.100 s …  1.132 s    10 runs

Benchmark 1: With the new granular-marshall:
  Time (mean ± σ):      1.177 s ±  0.011 s    [User: 1.004 s, System: 0.158 s]
  Range (min … max):    1.162 s …  1.195 s    10 runs
 
```
@liam923 I am going to merge these improvements make sense for most users, but we should also evaluate the impact of this PR on "larger" codebases.

PR #1891, opened 2025-01-31
Implementation to get list of possible jumps

PizieDust:
Add a new function `get_all` that returns all the possible jump targets in a code buffer.
In some client implementations, we repeatedly query the `Jump` command of Merlin which means many trips to the server from the client. This new function ensures that we only make the trip once and we get all the information we need.

cc @voodoos

### DIFF
--- a/src/analysis/jump.ml
+++ b/src/analysis/jump.ml
@@ -136,19 +136,30 @@ let find_case_pos cases pos direction =
   in
   let case = find_pos pos cases direction in
   match case with
-  | Some location -> `Found location
+  | Some location -> location
   | None -> (
     match direction with
     | Next -> raise No_next_match_case
     | Prev -> raise No_prev_match_case)
 
-let get typed_tree pos target =
+let get_enclosings typed_tree pos =
   let roots = Mbrowse.of_typedtree typed_tree in
-  let enclosings =
-    match Mbrowse.enclosing pos [ roots ] with
-    | [] -> []
-    | l -> List.map ~f:snd l
-  in
+
+  match Mbrowse.enclosing pos [ roots ] with
+  | [] -> []
+  | l -> List.map ~f:snd l
+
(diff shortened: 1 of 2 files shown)


src/analysis/jump.ml
```diff
@@ -173,18 +176,28 @@ let get typed_tree pos target =
...
-      match target with
-      | "match-next-case" -> find_case_pos (get_cases_from_match node) pos Next
-      | "match-prev-case" ->
-        find_case_pos (List.rev (get_cases_from_match node)) pos Prev
-      | _ ->
-        let node_loc = Browse_raw.node_real_loc Location.none node in
-        `Found node_loc.Location.loc_start
+      match get_node_position target pos node with
+      | Some loc -> `Found loc
+      | None -> `Error ("No matching case found for " ^ target)
```
xvw, line 181: I know that it is not particularly related to this PR but just out of curiosity, why going for a polymorphic variant? It looks like `result` is sufficient here no?

src/analysis/jump.ml
```diff
@@ -173,18 +176,28 @@ let get typed_tree pos target =
...
+  let nodes = skip_non_moving pos enclosings in
+  let results =
+    List.filter_map
+      ~f:(fun (target, pred) ->
+        match find_node [ pred ] nodes with
+        | exception No_matching_target -> None
+        | node -> (
+          match get_node_position target pos node with
+          | Some position -> Some (target, position)
+          | None -> None))
```
xvw, line 197: wdyt about 
```ocaml
| node -> 
   Option.map 
      (fun pos -> (target, pos)) 
      get_node_position target pos node)
```

src/analysis/jump.ml
PizieDust: I think we can correct this in a different PR. But I just maintained what was already used, which was the variant

PR #1892, opened 2025-02-04
Inlay hints: traverse ghost locations

mlasson:
# Description  

Consider the following code:  

```ocaml
let f = fun x -> x ^ ""
let g _ = fun x -> x ^ ""
```  

In this case, an inlay hint appears for `x` in `f`, but no hint is shown for `g`.  

In VS Code, this results in the following display:  
image  

### Issue  

The inlay hints implementation includes logic to stop traversing "ghost locations" to avoid annotating generated code. However, a simple definition like:  

```ocaml
let f x = ...
```  

is syntactic sugar for:  

```ocaml
let f = fun x -> ...
```  

where `fun x ->` introduces a ghost location. This explains why the ghost location check had to be explicitly disabled in your test.  

### Proposed Fix  

This PR modifies the traversal logic to allow ghost locations even if they originate from "generated code." Instead of blocking traversal, ghost locations are filtered when generating inlay hints.

### DIFF
--- a/src/analysis/inlay_hints.ml
+++ b/src/analysis/inlay_hints.ml
@@ -4,8 +4,6 @@ let { Logger.log } = Logger.for_section "inlay-hints"
 
 module Iterator = Ocaml_typing.Tast_iterator
 
-let is_ghost_location avoid_ghost loc = loc.Location.loc_ghost && avoid_ghost
-
 let pattern_has_constraint (type a) (pattern : a Typedtree.general_pattern) =
   List.exists
     ~f:(fun (extra, _, _) ->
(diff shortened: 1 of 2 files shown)


mlasson:
I tested my "fix" with generated code and got the following result:  
image  

This highlights why simply filtering the output isn't a viable solution. Code generators actively use locations, sometimes reusing them multiple times, which makes it difficult to distinguish between meaningful and redundant hints.

mlasson:
In commit [[9a977d7](https://github.com/ocaml/merlin/pull/1892/commits/9a977d7dcce08827285a03e4d82555aec49a3aac)](https://github.com/ocaml/merlin/pull/1892/commits/9a977d7dcce08827285a03e4d82555aec49a3aac), I reverted my previous fix and implemented a new approach.  

The new fix stops traversal for structure items with a ghost location that are likely produced by deriving code generators while still allowing traversal inside expressions with ghost locations.

mlasson:
In commit [598e863](https://github.com/ocaml/merlin/pull/1892/commits/598e8631638f6fb5602874f33b65f6aabd6f1057) I proposed yet another fix. We can rely on the merlin.hide attribute introduced by well-behaved code generators !

voodoos:
Seems reasonnable to rely on merlin.hide attributes for this indeed.

I wonder what happens in other cases were the compiler introduces ghost locations. Standard example of these are the use of punning in record's label names and function labels. Could you add such examples to your test ?

mlasson:
Closed in favor of https://github.com/ocaml/merlin/pull/1894

mlasson:
@voodoos I added the tests #1894 (but they were already broken even when we were not checking for ghost locations)

PR #1894, opened 2025-02-05
Improve Inlay Hints Handling with [@merlin.hide]

mlasson:
# Description  

Previously, inlay hints logic relied on ghost locations to avoid annotating generated code. But it lead to some inconsistencies because ghost locations are used in non-generated code, for instance: 

```ocaml
let f = fun x -> x ^ ""
let g _ = fun x -> x ^ ""
```  
In this case, an inlay hint appears for `x` in `f`, but no hint is shown for `g`.  

In VS Code, this results in the following display:  
image  

This is because

```ocaml
let f _ = ...
```  
is syntactic sugar for:  

```ocaml
let f = fun _ -> ...
```  
where `fun _ ->` introduces a ghost location (the whole fun node). 

This is why you needed to disable the check on ghost location in your `samples.t` test. 

### Proposed Fix  

This PR replaces ghost location handling with the [[@merlin.hide]](https://github.com/ocaml/merlin) attribute. This allows for more precise control over which parts of the AST should be ignored for inlay hints.

To make this approach work correctly, I refactored the main AST iterator to use `Ast_iterators.iter_only_visible`.  The iterator now ensures attributes are not skipped by recursing on non-immediate sub-terms.  

Also, this restructuring simplifies the code and makes inlay hints more consistent across different contexts (e.g., top-level `let` bindings, class-level `let` bindings, and expression `let` bindings).  

### Additional Changes  

- Updated `samples.t` with new test cases illustrating the improved inlay hints behavior.  
- Removed support for the `avoid-ghost-location` option (both in the CLI and protocol).  

### Related  

- https://github.com/ocaml/merlin/pull/1892 (an early version of this PR)
- https://github.com/ocaml/ocaml-lsp/pull/1290 : the introduction of the ghost location check
- https://github.com/ocaml/ocaml-lsp/pull/1159

### Screenshot 

image


mlasson:
Is it ok to change the protocol/library in a non-backward compatible way ?

voodoos:
Hi @mlasson, sorry for not having answered earlier, I must have missed it.

Looking at the tests this seems really interesting, thanks for that well though of and tested PR.

Are you happy with it's current state ? Do you think you could rebase it so that we can do a proper review ?

panglesd:
I think this has mostly been superseded by #1923: I used this PR's test file, added `-function-params true` where applicable, and the diff produced by `dune test` is rather small:

```diff
File "tests/test-dirs/inlay-hint/samples.t", line 1, characters 0-0:
diff --git a/_build/default/tests/test-dirs/inlay-hint/samples.t b/_build/default/tests/test-dirs/inlay-hint/samples.t.corrected
index 1ccb4021..f23d7164 100644
--- a/_build/default/tests/test-dirs/inlay-hint/samples.t
+++ b/_build/default/tests/test-dirs/inlay-hint/samples.t.corrected
@@ -239,6 +239,13 @@ Class-level let bindings without let hinting
   {
     "class": "return",
     "value": [
+      {
+        "pos": {
+          "line": 1,
+          "col": 17
+        },
+        "label": "int"
+      },
       {
         "pos": {
           "line": 1,
@@ -488,13 +495,27 @@ Support for @merlin.hide
   {
     "class": "return",
     "value": [
+      {
+        "pos": {
+          "line": 2,
+          "col": 16
+        },
+        "label": "int"
+      },
       {
         "pos": {
           "line": 2,
           "col": 7
         },
         "label": "int"
+      },
+      {
+        "pos": {
+          "line": 1,
+          "col": 21
+        },
+        "label": "'a"
       }
     ],
     "notifications": []
-  }
\ No newline at end of file
+  }
```

so only a few added inlay hints.

So I think the initial problem was fixed by #1923, in a way that seems more satisfying than removing the `loc_ghost` check, as these are generated by PPXs and we don't want to introduce inlay hint for generated nodes! This PR can probably be closed, but thanks for the work!

voodoos:
Thanks for having a look @panglesd ! Can you open a follow-up PR with a test ?

And thanks again @mlasson, don't hesitate to open an issue if you still feel the behavior is unsatisfying. We are trying to be more pro-active with issue triage and PR reviews these days 🙂

PR #1909, opened 2025-03-12
Upstream lyrm in merlin-domains branch

lyrm:
Replaced #1908. 

Note: These changes break one test due to the removal of laziness. Previously, `Mpipeline.for_completion` used laziness to avoid recomputing the entire pipeline a second time. Since this function is no longer called, the test for label completion, which seems to be the main use for the `for_completion` function, is now broken. This has been marked as a TODO in the code.


xvw:
@voodoos can you just approve Workflows? 
Thanks 👍🏾

PR #1917, opened 2025-03-31
Improve location of some custom operators

### LINKED ISSUE #1915: Go-to-definition does not work on custom operators
Go-to-definition doesn't work on (all) [custom operators](https://ocaml.org/docs/operators), which would be very useful for navigating DSLs.

For example,

```ocaml
let ( := ) v a = Printf.printf "%s = %d" v a
let () = "foo" := 3
(*              ^ *)
```

If I place my cursor at the point indicated by the `^`, go-to-definition does nothing.

Similarly, for custom indexing operators:

```ocaml

let (.%{;..}) a k = Printf.printf "%s.coeffRef(%d);\n" a k.(0)
let (.%{ }) a k = Printf.printf "%s.coeffRef(%d);\n" a k
let name = "baz"
let () = name.%{2;4}
let () = name.%{5}
```

Placing the cursor at the `%` on either like 4 or 5 does not succeed.

_Originally opened as https://github.com/ocamllabs/vscode-ocaml-platform/issues/1766_

WardBrian:
This (maybe?) closes #1915.

It adds the same test as #1916, plus a:

- Definite fix for `:=`
  - Just lex this as an operator.
- Definite fix for reified custom indexing (e.g, something like `( .@(;..)<- )`)
  - We lex this as `DOTOP SEMI DOTDOT ...` and detect it in the driver code
- Heuristic fix for other custom indexing
  - When we see a `DOTOP` in the wild, we assume it is single-indexing without assignment and complete it accordingly. E.g., the code `name.@( (* whatever here ... *)`  will get turned into `.@ ( )` by reconstruct_id, even if we'd like to also detect things like `;` or `<-` to disambiguate. 
  - This is obviously wrong-in-general, but I think it's more _useful_ than the current behavior, since users are encouraged to always define this variant if using the custom indexing, and they're often near each other. I also don't know if it can be done any better, since truly disambiguating requires I think arbitrary amounts of parsing?

The test shows both the cases that don't work before (see existing output in #1917) and the cases that "work" but produce incorrect locations due to the third bullet point above.


WardBrian:
Fuzzy CI diff looks all intentional to me

voodoos:
Thanks for the patch @WardBrian. I will review it when I get some spare cycles (it might take some time..).

WardBrian:
@voodoos gentle ping -- is there still interest in this being looked at?

WardBrian:
@voodoos is there anything I could do on my end to make this more reviewable? I'd be happy to split it into smaller PRs for example. The inability to go-to-def on custom operators continues to be a thorn in my side

voodoos:
Hi @WardBrian, sorry for the long waiting time ! I have been thinking about this on-and-off for a while, and convinced myself that the approach I first suggested and that you implemented was not robust enough.

Things have changed (in 5.4 I think), and there are now additional location information in the Typedtree's longidents that allow us to skip the sketchy reconstruction in many cases. I took your new tests and opened a new PR implementing that new approach here: https://github.com/ocaml/merlin/pull/2126

It would be great if you could check that the results are the one you expect :-)

I limited the new backend to locate for now, but I expect it should greatly improve the results for type enclosing too.

PR #1923, opened 2025-04-09
Inlay hint function params

xvw:
Adds triggerable support for inlay-hints for function parameters. The positive point is that we were disturbed by the ‘slightly non-deterministic’ support for inlay-hints, and this is in fact because since the change to the representation of function arguments, these were never taken into account. This PR also fixes this behaviour.

(cc @awilliambauer)

### DIFF
--- a/tests/test-dirs/issue1900.t/run.t
+++ b/tests/test-dirs/issue1900.t/run.t
@@ -1,36 +1,14 @@
   $ dune exec ./main.exe
   test
 
-FIXME: There should be no error.
+There should be no error.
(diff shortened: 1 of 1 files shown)


xvw:
Hmm CI failures seems weird.

xvw:
(the failure related to `ocaml-lsp-compat` is legit, see: https://github.com/ocaml/ocaml-lsp/pull/1515)

voodoos (approved):
Thanks, looks good to me. It a clear improvement and long-awaited fix.

Issue #1588, opened 2023-04-01
Jump to definition broken on file with typing error


ddickstein:
I don't know if this is related to https://github.com/ocaml/merlin/issues/1580, so I'm opening this as a separate issue. Jumping to `test` on line 3 gives "Not in environment 'test'". As in that issue, fixing the type error (in this case, removing the annotation on `~f` in the first line) fixes locate.

```ocaml
let test ~f:(_ : unit -> unit) = ()
type t = F : { f : unit -> 'fn } -> t
let call (F { f }) = test ~f
```

Tested on https://github.com/ocaml/merlin/commit/945e5e10373640b622c47a54b84f4c2a0252c477

voodoos:
Thanks for the report, probably a similar issue than the others: Merlin recovery lacks finesse in some cases.

Issue #1580, opened 2023-03-16
Jump to definition broken on file with simple typing error


ddickstein:
```ocaml
module type S = sig
  val foo : unit -> ('a -> 'a -> bool) -> unit
end

module F (M : S) = struct
  let z () = M.foo () compare
end
```

Trying to jump to `foo` gives "Not in environment 'M.foo'". Fixing the type of the function `foo` takes to return `int` instead of `bool` makes jumping to definition work properly.

Tested on https://github.com/ocaml/merlin/commit/945e5e10373640b622c47a54b84f4c2a0252c477

PR #1930, opened 2025-05-02
Add reproduction for some locate issues

voodoos:
- Issues #1580 #1588 are due to bad typing recovery.
- #1585 is due to Dune not generating configuration force the source of a copied file
- #1610 is due to reconstruct identifier not working properly with functor patches

### DIFF
--- a/tests/test-dirs/config/copy-issue.t
+++ b/tests/test-dirs/config/copy-issue.t
@@ -1,0 +1,36 @@
+  $ cat >dune-project <<'EOF'
+  > (lang dune 3.0)
+  > EOF
+
+  $ mkdir dep
+  $ cat >dep/dep.ml <<'EOF'
+  > let txt = "Hello!"
+  > EOF
+
+  $ mkdir exe
+  $ cat >exe/main.ml << 'EOF'
+  > print_endline Dep.txt
+  > EOF
+
+  $ cat >exe/dune << 'EOF'
+  > (executable
+  >  (name main))
+  > (copy_files# %{project_root}/dep/*.ml)
+  > EOF
+
+  $ dune exec ./exe/main.exe
+  Hello!
+
+  $ $MERLIN single errors -filename exe/main.ml <exe/main.ml | jq '.value'
+  []
+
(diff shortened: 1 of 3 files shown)


voodoos:
@ddickstein I added reproduction for the issues you listed when they were missing.
I also fixed the one linked to the typer recovery.

src/ocaml/typing/typecore.ml
```diff
@@ -5408,7 +5408,7 @@ and type_label_exp create env loc ty_expected
   if is_poly then check_univars env "field value" arg label.lbl_arg vars;
   (lid, label, {arg with exp_type = instance arg.exp_type})
 
-and type_argument ?explanation ?recarg env sarg ty_expected' ty_expected =
+and type_argument_ ?explanation ?recarg env sarg ty_expected' ty_expected =
```
xvw, line 5411: what is the difference between `type_argument` and `type_argument_` ? could we find more meaningful names? 
And it looks strange that we touch to something inside `src/ocaml/typing`? Is this related to type recovery?

xvw: Oh ok! In fact `type_argument_` is the OCaml one and `type_argument` is a wrapper dealing with Merlin specific stuff? Make sense.

src/ocaml/typing/typecore.ml
```diff
@@ -5524,6 +5524,36 @@ and type_argument ?explanation ?recarg env sarg ty_expected' ty_expected =
...
+        let saved = save_levels () in
+        try
+          type_argument_ ?explanation ?recarg env sarg ty_expected' ty_expected
+        with exn ->
+          Msupport.erroneous_type_register ty_expected;
+          raise_error exn;
+          set_levels saved;
+          let loc = sarg.pexp_loc in
+          {
+            exp_desc = Texp_ident
```
xvw, line 5540: WDYT about extracting the `exp_desc` in a dedicated variable?

voodoos: Let's keep that for the type-recovery cleanup and upstreaming 😉

PR #1936, opened 2025-05-16
Handle object expression inside a let in outline

Tim-ats-d:
ocaml-lsp outline reports imbricated object inside let expression but merlin not. This fix aims to use merlin to generate outlines in ocaml-lsp without introduce regression in ocaml-lsp

The fix is ugly but I can refactor if needed.

cc @voodoos @xvw

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ unreleased
     - `inlay-hints` fix inlay hints on function parameters (#1923)
     - Fix issues with ident validation and Lid comparison for occurrences (#1924)
     - Handle class type in outline (#1932)
+    - Handle locally defined value in outline (#1936)
   + ocaml-index
     - Improve the granularity of index reading by segmenting the marshalization
       of the involved data-structures. (#1889)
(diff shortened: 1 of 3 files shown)


xvw (approved):
For me, the fix is not ugly! Good job!

voodoos (changes requested):
Thanks a lot @Tim-ats-d 

I suspected that you could have a simpler implementation by relying more on the flat structure of the browse tree but I had to play with the code to find it, so i decided to directly push the change to your code. Please have a look, it does remove quite a lot of the ad-hoc treatments:

https://github.com/ocaml/merlin/pull/1936/commits/0db503f690705f8f2b3aeaee221ddcaee7406db1

Apart from that, could you make the changelog entry more complete ? There is a "major' change in behavior here since before that patch Merlin outline's did not account for locally defined value, unlike ocaml-lsp. It seems reasonable to have them, but this should be made clear to users.

Tim-ats-d:
>  Please have a look, it does remove quite a lot of the ad-hoc treatments
Nice, I understand Mbrowse better now. Thank you for the help!

I have also edited the change entry.

voodoos (approved):
Thanks Tim, I pushed a commit with slightly more code reuse, and there might be some more remaining :-) 

But this is probably good to go like that.

xvw (approved):

PR #1948, opened 2025-07-02
Add a new command to extract expression into a fresh let binding

Tim-ats-d:
This PR adds a new command `refactoring-extract-region` to merlin protocol which allows extracting an expression into a let binding. See the [interface comments](https://github.com/Tim-ats-d/merlin/blob/0ddf2a3564cc64515564dd282f02b6519bff6f0d/src/analysis/refactor_extract_region.mli) for supported cases

## Implementation

The implementation of the substitution is a bit hacky and relies on string manipulation, especially when the generated binding is a `and`. I couldn't deal with typedtree locations for the generated code, so I had to make a few manual location calculations that I tried to make as explicit as possible. Please tell me if there's a better way to achieve it.

It still lacks a bit of testing, I'll be adding more in the next few days.

cc @xvw @voodoos

### DIFF
--- a/src/analysis/refactor_extract_region.ml
+++ b/src/analysis/refactor_extract_region.ml
@@ -36,11 +36,11 @@ let clean_up_for_printing expr =
                   { pexp_desc =
                       Pexp_tuple
                         (_
-                        :: ({ pexp_desc =
+                         :: (_, ({ pexp_desc =
                                 Pexp_constant
                                   { pconst_desc = Pconst_string _; _ };
                               _
-                            } as const)
+                                } as const))
                         :: _);
                     _
                   } )
@@ -206,13 +206,13 @@ let rec find_pattern_var : type a. a Typedtree.general_pattern -> Path.t list =
  fun { Typedtree.pat_desc; _ } ->
   match pat_desc with
   | Typedtree.Tpat_var (ident, _, _) -> [ Pident ident ]
-  | Tpat_tuple pats -> List.concat_map ~f:find_pattern_var pats
-  | Tpat_alias (pat, ident, _, _) -> Pident ident :: find_pattern_var pat
+  | Tpat_tuple fields -> List.concat_map ~f:(fun (_, pat ) -> find_pattern_var pat) fields
+  | Tpat_alias (pat, ident, _, _, _) -> Pident ident :: find_pattern_var pat
   | Tpat_construct (_, _, pats, _) -> List.concat_map ~f:find_pattern_var pats
   | Tpat_variant (_, Some pat, _) -> find_pattern_var pat
   | Tpat_record (fields, _) ->
     List.concat_map ~f:(fun (_, _, field) -> find_pattern_var field) fields
-  | Tpat_array arr -> List.concat_map ~f:find_pattern_var arr
+  | Tpat_array (_, arr) -> List.concat_map ~f:find_pattern_var arr
   | Tpat_lazy pat | Tpat_exception pat -> find_pattern_var pat
   | Tpat_value pat ->
     find_pattern_var (pat :> Typedtree.value Typedtree.general_pattern)


CHANGES.md
```diff
@@ -10,6 +10,7 @@ unreleased
     - `inlay-hints` fix inlay hints on function parameters (#1923)
     - Fix issues with ident validation and Lid comparison for occurrences (#1924)
     - Handle class type in outline (#1932)
+    - Handle object expression inside a let in outline (#1936)
```
voodoos: That looks unrealted to this PR

xvw: Yes, I think the branch should be rebased on `main` ;)

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,382 @@
...
+let () =
+  Location.register_error_of_exn (function
+    | Nothing_to_do -> Some (Location.error "Nothing to do")
+    | Not_allowed_in_interface_file ->
+      Some
+        (Location.error
+           "Expression extraction is only allowed in implementation file")
+    | _ -> None)
+
+module FreshName = struct
```
xvw: It is nitpicking but I definitively prefer `Fresh_name` ;)

xvw:
There is one test that seems fail. Is that an issue or you need to repromote it?

doc/dev/PROTOCOL.md
```diff
@@ -485,6 +485,25 @@ The result is returned as a list of:
...
 
+### `refactor-extract-region -start <position> -stop <position> -extract-name <name>`
+
+```
+    -start <position>  Where extracted region start
+    -stop <position>  Where extracted region end
+    -extract-name <name>  Name used for the generated let binding
+```
+
+Returns the string `Nothing to do` (if extractor is not ables to select an expression to extract in the given position interval) or the following object:
```
xvw, line 499: Returning a string seems to be an issue, we should take look at other commands in order to throw a proper exception>

src/analysis/parsetree_utils.mli
```diff
@@ -6,3 +6,5 @@ open Parsetree
 type nonrec constant_desc = constant_desc
 
 val constant_desc : constant -> constant_desc
+
+val filter_expr_attr : expression -> expression
```
xvw: Can you add a small docstring that explain the behaviour of the function (it is less obvious than `constant_desc` :)

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,395 @@
...
+exception Not_allowed_in_interface_file
+
+let () =
+  Location.register_error_of_exn (function
+    | Nothing_to_do -> Some (Location.error "Nothing to do")
+    | Not_allowed_in_interface_file ->
+      Some
+        (Location.error
+           "Expression extraction is only allowed in implementation file")
+    | _ -> None)
```
xvw, line 13: Iam not sure about `register_error`. Does the result is wrapped into an error like other commands?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,395 @@
...
+
+module Gen = struct
+  let unit = Longident.Lident "()" |> Location.mknoloc
+
+  (* Generates [let name = body]. *)
+  let toplevel_let ~name ~body =
+    let open Ast_helper in
+    let pattern = Pat.mk (Ppat_var { txt = name; loc = Location.none }) in
+    let body = Parsetree_utils.filter_expr_attr body in
+    Str.value Nonrecursive [ Vb.mk pattern body ]
```
xvw: Looking to `Nonrecursive`.
I am a little but curious about case like that: 

```ocaml
let f = 
  let x = 1 in 
  let y = 2 in 
  let z = x + y in 
  z + 1
```

Let's imagine that we extract `let z = x + y in` or `let z = x + y` we can imagine that it will produce the following code: 

```ocaml
let z x y = x + y

let f = 
  let x = 1 in 
  let y = 2 in 
  (z x y) + 1
```

No? In other terms, I think that if the selected range is a let-binding, we change the binding as a toplevel value (and we preserve the name if it makes senses, ie: if the binding is free, using Fresh_name, or using a fresh binding) 

So we can imagine that if we extract a recursive definition, we preserve it.

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,395 @@
...
+
+module Fresh_name = struct
+  (* Generate a fresh name that does not already exist in given environment. *)
+  let gen_val_name basename env =
+    let rec loop n =
+      let guess = basename ^ Int.to_string n in
+      if Env.bound_value guess env then succ n |> loop else guess
+    in
+    loop 1
+end
```
xvw: We have a lot of discussion related to identifying free variables, but can we not just using `Env.bound_value`? For example: 

```ocaml
let a = 10 
let b = 
  let c = 2 in 
  let d = 3 in 
  let z = a + c d 
  in z + 1
``

Why not using the `env` related to `b` ?

File: src/analysis/refactor_extract_region.ml [hunk 8]
```diff
@@ -0,0 +1,395 @@
...
+      let line, col = Lexing.split_pos loc.Location.loc_start in
+      Msource.get_offset src (`Logical (line, col))
+    in
+    let (`Offset end_offset) =
+      `Logical (Lexing.split_pos loc.loc_end) |> Msource.get_offset src
+    in
+    String.sub (Msource.text src) ~pos:start_offset
+      ~len:(end_offset - start_offset)
+    |> Msource.make
+end
```
xvw: Not sure about the pattern `module A = struct include A, and reuse A as fully qualified expr`. 
You can have a function, outside of your `Msource`?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,395 @@
...
+         with a unit parameter. *)
+      (Gen.let_unit_toplevel, Gen.fun_apply_unit)
+    | _ ->
+      (Gen.toplevel_function bounded_vars, Gen.fun_apply_params bounded_vars)
+  in
+  let name =
+    Option.fold extract_name
+      ~none:(Default { basename = "fun_name" })
+      ~some:(fun name -> Fixed name)
+  in
```
xvw, line 440: Not sure that `fold` is really more readable than `match` ;)

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,395 @@
...
+  let is_function = function
+    | { Typedtree.exp_desc = Texp_function _; _ } -> true
+    | _ -> false
+  in
+  let { bounded_vars; gen_binding_kind } =
+    analyze_expr expr expr_env ~toplevel_item ~local_defs ~mconfig
+  in
+  let generated_binding, generated_call =
+    match bounded_vars with
+    | [] when Fun.negate is_function expr ->
```
xvw: `not (is_function expr)`

doc/dev/PROTOCOL.md
Tim-ats-d: Yes, I don't really return a raw string but I'm referring to the error message value which look like this:
```json
{
   "class": "error",
   "value": "Nothing to do",
   "notifications": []
}
```

Tim-ats-d: and since other commands uses this "abuse of language" in their documentation (like [here](
https://github.com/ocaml/merlin/blob/500487f53ac4b08df4fdecf45465941d358866fb/doc/dev/PROTOCOL.md#syntax-document--position-position)) I think it was fine

src/analysis/parsetree_utils.mli
Tim-ats-d: Does the following comment do the job?
> Filter parsetree attributes which are prefixed by `merlin.` in given expression.

src/analysis/refactor_extract_region.ml
Tim-ats-d: Yes, the error message looks like this:
```json
{
   "class": "error",
   "value": "Nothing to do",
   "notifications": []
}
```

Tim-ats-d: I've asked @voodoos about the exception raising policy in Merlin and he told me it's the right way to go.

src/analysis/refactor_extract_region.ml
Tim-ats-d: I wanted to know if you think it's worth including the function in the `Msource` module?

src/analysis/refactor_extract_region.ml
xvw: Perfect!

doc/dev/PROTOCOL.md
xvw: Nice! Thanks!

src/analysis/refactor_extract_region.ml
Tim-ats-d: Fixed.

src/analysis/parsetree_utils.mli
xvw: I think that we can name it: `expr_remove_merlin_attributes` so?

src/analysis/refactor_extract_region.ml
Tim-ats-d: Fixed (`Fun.negate` is so cool 😩).

src/analysis/refactor_extract_region.ml
xvw: Since it seems that is used only here, I do not think it is necessary. We can delay the question to another feature and just extracting the function

src/analysis/refactor_extract_region.ml
xvw: For my point of view, if make more sense where it is used in partial application context. Like: `List.filter (Fun.negate predicate) list` (instead of `List.filter (fun x -> not (predicate x)) list`.

src/analysis/parsetree_utils.mli
Tim-ats-d: Absolutely.

src/analysis/refactor_extract_region.ml
Tim-ats-d: Fixed.

Tim-ats-d:
> There is one test that seems fail. Is that an issue or you need to repromote it?

I replaced it by other more relevant tests from ocaml-lsp (https://github.com/ocaml/merlin/pull/1948/commits/38055ff0515a76ea2e7cceae75cfcd4aeaaafb11#diff-1b5f000f9708e3dfe32d04e4ca43f63eb3d55ce173e7c8cd63bbca41e61639d1)

doc/dev/PROTOCOL.md
```diff
@@ -485,6 +485,25 @@ The result is returned as a list of:
 }
 ```
 
+### `refactor-extract-region -start <position> -stop <position> -extract-name <name>`
+
+```
+    -start <position>  Where extracted region start
+    -stop <position>  Where extracted region end
```
xvw: Is, in Merlin, usual to use `start` and `stop` ? Can we returning the "délétion range" into a specific object ? Like in sélection range?

src/commands/new_commands.ml
```diff
@@ -565,6 +565,37 @@ let all_commands =
...
+        ]
+      ~default:(`None, `None, None)
+      begin
+        fun buffer (start, stop, name) ->
+          match (start, stop, name) with
+          | `None, `None, _ -> failwith "-start <pos> and -end are mandatory"
+          | `None, _, _ -> failwith "-start <pos> is mandatory"
+          | _, `None, _ -> failwith "-end <pos> is mandatory"
+          | (#Msource.position as start), (#Msource.position as stop), name ->
+            let raw_source = Mpipeline.raw_source buffer in
```
voodoos: Is there a reason why this is done here and not in `Query_commands` ?

src/analysis/refactor_extract_region.mli
```diff
@@ -0,0 +1,113 @@
+(** Region extractor allows extracting arbitrary expression into a fresh
```
voodoos:
```suggestion
(** Region extractor allows extracting arbitrary expressions into a fresh
```

src/analysis/refactor_extract_region.mli
```diff
@@ -0,0 +1,113 @@
...
+    toplevel binding. The extractor detects bounded variables inside the
```
voodoos:
```suggestion
    toplevel binding. The extractor detects bound variables inside the
```

src/analysis/refactor_extract_region.mli
```diff
@@ -0,0 +1,113 @@
...
+    extracted expression and performs code generation acordingly.
```
voodoos:
```suggestion
    extracted expression and performs code generation accordingly.
```

src/analysis/refactor_extract_region.mli
```diff
@@ -0,0 +1,113 @@
...
+      and y = 80
+
+      and fun_name2 x = 10 + y + x
+    ]}
+
+    We can see that extractor detects this kind of pattern and extracts the
+    expression inside an [and] binding. It also substitutes the expression by a
+    call to the fresh generated function with the correct parameters.
+
+    Finally, if there is no bounded variable in the expression, a trailing unit
```
voodoos:
```suggestion
    Finally, if there is no bound variable in the expression, a trailing unit
```

src/analysis/refactor_extract_region.mli
```diff
@@ -0,0 +1,113 @@
...
+      let my_list =
+        print_endline "Wild side effect!";
+        1 :: [ 2; 3; 4 ]
+    ]}
+
+    {[
+      let fun_name1 () =
+        print_endline "Wild side effect!";
+        [ 1; 2; 3; 4 ]
+      let f = fun_name1 ()
```
voodoos: Shouldn't it be `my_list =` ?

src/analysis/refactor_extract_region.mli
```diff
@@ -0,0 +1,113 @@
...
+      and fun_name2 x = 10 + y + x
+    ]}
+
+    We can see that extractor detects this kind of pattern and extracts the
+    expression inside an [and] binding. It also substitutes the expression by a
+    call to the fresh generated function with the correct parameters.
+
+    Finally, if there is no bounded variable in the expression, a trailing unit
+    parameter is added to the generated let binding in order to preserve the
+    evaluation order. Let's extract the entire body of [x]:
```
voodoos: There's no `x` ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+  | None -> false
+  | Some _ -> true
+
+let substitute ~start ~stop ?extract_name buffer typedtree =
+  match typedtree with
+  | `Interface _ -> raise Not_allowed_in_interface_file
+  | `Implementation structure -> begin
+    let enclosing =
+      Mbrowse.enclosing start [ Mbrowse.of_structure structure ]
+    in
```
voodoos: Nitpick: usually this kind of filtering and browse tree manipulation is done in `Query_commands`.

src/analysis/refactor_extract_region.mli
```diff
@@ -0,0 +1,113 @@
...
+(** Raised when extractor is not ables to select an expression to extract in
+    given location interval. *)
+exception Nothing_to_do
+
+(** Raised when extraction is called inside an interface file. *)
+exception Not_allowed_in_interface_file
+
+(** Is an expression is extractable in the given region? *)
+val is_region_extractable :
+  start:Lexing.position -> stop:Lexing.position -> Mbrowse.t -> bool
```
voodoos, line 92: Looks like dead code... Or is it meant for users of Merlin as a lib ?

edit: for code action detection ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+            cs
+        | _ -> None
+      end
+      | _ -> None)
+    enclosing
+
+let extract_region ~start ~stop enclosing =
+  let open Option.Infix in
+  most_inclusive_expr ~start ~stop enclosing >>= fun (expr, expr_env) ->
+  (* si contenu de l'expr contient une expression local alors inextrayable *)
```
voodoos: I suggest we stay true to Gabriel's legacy and stick with English comments here.

(Pull request https://github.com/ocaml/merlin/pull/1 😮)

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+    | Expression expr
+      when node_loc.loc_ghost = false && is_inside_region node_loc ->
+      (* We filter expression that have a ghost location. Otherwise, expression
+        such as [let f x = 10 + x] can be extracted and this can lead to invalid 
+        code gen.      ^^^^^^^^^^ *)
+      Some (expr, env)
+    | _ -> select_deeper node env
+  in
+  nodes |> List.rev
+  |> Stdlib.List.find_map (fun (env, node) -> select_among_child env node)
```
voodoos: I am not saying that's wrong but it's a bit complex. It might be worth to describe the algorithm in a comment. 

What does `most_inclusive_expr` means ? the largest expression around start and stop ?

The parameter `nodes` is a list of enclosings around the start position from the deepest to the topelevel.
It's reversed searched for an expression that fits the range. 

Your function `select_deeper` appears to mean that you need to look for expressions that are siblings of the nodes forming the enclosing your considering, this is worth explaining.

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+
+let find_associated_toplevel_item expr enclosing =
+  Stdlib.List.find_map
```
voodoos, line 481: Why not `List.find_map` ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+
+type extraction =
+  { expr : Typedtree.expression;  (** Expression that being extracted *)
+    expr_env : Env.t;  (** Environment of the extracted expression *)
+    toplevel_item : toplevel_item;
+        (** The value binding toplevel or class declaration enclosing the extracted expression. *)
+    name : extraction_name;  (** Binding name of the extracted expression. *)
+    gen_binding_kind : rec_flag;
+    generated_binding : generated_binding;
+    generated_call : generated_call;
```
voodoos: Technically, this is a binding_generator, not yet generated.

Other possible names:

`generate_binding : binding_generator`
`make_binding : ...`

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+      (Gen.let_unit_toplevel, Gen.fun_apply_unit)
+    | _ ->
+      (Gen.toplevel_function bounded_vars, Gen.fun_apply_params bounded_vars)
+  in
+  let name =
+    match extract_name with
+    | None -> Default { basename = "fun_name" }
+    | Some name -> Fixed name
+  in
+  let remove_path_prefix expr =
```
voodoos: I think we should have a more precise name to describe what this non-trivial mapping is doing. Something like `remove_path_prefix_of_bound_values` ? Along with a comment to explain why we need it.

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+    let mapper =
+      { Tast_mapper.default with
+        expr =
+          (fun mapper expr ->
+            match expr.Typedtree.exp_desc with
+            | Texp_ident (Pdot (path, val_name), longident, vd)
+              when is_bound_var path && is_module_bound path ->
```
voodoos: How can the same thing `path` be a `var` and a `module` at the same time ? Names are misleading ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+      path :: acc
+    | Pattern pat -> find_pattern_var pat @ acc
+    | _ ->
+      Lazy.force node.t_children
+      |> List.concat_map ~f:occuring_vars_path
+      |> List.append acc
+  in
+  loop [] node |> Path.Set.of_list |> Path.Set.elements |> List.rev
+  |> List.filter ~f:(fun path ->
+         (* TODO: fix this *)
```
voodoos: Is it fixed ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+    match node.Browse_tree.t_node with
+    | Browse_raw.Expression { exp_desc = Texp_ident (path, _, _); _ } ->
+      path :: acc
+    | Pattern pat -> find_pattern_var pat @ acc
+    | _ ->
+      Lazy.force node.t_children
+      |> List.concat_map ~f:occuring_vars_path
+      |> List.append acc
+  in
+  loop [] node |> Path.Set.of_list |> Path.Set.elements |> List.rev
```
voodoos: Is the order relevant here ? If not we can remove the `List.rev`. Also, you could accumulate directly on the `Path.Set.t`, return it and have `analyze_expr` use `Path.Set.fold` if the order is not relevant, skipping the list altogether.

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+
+  let fun_apply_unit = fun_apply [ Ast_helper.Exp.ident unit ]
+
+  let fun_apply_params params =
+    params
+    |> List.map ~f:(fun param -> ident ~name:(Path.name param))
+    |> fun_apply
+end
+
+let source_sub_loc src loc =
```
voodoos: For me, that names implies that the resutl is a `loc`. But the result is a fragment of the source corresponding to the given loc. Maybe `extract_source_around_loc` ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+  let val_name = choose_name name expr_env in
+  let fresh_call =
+    let parenthised_opt s =
+      if call_need_parenthesis then "(" ^ s ^ ")" else s
+    in
+    generated_call ~name:val_name
+    |> Format.asprintf "%a" Pprintast.expression
+    |> parenthised_opt
+  in
+  let toplevel_item_span = source_sub_loc buffer toplevel_item.loc in
```
voodoos: This is actually a fragment of the source, not simply locs. Better name ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+    let end_lnum =
+      start_lnum + expr.exp_loc.loc_end.pos_lnum
+      - expr.exp_loc.loc_start.pos_lnum
+    in
+    { expr.exp_loc with
+      loc_start = { expr.exp_loc.loc_start with pos_lnum = start_lnum };
+      loc_end = { expr.exp_loc.loc_end with pos_lnum = end_lnum }
+    }
+  in
+  let substitued_toplevel_binding =
```
voodoos:
```suggestion
  let substituted_toplevel_binding =
```

voodoos: Also, are these always "toplevel" ? Toplevel means an item at the root of the file. If this is the closest (structure-)item enclosing the refactoring, then it's just an (structure-)item, not a toplevel one.

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+  loop [] node |> Path.Set.of_list |> Path.Set.elements |> List.rev
+  |> List.filter ~f:(fun path ->
+         (* TODO: fix this *)
+         Ident.name (Path.head path) <> "Stdlib")
+
+let analyze_expr expr expr_env ~toplevel_item =
+  let is_value_unbound path =
+    let is_bound path env =
+      try
+        let _ = Env.find_value path env in
```
voodoos, line 243: `Env.bound_vale` would be a better choice here.

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+
+  (* Generates [let name = body]. *)
+  let toplevel_let ~name ~body =
+    let open Ast_helper in
+    let pattern = Pat.mk (Ppat_var { txt = name; loc = Location.none }) in
+    let body = clean_up_for_printing body in
+    Str.value Nonrecursive [ Vb.mk pattern body ]
+
+  (* Generates [let name () = body]. *)
+  let let_unit_toplevel ~name ~body =
```
voodoos, line 73: Couldn't this one be a special case of the next function `toplevel_function` ?

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,479 @@
...
+            (* We also have to remove poly extra that cause unexpected "!poly!"
+               to be printed in generated code. This happens when you try
+               to extract the body of a method. *)
+            expr
+          | _ -> Ast_mapper.default_mapper.expr mapper expr)
+    }
+  in
+  mapper.expr mapper expr |> Parsetree_utils.expr_remove_merlin_attributes
+
+module Gen = struct
```
voodoos, line 60: I expect these utilities to land in a separate module common to multiple refactoring in the future. We should start documenting precisely their behavior. (the last few ones don't have a comment).

voodoos (commented):
Thanks @Tim-ats-d !

It looks good overall, I mostly have some naming issues that made the re-reading of the code harder at some places.

I did not review thoroughly the correctness of the code generation.

src/commands/new_commands.ml
Tim-ats-d: no :smile:

Tim-ats-d: Fixed.

src/analysis/refactor_extract_region.mli
Tim-ats-d: Fixed.

src/analysis/refactor_extract_region.mli
Tim-ats-d: Fixed.

src/analysis/refactor_extract_region.mli
Tim-ats-d: Fixed.

src/analysis/refactor_extract_region.mli
Tim-ats-d: Fixed.

src/analysis/refactor_extract_region.mli
Tim-ats-d: Fixed.

src/analysis/refactor_extract_region.mli
Tim-ats-d: Yes absolutely. See https://github.com/ocaml/ocaml-lsp/pull/1546/files#diff-036a30d4248c7a5b33df77be801210b301f4a846c270717a3561896ccb3c8968R40

src/analysis/refactor_extract_region.ml
Tim-ats-d: Because `List.find_map` is overridden by `Std.List.find_map` which sadly raises an exception instead of returning an `option`.

src/analysis/refactor_extract_region.ml
Tim-ats-d: The order is relevant so that the order of the parameters corresponds to the order in which the variables appear in the extracted expression.

src/analysis/refactor_extract_region.ml
Tim-ats-d: No, I still don't know how to be sure and certain that I'm filtering only the original Stdlib module and not just any module named “Stdlib.” We discussed this with @xvw and perhaps module stamp can be relevant.

src/analysis/refactor_extract_region.ml
Tim-ats-d: I use `Env.find_value` instead of `Env.bound_value` here because the latter takes a string instead of a path.

src/analysis/refactor_extract_region.ml
xvw: I am not sure if it is the case in the current implementation. 
@Tim-ats-d I remember that extraction was failing inside modules, is that fixed?

src/analysis/refactor_extract_region.ml
Tim-ats-d: I thought it was simpler to have two separate functions because `toplevel_function` takes a list of `Path.t` 

Maybe I could refactor these two so that they use another function or consider that if the list of paths given as an argument is empty, this generates a function that simply takes unit? What do you think?

src/analysis/refactor_extract_region.ml
Tim-ats-d: Yeah it's the plan :slightly_smiling_face: (i added a comment for the two functions that didn't have one.)

src/analysis/refactor_extract_region.ml
Tim-ats-d: I have moved this in the `Query_commands`.

src/analysis/refactor_extract_region.ml
voodoos: Ah, fine then.

src/analysis/refactor_extract_region.ml
Tim-ats-d: > What does most_inclusive_expr means ? the largest expression around start and stop ?
Yes.

I have removed the definition of `select_deeper`. This is confusing because I only use it once and it's hard to find a good name for it. Also add some comments.
```ocaml
(* [most_inclusive_expr ~start ~stop nodes] tries to find the most inclusive expression
   within the range [start]-[stop] among [nodes].

   [nodes] is a list of enclosings around the start position from the deepest
   to the topelevel. It's reversed searched for an expression that fits the range. *)
let most_inclusive_expr ~start ~stop nodes =
  let is_inside_region =
    Location_aux.included
      ~into:{ Location.loc_start = start; loc_end = stop; loc_ghost = true }
  in
  let rec select_among_child env node =
    let node_loc = Mbrowse.node_loc node in
    match node with
    | Expression expr
      when node_loc.loc_ghost = false && is_inside_region node_loc ->
      (* We filter expression that have a ghost location. Otherwise, expression
        such as [let f x = 10 + x] can be extracted and this can lead to invalid 
        code gen.      ^^^^^^^^^^ *)
      Some (expr, env)
    | _ ->
      (* Continue to browse through the child of [node]. *)
      let node = Browse_tree.of_node ~env node in
      Lazy.force node.t_children |> List.rev
      |> Stdlib.List.find_map (fun node ->
             select_among_child node.Browse_tree.t_env node.t_node)
  in
  nodes |> List.rev
  |> Stdlib.List.find_map (fun (env, node) -> select_among_child env node)
```

src/analysis/refactor_extract_region.ml
voodoos: Ah I see the concern now, you should explain it in the comment :-) 

We probably do similar things in other places, maybe have a look at these ? (we probably just don't care about soundness  in other places too 😅)

src/analysis/refactor_extract_region.ml
voodoos: But `Path.Set.elements` returns elements in order using the polymorphic `compare` function. So you're loosing the initial order anyway, don't you ?

src/analysis/refactor_extract_region.ml
voodoos: All right, not worth it :-)

src/analysis/refactor_extract_region.ml
Tim-ats-d: I like `binding_generator`, I will use it :+1:

src/analysis/refactor_extract_region.ml
Tim-ats-d: I've renamed it and add a comment.

src/analysis/refactor_extract_region.ml
Tim-ats-d: Actually `is_module_bound` checks if a module that have the given path exists in the toplevel env and `is_bound_var` checks that the head of given path is bound.

Tim-ats-d: I think I will rename `is_module_bound` `is_module_bound_in_toplevel_env`

src/analysis/refactor_extract_region.ml
Tim-ats-d: Yes, that's better.

src/analysis/refactor_extract_region.ml
Tim-ats-d: `toplevel_item_source` or maybe `toplevel_item_raw`?

src/analysis/refactor_extract_region.ml
Tim-ats-d: Absolutely, I got a little confused.

src/analysis/refactor_extract_region.ml
voodoos: Thanks, that's much clearer !

Still, I don't think "most inclusive" means what you mean in English :-)
https://www.linguee.com/english-french/search?source=auto&query=most+inclusive

Maybe `largest_expression_between ~start ~stop` would be clearer.

src/analysis/refactor_extract_region.ml
voodoos: `toplevel_item_source` seems reasonable

src/analysis/refactor_extract_region.ml
Tim-ats-d: Totally, I'm going to use a fold.

src/analysis/refactor_extract_region.ml
Tim-ats-d: I found this https://github.com/ocaml/merlin/blob/main/src/analysis/construct.ml#L107 and I'm relieved that the same method is being used. That should do the trick.

Tim-ats-d: I've also added a small comment:
>  ```ocaml
> loop Path.Set.empty node
> |> Path.Set.filter (fun path ->
>         (* Filter identifier that are in Stdlib to avoid cluttering the list
>             of generated parameters.
>             TODO: there probably a more correct way to do this *)
>         Ident.name (Path.head path) <> "Stdlib")
> ```

xvw:
And after `FIXME` are added, I'll also make a last round of review this week!

doc/dev/PROTOCOL.md
xvw: `-start` and `-end` (I think)

src/analysis/refactor_extract_region.ml
```diff
@@ -0,0 +1,493 @@
...
+    | _ ->
+      (Gen.toplevel_function bounded_vars, Gen.fun_apply_params bounded_vars)
+  in
+  let name =
+    match extract_name with
+    | None -> Default { basename = "fun_name" }
+    | Some name -> Fixed name
+  in
+  let remove_path_prefix_of_bound_values expr =
+    (* We need to unquality bound values. Otherwise, the generated call will use
```
xvw: `unqualify` (instead of `unquality`?)

CHANGES.md
```diff
@@ -1,6 +1,8 @@
 unreleased
 ==========
+
   + merlin library
+    - Implement new refactor-extract-region command for extracting region to a fresh let binding (#1546)
```
xvw: Is'nt `#1948`?

xvw (approved):

tests/test-dirs/refactor-extract-region/extraction-issue.t/run.t
```diff
@@ -0,0 +1,77 @@
...
+    "value": {
+      "start": {
+        "line": 3,
+        "col": 0
+      },
+      "end": {
+        "line": 10,
+        "col": 31
+      },
+      "content": "let fun_name1 x y a b c d_x = (((a + b) + ((c * x) * y)) + z) + x
```
voodoos: I have mixed feelings about this. First, is this a minimal example ? Do you need a, b and c to trigger the issue ? 
The more minimal the example is, when reproducing a precise error, the easier it will be to reason about for someone who would like to try and fix it later.

Also it feels like the fix would not be too hard... would it ?

PR #1951, opened 2025-07-22
Add `locate_type_multi` query

liam923:
This PR adds a new command, locate-type-multi, that we plan to use at Jane Street for some new VSCode functionality.

A useful current command in VSCode is "Go to Type Definition", which uses the `locate-type` query to go to the definition of the type of the identifier under the cursor. However, this command is rather limited - it only works on `Tconstrs`, and when the `Tconstr` carries arguments, it just ignores the args. In other words, the command doesn't provide the ability to go to a's defintion if the type is any of the following: `a t, (a * _), a -> _, _ -> a`.

`locate-type-multi`, instead of always just returning the location of a single identifier, returns a tree structure that represents all the identifiers within a type. So for example, if the type is `a -> (b * c)`, it returns the locations of all of `a`, `b`, and `c`, and this is in a tree-like data structure that makes it easy for VSCode to present this to the user.

### DIFF
--- a/src/analysis/locate_types.ml
+++ b/src/analysis/locate_types.ml
@@ -1,0 +1,59 @@
+open Std
+
+module Type_tree = struct
+  type node_data =
+    | Arrow
+    | Tuple
+    | Poly_variant
+    | Object
+    | Type_ref of { path : Path.t; ty : Types.type_expr }
+
+  type t = { data : node_data; children : t list }
+end
+
+let rec flatten_arrow ret_ty =
+  match Types.get_desc ret_ty with
+  | Tarrow (_, ty1, ty2, _) -> ty1 :: flatten_arrow ty2
+  | _ -> [ ret_ty ]
+
+let rec create_type_tree ty : Type_tree.t option =
+  match Types.get_desc ty with
+  | Tarrow (_, ty1, ty2, _) ->
+    let tys = ty1 :: flatten_arrow ty2 in
+    let children = List.filter_map tys ~f:create_type_tree in
+    Some { data = Arrow; children }
+  | Ttuple tys ->
+    let children = List.filter_map tys ~f:create_type_tree in
+    Some { data = Tuple; children }
+  | Tconstr (path, arg_tys, abbrev_memo) ->
+    let ty_without_args =
+      Btype.newty2 ~level:Ident.highest_scope (Tconstr (path, [], abbrev_memo))
+    in
+    let children = List.filter_map arg_tys ~f:create_type_tree in
+    Some { data = Type_ref { path; ty = ty_without_args }; children }
+  | Tlink ty | Tpoly (ty, _) -> create_type_tree ty
+  | Tobject (fields_type, _) ->
+    let rec extract_field_types (ty : Types.type_expr) =
+      match Types.get_desc ty with
+      | Tfield (_, _, ty, rest) -> ty :: extract_field_types rest
+      | _ -> []
+    in
+    let field_types = List.rev (extract_field_types fields_type) in
+    let children = List.filter_map field_types ~f:create_type_tree in
+    Some { data = Object; children }
+  | Tvariant row_desc ->
+    let fields = Types.row_fields row_desc in
+    let children =
+      List.filter_map fields ~f:(fun (_, row_field) ->
+          match Types.row_field_repr row_field with
+          | Rpresent (Some ty) -> create_type_tree ty
+          | Reither (_, tys, _) ->
+            (* If there are multiple types in [tys], they are types that are meant to
+               unify with each other (it'd be a type error if not, see
+               [Ctype.collapse_conj]). So just using the head of the list seems fine
+               (using the entire list results in types being duplicated). *)
+            List.hd_opt tys |> Option.bind ~f:create_type_tree
+          | Rpresent None | Rabsent -> None)
+    in
+    Some { data = Poly_variant; children }
+  | Tnil | Tvar _ | Tsubst _ | Tunivar _ | Tpackage _ | Tfield _ -> None
--- a/src/analysis/locate_types.ml
+++ b/src/analysis/locate_types.ml
@@ -1,14 +1,8 @@
 open Std
 
 module Type_tree = struct
-  type node_data =
-    | Arrow
-    | Tuple
-    | Poly_variant
-    | Object
-    | Type_ref of { path : Path.t; ty : Types.type_expr }
-
-  type t = { data : node_data; children : t list }
+  type type_ref_payload = { path : Path.t; ty : Types.type_expr }
+  type t = type_ref_payload Query_protocol.Locate_types_result.Tree.t
 end
 
 let rec flatten_arrow ret_ty =
--- a/src/commands/new_commands.ml
+++ b/src/commands/new_commands.ml
@@ -500,6 +500,23 @@ let all_commands =
           | #Msource.position as pos ->
             run buffer (Query_protocol.Locate_type pos)
       end;
+    command "locate-types"
+      ~spec:
+        [ arg "-position" "<position> Position to locate the type of"
+            (marg_position (fun pos _ -> pos))
+        ]
+      ~doc:
+        "Locate the declaration of the type of the expression. If the type is \
+         expressed via multiple identifiers, it returns the location of each \
+         identifier."
+      ~default:`None
+      begin
+        fun buffer pos ->
+          match pos with
+          | `None -> failwith "-position <pos> is mandatory"
+          | #Msource.position as pos ->
+            run buffer (Query_protocol.Locate_types pos)
+      end;
     command "occurrences"
       ~spec:
         [ arg "-identifier-at" "<position> Position of the identifier"
--- a/src/commands/query_json.ml
+++ b/src/commands/query_json.ml
@@ -70,6 +70,7 @@ let dump (type a) : a t -> json =
         ("position", mk_position pos)
       ]
   | Locate_type pos -> mk "locate-type" [ ("position", mk_position pos) ]
+  | Locate_types pos -> mk "locate-types" [ ("position", mk_position pos) ]
   | Enclosing pos -> mk "enclosing" [ ("position", mk_position pos) ]
   | Complete_prefix (prefix, pos, kind, doc, typ) ->
     mk "complete-prefix"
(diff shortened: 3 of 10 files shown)


src/frontend/kernel/dune
```diff
@@ -0,0 +1,5 @@
+(library
+ (name        query_protocol_kernel)
+ (public_name merlin-lib.query_protocol_kernel)
+ (libraries yojson)
+ (preprocess (pps ppx_yojson_conv)))
```
liam923: I created this new directory to put useful definitions that are compatible with js_of_ocaml. This is useful for us internally so that a VSCode extension (which is compiled with js_of_ocaml) can depend on the definitions in this directory. Unfortunately `query_protocol.ml` itself can't be compiled with `js_of_ocaml` naively due to some of its dependencies.

xvw:
Oh! That's a great feature!
However, I have a few reservations before embarking on a full review:
- I'm not at all convinced by the name and wonder if `locate-types` wouldn't be... better?
- Currently, Merlin does not depend on Yojson and does not rely on any drifters, and I feel that the current Json model is more than sufficient to avoid including these two somewhat resource-intensive dependencies. What do you think?

src/analysis/locate_types.ml
```diff
@@ -0,0 +1,44 @@
...
+
+module Type_tree = struct
+  type node_data =
+    | Arrow
+    | Tuple
+    | Object
+    | Type_ref of { path : Path.t; ty : Types.type_expr }
+
+  type t = { data : node_data; children : t list }
+end
```
xvw, line 6: Maybe it is related to VSCode but I am not see the point of splitting repr into two types? (Maybe because I do not see cases when we have `children` in `Type_ref`?

xvw: Oh I see, in case of type parameters, in `(t, k) r`: 
- `r` is the type
- `t` and `k` are children?

liam923:
I should also mention that I intend to open a follow-up PR that adds `ppx_jane` as a dependency. So we may as well hash that out while discussing `ppx_yojson_conv`.

sidkshatriya:
Thanks for working on locating multiple (related) type definitions. This is hugely needed in ocamllsp.

Before discussing the approach for OCaml I want to compare with `rust-analyzer`:

Let's say you have a type in Rust:  `Sender<EditorResponse>`

Let's say in your editor you try to locate the type definition of `response` when your cursor is positioned on the word `response` by clicking or using shortcuts corresponding to LSP request `textDocument/typeDefinition`
```rust
let response : Sender<EditorResponse> = ...
```

The response of `rust-analyzer` to the above request is that you are presented with a *list* of options to navigate to:
- Navigate the definition of type `Sender`
- Navigate to the definition of type `EditorReponse`

I have checked and the standard LSP request `textDocument/typeDefinition` is issued. This is not some custom request. The response from the LSP looks like this:

```json
[
  {
    "uri": "file:///home/abczy/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/crossbeam-channel-0.5.14/src/channel.rs",
    "range": {
      "start": {
        "line": 354,
        "character": 11
      },
      "end": {
        "line": 354,
        "character": 17
      }
    }
  },
  {
    "uri": "file:///home/abcxyz/kakoune-lsp/src/types.rs",
    "range": {
      "start": {
        "line": 326,
        "character": 11
      },
      "end": {
        "line": 326,
        "character": 25
      }
    }
  }
]
``` 
i.e. basically the location of the types `EditorResponse` and `Sender`.

**Question:** Why not follow this behavior of rust-analyzer ? Pardon me if my understanding the discussion above is wrong but:
- We are just using the standard `textDocument/typeDefinition`
- We don't need to have any tree complication. Just return the list of related types

This also means that the reply will be understood by editors other than VSCode (e.g. neovim, kakoune etc.)

**So coming back to OCaml.** 

If you try to navigate to the type of a variable that has the type `Queue list` you should be given the option to look the type of `Queue` and `list`. This can ride on the standard `textDocument/typeDefinition`

src/analysis/locate_types.ml
liam923: Yes exactly

sidkshatriya:
@liam923 Thanks for your thoughts !

> We don't yet feel ready to replace the current `textDocument/typeDefinition` implementation with this new feature

It would be cool if you can allow the current `textDocument/typeDefinition` to return a list also (simply linearize the tree) and then hide this new feature behind an environment variable. There are a lot of people using OCaml with emacs / vim / neovim etc. and it would be great if you can include them in this important feature addition too :-) !

liam923:
I expect a lot of Jane Street users will also be excited about using this toggle.

Octachron:
Merlin for new version of the compiler is built with the new version of the compiler. Adding a dependency on ppxlib is already not ideal, and I would argue it is not worth the increased depth of the ecosystem bootstrap just to avoid writing a handful of one-line functions.

Moreover, adding a dependency on a `ppx_jane` feels like an even worse idea to me, since this adds a dependency on `base` . Unfortunately, currently the `base` library does not have a good history of anticipating the update of the compiler in time for new alpha releases of the compiler. Typically, `base` was only made compatible with Ocaml 5.4.0 at version `v17.3` released on the 10th June, while the Merlin-preview for 5.4 was released on the 26th May.

liam923:
I've discussed this with @pitag-ha and @xvw offline and they convinced me the ppxes are a bad idea. I'll update the PR to remove them.

src/frontend/kernel/dune
```diff
@@ -0,0 +1,18 @@
...
+ ; https://github.com/ocaml-ppx/ppxlib/issues/342
+ ; But in the foreseeable future we will not use any PPXes normally, so this is a
+ ; non-issue.
+ (lint
+  (pps ppx_yojson_conv)))
+
+; [@@deriving_inline] gets expanded by the `@lint` dune alias. This makes `@lint` get run
+; when we run the `@runtest` alias, which aides development.
+(alias
+  (name runtest)
```
liam923: It would be convenient for this to get run as part of the `default` alias. But there's two things that give me pause:
1. I'm not sure if that will make opam try to run the ppx during compilation.
2. Maybe this is bad practice since `ppx_yojson_conv` is a `with-test` dependency, and this would make that a bit of a lie.

src/commands/query_json.ml
```diff
@@ -70,6 +70,7 @@ let dump (type a) : a t -> json =
         ("position", mk_position pos)
       ]
   | Locate_type pos -> mk "locate-type" [ ("position", mk_position pos) ]
+  | Locate_type_multi pos -> mk "locate-types" [ ("position", mk_position pos) ]
```
xvw: WDYT about using `Locate_types` instead of `Locate_type_multi`?

xvw:
Hi! 
The current patch seems to still depends/introduces PPX. Are you plan to remove it?
Thanks in advance!

liam923:
> Hi! The current patch seems to still depends/introduces PPX. Are you plan to remove it? Thanks in advance!

See my above comment about using `@@deriving_inline`, which I think addresses a lot of the hestitations we had with using a ppx. But let me know if you think that still would be problematic.

src/frontend/query_commands.ml
```diff
@@ -351,6 +351,76 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a = function
...
     end
+  | Locate_types pos -> (
+    let typer = Mpipeline.typer_result pipeline in
+    let verbosity = verbosity pipeline in
+    let local_defs = Mtyper.get_typedtree typer in
+    let structures = Mbrowse.of_typedtree local_defs in
+    let pos = Mpipeline.get_lexing_pos pipeline pos in
+    let result =
+      let ( let* ) opt f = Option.bind opt ~f in
+      let ( let+ ) opt f = Option.map opt ~f in
```
xvw: Could you move `let*` and `let+` (and maybe `and+`) into `util/std.ml` inside `Option.Infix` and just locally open `Option.Infix` here?

xvw (approved):
I just have a small nitpicking but I approve the PR. 
Thanks **a lot** for the feature (and for being understanding about PPXs)

src/frontend/query_commands.ml
liam923: Ah yes good call

xvw (approved):
Thanks!
(cc @pitag-ha)

src/frontend/query_commands.ml
```diff
@@ -351,6 +351,75 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a = function
...
+            let result =
+              match
+                Locate.from_path ~config ~env ~local_defs ~namespace:Type path
+              with
+              | `Builtin (_, s) -> `Builtin s
+              | `Not_in_env _ as s -> s
+              | `Not_found _ as s -> s
+              | `Found { file; location; _ } ->
+                `Found (Some file, location.loc_start)
+              | `File_not_found result -> `File_not_found result.file
```
voodoos, line 409: Do we need to perform that additional mapping here ? It could be done in one pass when `Query_json` serializes it, as long as we pass along enough information.

src/frontend/query_protocol.ml
```diff
@@ -133,6 +133,27 @@ type occurrences_status =
...
+            | `Builtin of string
+            | `Not_in_env of string
+            | `File_not_found of string
+            | `Not_found of string * string option ]
+        }
+
+  type type_tree = { data : node_data; children : type_tree list }
+
+  type t = Success of type_tree | Invalid_context
+end
```
voodoos, line 154: At first I was a bit sad to see this type repeated in `Locate_type`, and thought it would be better to parametrize the tree. But after making the change in I am not sure if it's actually "better". WDYT ? (and @xvw ?)

My attempt: https://github.com/voodoos/merlin/commit/d7cb643668fc4c3f888833e39c51096b0ce99c52

voodoos (approved):
Thanks a lot @liam923, that does sounds like a useful feature. Additionally, I suspect the client work for this will also be reusable for when we will finally propose "locate declarationS".

Do you already have such work done on your LSP server / client ?

src/commands/new_commands.ml
```diff
@@ -500,6 +500,23 @@ let all_commands =
           | #Msource.position as pos ->
             run buffer (Query_protocol.Locate_type pos)
       end;
+    command "locate-types"
```
voodoos, line 503: This new command should be documented in https://github.com/ocaml/merlin/blob/main/doc/dev/PROTOCOL.md

src/frontend/query_protocol.ml
xvw: Even if it's just a minor detail, I also prefer @voodoos's approach (parametric).

src/frontend/query_commands.ml
liam923: The reason for this was to make `Locate_types_result.t` a type that can round trip back from json easily. That way, clients of Merlin can use `Locate_types_result.t` to represent the parsed result. (We make use of this fact in our internal client implementation.)

liam923:
> Do you already have such work done on your LSP server / client ?

We have, but we don't think that it's work that we can easily contribute. We didn't make any lsp changes to support this feature. Rather, we made our internal VSCode extension use the [call-compatible lsp request](https://github.com/ocaml/ocaml-lsp/blob/d44b27dfc37a4e78294c99b8b9d5bdf8758eba83/ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.mli) to directly call the Merlin query. My understanding is that our extension is substantially different from the OCaml Platform extension, so unfortunately I don't think there's a staightforward way to share those changes.

To give some more details of the user experience:
1. The user runs a vscode command that runs this new Merlin command.
2. A modal pop-up displays the result of the call.
3. The user can select which type they want to jump to.
Here's some screenshots to give an idea. This is displaying the type:
```
Mpipeline.t ->
[> `Assoc of (modname * [> `Int of int ]) list
 | `Int of int
 | `String of modname ]
list -> json
```
but using an older version of this command that doesn't support polymorphic variants.
image
image

src/frontend/query_protocol.ml
liam923: Thanks, I pushed a change to make this parametric.

src/commands/new_commands.ml
liam923: I've done so

liam923:
Great, thanks! Is this ready to merge then, or are there other changes I should make? (I don't have the power to click merge.)

voodoos:
We're all good, I was just sleeping for the CI !

PR #1952, opened 2025-07-23
Refactor some types into `query_protocol_kernel`

liam923:
This PR is difficult for me to motivate as it makes much more sense in the https://github.com/oxcaml/merlin context, but I figured it would be worth opening for the sake of keeping that fork in line with the main fork.

The main goal of this PR is to move some more type definitions (as well as some serializers/deserializers for them) into `query_protocol_kernel`. This is a new directory that I introduced in #1951 that facilitates sharing code between Merlin and VSCode extensions by being js_of_ocaml compatible.

This also replaces some functions with ppx-derived functions. This involves adding `ppx_jane` as a dependency, which I expect to be controversial. There's a related discussion in-progress in #1951 about `ppx_yojson_conv`, so I expect the resolution of that discussion to inform this one.

## Reviewing Note

These changes are based on the changes from #1951 as they overlap slightly, but I pointed the PR at the main branch so that this PR can live in the oxcaml/merlin repo rather than my personal fork. Thus, the commits up to and including f71ee32b7128278a62e72432bc32333508484e79 should be ignored while reviewing.


liam923:
I've realized that the `ppx_jane` question is actually a moot point as `ppx_yojson_conv` depends on `ppx_jane`.

liam923:
Based on our offline conversation about not using ppxes in Merlin, I don't think this change is worth the complexity.

PR #1956, opened 2025-08-14
Fix `merlin_reader` functionality on OpenBSD

omegametabroccolo:
OpenBSD's [`execve(2)`](https://man.openbsd.org/execve#EINVAL) needs `argv` to contain at least one element.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+unreleased
+==========
+  + merlin library
+    - Fix `merlin_reader` for OpenBSD (#1956)
+
 merlin 5.5
 ==========
 Tue Jun 24 16:10:42 CEST 2025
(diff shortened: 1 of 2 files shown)


xvw:
Hi! 
I am not really familiar with that, does the fix can have issue on other unix distro?

omegametabroccolo:
`Unix.create_process` uses [`caml_unix_spawn`](https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/spawn.c) from the runtime, which can use either [`posix_spawn*`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_spawn.html) or [`exec*`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/exec.html) depending on availability.

From the `exec*` POSIX manpage:

> The value in argv[0] should point to a filename string that is associated with the process being started by one of the exec functions.

Same thing from the `posix_spawn*` manpage:

> The value in argv[0] should point to a filename string that is associated with the process image being started by the posix_spawn() or posix_spawnp() function.

And since merlin readers are currently spawned with no arguments, I assume they just ignore their arguments so this fix shouldn't cause problems on other Unixes.

voodoos (approved):
Thanks @omegametabroccolo Merlin should respect the calling conventions, if a reader is impacted by this it should be fixed too.

@omegametabroccolo can you add a changelog entry ?

omegametabroccolo:
Done! @voodoos

PR #1962, opened 2025-09-08
Test & Fix recovery issue 1953

### LINKED ISSUE #1953: Missing type recovery
Type recovery at the level of structure items is lackluster: mostly it just drops the item if an error happens at the level of the item, which is painful for blocks of recursive definitions.

For instance:
```ocaml
let rec foo x = x + a + b
and (a,b) = (1,2)
```
The whole block is dropped.

The are a couple of similar errors in that area, which could be gracefully recovered from. But mostly I think recursive blocks should get special treatment where each definition gets recovered/dropped separately from the others.

voodoos:
Fixes #1953.

The classic tell of the missing `raise_error`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@ unreleased
 ==========
   + merlin library
     - Fix `merlin_reader` for OpenBSD (#1956)
+    - Improve recovery of mutually recursive definitions (#1962, fixes #1953)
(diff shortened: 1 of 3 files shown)


xvw (approved):
I already have it into my current branch of OCaml!
Thanks!

PR #1972, opened 2025-10-02
Introduce `buffer-cache` flag to disable typedtree cache

Tim-ats-d:
Use `-buffer-cache=no` to disable it.

cc @pitag-ha @xvw @lyrm

### DIFF
--- a/src/kernel/mconfig.ml
+++ b/src/kernel/mconfig.ml
@@ -377,7 +377,7 @@ let merlin_flags =
       " DEPRECATED" );
     ( "-use-typer-cache",
       Marg.bool (fun b merlin -> { merlin with use_typer_cache = b }),
-      "Whether to enable the typer cache. Default to true." )
+      "Whether to enable the typer cache. Defaults to true." )
   ]
 
 let query_flags =
--- a/src/kernel/mconfig.ml
+++ b/src/kernel/mconfig.ml
@@ -94,7 +94,8 @@ type merlin =
     flags_applied : string list with_workdir list;
     failures : string list;
     extension_to_reader : (string * string) list;
-    cache_lifespan : int
+    cache_lifespan : int;
+    use_typer_cache : bool
   }
 
 let dump_merlin x =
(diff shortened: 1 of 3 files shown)


src/kernel/mconfig.ml
```diff
@@ -94,7 +94,8 @@ type merlin =
     flags_applied : string list with_workdir list;
     failures : string list;
     extension_to_reader : (string * string) list;
-    cache_lifespan : int
+    cache_lifespan : int;
+    buffer_cache : bool
```
voodoos: Since the internal name is `typer cache` let's reuse it ! It's also a more informative name, since other kind of buffer-related caches are not disabled by that option. And there is also an existing `use_ppx_cache`, so let's reuse the same clearer format:

```suggestion
    use_typer_cache : bool
```

src/kernel/mconfig.ml
```diff
@@ -373,7 +374,10 @@ let merlin_flags =
     ( (* Legacy support for janestreet. Ignored. To be removed soon. *)
       "-attributes-allowed",
       Marg.unit_ignore,
-      " DEPRECATED" )
+      " DEPRECATED" );
+    ( "-buffer-cache",
+      Marg.bool (fun b merlin -> { merlin with buffer_cache = b }),
+      "Whether to enable buffer cache. Default to true." )
```
voodoos:
```suggestion
    ( "-use-typer-cache",
      Marg.bool (fun b merlin -> { merlin with use_typer_cache = b }),
      "Whether to enable the typer cache. Default to true." )
```

src/kernel/mconfig.ml
Tim-ats-d: Done

src/kernel/mconfig.ml
Tim-ats-d: Done

src/kernel/mconfig.ml
```diff
@@ -373,7 +374,10 @@ let merlin_flags =
     ( (* Legacy support for janestreet. Ignored. To be removed soon. *)
       "-attributes-allowed",
       Marg.unit_ignore,
-      " DEPRECATED" )
+      " DEPRECATED" );
+    ( "-use-typer-cache",
+      Marg.bool (fun b merlin -> { merlin with use_typer_cache = b }),
+      "Whether to enable the typer cache. Default to true." )
```
voodoos:
```suggestion
      "Whether to enable the typer cache. Defaults to true." )
```

PR #1978, opened 2025-10-15
Fix completion not working for inlined records labels

### LINKED ISSUE #1977: Completion fails on inlined record field
```
$ cat inlined_record.ml
type t = Constr of { some_lbl: int }

let foo (x : t) =
  match x with
  | Constr { som }

❯ ocamlmerlin single complete-prefix -position 5:16 -prefix 'some' -filename inlined_record.ml < ./inlined_record.ml | jq .value
{
  "entries": [],
  "context": null
}

❯ cat indirect.ml
type r = { some_lbl: int }
type t = Constr of r

let foo (x : t) =
  match x with
  | Constr { som }

❯ ocamlmerlin single complete-prefix -position 6:16 -prefix 'some' -filename indirect.ml < ./indirect.ml | jq .value
{
  "entries": [
    {
      "name": "some_lbl",
      "kind": "Label",
      "desc": "r -> int",
      "info": "",
      "deprecated": false
    }
  ],
  "context": null
}
```

There's not much in the log:
- ```
  # 0.01 Completion - fold_sumtype_constructors
  node type: t.Constr
  ```
  for the first case
- ```
  # 0.01 Completion - fold_sumtype_constructors
  node type: r
  ```
  for the second

voodoos:
Fixes #1977 

cc @Lucccyo I tried to show the debugging process if you look at this PR commit by commit. However this was not a completely trivial fix.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+unreleased
+==========
+
+  + merlin library
+    - Fix completion not working for inlined records labels (#1978, fixes #1977)
+
 merlin 5.6
 ==========
 Sat Oct 04 15:10:42 CEST 2025


src/analysis/completion.ml
```diff
@@ -452,14 +452,39 @@ let get_candidates ?get_doc ?target_type ?prefix_path ~prefix kind ~validate env
...
+            Format.pp_print_option Pprintast.longident fmt prefix_path);
+        let consider_label ({ lbl_name = name; _ } as l) candidates =
+          log ~title:"get_candidate" "Found label %s in env" name;
+          if not (validate `Lident `Label name) then candidates
+          else
+            make_weighted_candidate ~exact:(name = prefix) name (`Label l)
+              ~attrs:(lbl_attributes l)
+            :: candidates
+        in
+        let inlined_record_labels =
```
xvw, line 467: You know that for me, it is `Option.map` :) but it is just nitpicking :3

xvw: (It deserve a change entry :))

voodoos: Just nitpicking, but that's `Option.bind` 😁

PR #1988, opened 2025-11-04
Make short paths lazier

liam923:
This PR brings in changes from oxcaml Merlin: https://github.com/oxcaml/merlin/pull/193


voodoos (approved):
Thanks @liam923, looks great ! Do you have more insights on the potential performance improvements of this patch ?

Also, if you can add a changelog entry ? If not I will do it before merging (I restarted the windows CI).

PR #1989, opened 2025-11-05
Add --lalr flag to menhir invocations

liam923:
This is a flag that [the compiler passes to menhir](https://github.com/ocaml/ocaml/blob/d197927a89e08d84cbd030ae3112295b31de3f18/Makefile.menhir#L67) as the compiler's parser (and also merlin's) is LALR. This flag enables Menhir to generate more optimal code, so it is a strict improvement given the parser is already LALR. Credit to @let-def for pointing this out to @ccasin, who in turn pointed this out to me.

This corresponds to OxCaml Merlin [PR 194](https://github.com/oxcaml/merlin/pull/194).

### DIFF
--- a/src/ocaml/preprocess/dune
+++ b/src/ocaml/preprocess/dune
@@ -10,7 +10,7 @@
  (modules parser_raw)
  (enabled_if (<> %{profile} "release"))
  (mode    (promote (only parser_raw.ml parser_raw.mli)))
- (flags :standard --inspection --table --cmly))
+ (flags :standard --inspection --table --cmly --lalr))
 
 (rule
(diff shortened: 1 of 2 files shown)


voodoos (commented):
Thanks! I guess we should update the generated parser files that are promoted to the sources then, they must be different?

(looks like our CI for that is not working as intended...)

liam923:
Hm, I expected them to be different, but I think they're not? When I build locally dune isn't promoting any changes. This is mildly surprising to me because it caused changes in the oxcaml merlin parser, but not so surprising that I think I'm doing something wrong.

PR #1990, opened 2025-11-06
Refactor index caching. Get rid of the stamping.

voodoos:
@liam923 not what you expected, but at least the use of the cache is clearer now. I don't know why I opted for the imperative stamped hashtbl in the initial version. This version relies on sharing and integrates much more naturally with the typer cache.

If the next step is to have "lazy" indexing, I think the most sensible approch would be to add a pipeline phase after typing.


liam923 (approved):
Thanks, this change definitely seems like an improvement and makes the cache a lot easier to reason about.

PR #1991, opened 2025-11-07
Only index the buffer if the query needs it

voodoos:
The refactor in https://github.com/ocaml/merlin/pull/1990 and @liam923 's suggestion made it trivial to lazily index buffers.


liam923 (approved):
Awesome thanks for this. I've done some testing and can verify that this gets a ~25% speedup in a `ocamlmerlin single errors` query on the file that initially motivated this change.

PR #1994, opened 2025-11-12
Some more short-path tests

xvw:
Here's a first batch of test.

### DIFF
--- a/tests/test-dirs/short-paths/short-paths-simple.t
+++ b/tests/test-dirs/short-paths/short-paths-simple.t
@@ -1,7 +1,6 @@
 Simple alias on string
 
-  $ echo "" > .merlin
-  > cat > sample.ml <<EOF
+  $ cat > sample.ml <<EOF
   > type t = string
   > let f = String.lowercase_ascii
   > EOF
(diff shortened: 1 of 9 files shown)


tests/test-dirs/short-paths/short-paths-simple.t
```diff
@@ -0,0 +1,91 @@
...
+
+  $ echo "FLG -short-paths" > .merlin
+  > cat > sample.ml <<EOF
+  > type t = string
+  > let f = String.lowercase_ascii
+  > EOF
+  $ $MERLIN single type-enclosing -position 2:5 -filename sample.ml < sample.ml \
+  > | tr '\r\n' ' ' \
+  > | jq .value[0].type -r
+  t -> t
```
voodoos: Is there a good reason why you overwrites the test file `sample.ml` with the exact same content ?

If not, you can systematically remove the second one 🙂

(You could also use the `-short-paths` command line argument directly, instead of writing a `.merlin` file but I don't mind keeping it the ways you did it.)

voodoos (approved):
Thanks ! Are some of these results currently unsatisfying ? If so you should mark them.

Also I don't know why you always duplicate the test source files, it bloats the tests a lot.

tests/test-dirs/short-paths/short-paths-simple.t
voodoos:
```suggestion
  $ echo "FLG -short-paths" > .merlin
  $ $MERLIN single type-enclosing -position 2:5 -filename sample.ml < sample.ml \
  > | tr '\r\n' ' ' \
  > | jq .value[0].type -r
  t -> t
```

PR #1997, opened 2025-11-17
Signature help should not appear on the function name

Lucccyo:
Merlin displays a `signature-help` response on every part of a function call, including the function name.
However, `signature-help` should appear only on function parameters.
The `signature-help` command should not return an answer for function names.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@ unreleased
 ==========
 
   + merlin library
+    - Signature help should not appear on the function name (#1997)
     - Fix completion not working for inlined records labels (#1978, fixes #1977)
     - Perform buffer indexing only if the query requires it (#1990 and #1991)
     - Stop unnecessarily forcing substitutions when initializing short-paths graph (#1988)


PizieDust:
Thank you for the PR

src/kernel/msource.ml
```diff
@@ -105,6 +105,14 @@ let get_logical { text } = function
     done;
     `Logical (!line, offset - !cnum)
 
+let compare_logical x y : int =
```
PizieDust: Just a little nitpicking, please feel free to ignore this..

```ocaml
let compare_logical (`Logical ((row_x: int), (col_x : int))) (`Logical ((row_y:int), (col_y: int))) =
  compare (row_x, col_x) (row_y, col_y)
```

src/kernel/msource.mli
```diff
@@ -27,6 +27,8 @@ val get_offset : t -> [< position ] -> [> `Offset of int ]
 
 val get_logical : t -> [< position ] -> [> `Logical of int * int ]
 
+val compare_logical : [> `Logical of int * int ] -> [> `Logical of int * int ] -> int
```
PizieDust:
```ocaml
val compare_logical : [< `Logical of int * int ] -> [< `Logical of int * int ] -> int
```

src/frontend/query_commands.ml
```diff
@@ -898,16 +898,19 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a = function
...
-        let fun_name = Option.value ~default:"_" s.function_name in
-        sprintf "%s : " fun_name
-      in
-      Some
-        { label = prefix ^ s.signature;
-          parameters = List.map ~f:(param (String.length prefix)) s.parameters;
-          active_param = Option.value ~default:0 s.active_param;
-          active_signature = 0
-        }
+      if (Msource.compare_logical (Msource.get_logical source position) (Msource.get_logical source s.function_position)) < 0 then None
```
voodoos: This line being too long hints that the api for `Msource.compare_logical` is not great.
Let's have a `Msource.compare_position : Msource.position -> Msource.position -> int` instead.

You can also put that as a guard in the pattern matching, I think it will make the intent clearer.

Lucccyo: Ok, thank you.
A naïve question: what is the difference between `position` and `[< position]`?
More generally, what do `[< ...]` or `[> ...]` means ?

voodoos: https://ocaml.org/manual/5.1/polyvariant.html 🙂

Lucccyo: Thank you !

tests/test-dirs/signature-help/issue_fun_name.t
```diff
@@ -0,0 +1,147 @@
...
+        }
+      ],
+      "activeParameter": 0,
+      "activeSignature": 0
+    },
+    "notifications": []
+  }
+
+
+FIXME: Signature help should not appear on the name of the function:
```
voodoos: This is not a FIXME anymore, is it ?

Also, it's a bit confusing that this test and the next one appear after you wrote the file `t.ml` but it relies on the file `test.ml` is that on purpose ? Let's move it closer to the concerned file :-)

tests/test-dirs/signature-help/issue_fun_name.t
```diff
@@ -0,0 +1,147 @@
...
+              ]
+            }
+          ]
+        }
+      ],
+      "activeParameter": 0,
+      "activeSignature": 0
+    },
+    "notifications": []
+  }
```
voodoos: Let's make these tests less verbose by filtering with `jq`. For exampel `jq '.value.signatures[0] | {label,activeParameter}'` ?

PR #1998, opened 2025-11-19
Minor patches for the C code in Merlin

MisterDA:
I did a review of `ocamlmerlin.c`. I found minor and relatively major issues:
1. Erroneous size computation of the `struct sockaddr_un`. On some BSD systems the `struct sockaddr` family of types have a leading `sxx_len` field. It was not taken into account. Use `offsetof`, as described by POSIX, to compute the socklen.
   https://man.freebsd.org/cgi/man.cgi?unix(4)
   https://man7.org/linux/man-pages/man7/unix.7.html
   https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_un.h.html

2. Functions from WinAPI don't set `errno`, and their failure code can be retrieved with `GetLastError` and the error string with `FormatMessage`. Functions from the CRT do set `errno`, and `perror` can be used to display the error string.

3. Structures for passing file descriptor using CMSG and Unix domain sockets were not properly aligned, potentially leading to inefficient or buggy behavior. 

There's also a few patches that improve the code quality, in my opinion. The code now assumes a C99 compiler, and makes use of C23 and C2y features or extensions, if available. For reference, the OCaml runtime requires at least C11.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ==========
 
+  + merlin binary
+    - Fix a plethora of minor issues with the C code (#1998)
   + merlin library
     - Signature help should not appear on the function name (#1997)
     - Fix completion not working for inlined records labels (#1978, fixes #1977)


CHANGES.md
```diff
@@ -8,6 +8,8 @@ unreleased
   + test suite
     - Add a test to ensure the behavior showed in issue #1517 is no longer relevant (#1995)
     - Add a test to ensure the code fragment exhibited in issue #1118 no longer makes Merlin crash (#1996)
+  + frontend
```
voodoos:
```suggestion
  + merlin binary
```

voodoos (approved):
We reviewed this with Antonin and I am confident that this is a very nice improvement to our C frontend. The main features are the correction of all warnings, and improved Windows compatibility.

Thanks a lot @MisterDA !

Issue #1949, opened 2025-07-02
Inconsistent type variables from type-enclosing


liam923:
Here's a low priority issue. In this program:
```
let foo a b = b
```
If you hover `foo`, Merlin will tell you its type is `'a -> 'b -> 'b`. But when you hover either `'b`, Merlin will tell you its type is `'a`, which is inconsistent with the previous hover. This can be a confusing user experience.

PR #2004, opened 2025-11-20
Add a test to ensure the illustrate the behavior of `type-enclosing` command showed in issue #1949

Tim-ats-d:
This PR adds a test to illustrate the issue https://github.com/ocaml/merlin/issues/1949 opened by Liam.
I don't know if it deserves to be fixed or if it is planned to be fixed but in doublt. What do you think @voodoos?

### DIFF
--- a/tests/test-dirs/type-enclosing/issue1949.t
+++ b/tests/test-dirs/type-enclosing/issue1949.t
@@ -1,0 +1,11 @@
+  $ cat >main.ml <<EOF
+  > let foo a b = b
+  > EOF
+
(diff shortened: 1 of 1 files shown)


xvw:
I think that here, the main issue is that (where the comment is `inlay hint`: 

```ocaml
(* 'a -> 'b -> 'b *) 
let f x y = y
```

But if you hover `y` or `x`, it will be `'a` because the type expression is different. So I guess that, since the HoverProvider is in OCaml-LSP, we should probably have a test on OCaml-LSP. (Even I think that this is not something that we can easily fix, or maybe resolving types accros enclosings)

PR #2009, opened 2025-11-25
Signature-help should trigger on unfinished `let ... in` bindings

Lucccyo:
Signature-help does not trigger on unfinished `let … in` bindings.
Nothing shows up if the `in` keyword is missing, and it works as expected if it is present.
This PR adds a test case to demonstrate this wrong behaviour, and a fix has been implemented in https://github.com/ocaml/merlin/pull/2036.

### DIFF
--- a/.github/workflows/emacs-lint.yml
+++ b/.github/workflows/emacs-lint.yml
@@ -29,6 +29,6 @@ jobs:
       with:
         version: ${{ matrix.emacs_version }}
 
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Run tests
       run: 'cd emacs && ./check.sh'
--- a/.github/workflows/fuzzy-ci-privileged.yml
+++ b/.github/workflows/fuzzy-ci-privileged.yml
@@ -21,7 +21,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       - name: Download instruction artifact
         env:
           GH_API_ACTIONS: https://api.github.com/repos/${{ github.repository }}/actions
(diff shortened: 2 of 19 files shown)


tests/test-dirs/signature-help/issue_let_in_binding.t
```diff
@@ -0,0 +1,50 @@
+  $ cat > test1.ml <<'EOF'
+  > let _ =
+  > let f = List.map   in
+  > let _ = 5 in
+  > 7
+  > EOF
+
+  $ $MERLIN single signature-help -position 2:17 -filename test < test1.ml
```
voodoos:
```suggestion
  $ $MERLIN single signature-help -position 2:17 -filename test1 < test1.ml
```

(not an issue in this context, but it can be in others, the filename should match the file name.

tests/test-dirs/signature-help/issue_let_in_binding.t
```diff
@@ -0,0 +1,50 @@
...
+              ]
+            }
+          ]
+        }
+      ],
+      "activeParameter": 0,
+      "activeSignature": 0
+    },
+    "notifications": []
+  }
```
voodoos: We often try to reduce the output by filtering with `jq`. 
For example you can use `jq '.value.signatures[0].label'` here, since we are only interested in the fact that there is a result.  `jq '.value.signatures[0]` is also a reasonable choice to also check that result.

tests/test-dirs/signature-help/issue_let_in_binding.t
```diff
@@ -0,0 +1,50 @@
...
+
+FIXME: Signature-help does not trigger on unfinished let ... in bindings.
+  $ cat > test2.ml <<'EOF'
+  > let _ =
+  > let f = List.map
+  > let _ = 5 in
+  > 7
+  > EOF
+
+  $ $MERLIN single signature-help -position 2:17 -filename test < test2.ml
```
voodoos:
```suggestion
  $ $MERLIN single signature-help -position 2:17 -filename test2 < test2.ml
```

PR #2011, opened 2025-12-05
Update ocamlformat to the latest version

Lucccyo:
This PR aims to update the version of `ocamlformat` from `0.27.0` to the latest `0.28.1`.


voodoos:
Thanks @Lucccyo... however there is already a PR that does that: https://github.com/ocaml/merlin/pull/1984
It is not merged yet because there is a bug in 1.28.1 formatting that impacts our codebase in many many places.

CC https://github.com/ocaml-ppx/ocamlformat/issues/2741
Hopefully 1.28.2 should arrive soon :-)

Issue #1980, opened 2025-10-21
Jump to definition: jumps to decl inside recursive modules


trefis:
- `/tmp/test.ml`:
  ```ocaml
  module rec M : sig
    val f : unit -> unit
  end = struct
    let f () = ()
  end
  
  and N : sig
    val foo : unit -> unit
  end = struct
    let foo () = M.f ()
  end
  
  let foo () = M.f ()
  ```
- tests:
```
$ ocamlmerlin single locate -position 10:18 -look-for implementation -filename test.ml < /tmp/test.ml | jq .value
{
  "file": "test.ml",
  "pos": {
    "line": 2,
    "col": 6
  }
}
$ ocamlmerlin single locate -position 13:16 -look-for implementation -filename test.ml < /tmp/test.ml | jq .value
{
  "file": "test.ml",
  "pos": {
    "line": 4,
    "col": 6
  }
}
```

voodoos:
So, of course it's linked to shapes of recursive modules:
```
 # 0.02 locate - shape_of_path
 initial: M<Test.0> . "f"[value]

 # 0.02 locate - shape_of_path
 reduced: Missing uid

 # 0.02 locate - from_path
 No definition uid, falling back to the declaration uid: Test.4
```

PR #2017, opened 2025-12-19
Add reproduction case for #1980

Lucccyo:
This PR aims to reproduce #1980.


xvw:
Should the test be fixed ? 
If so, can you add a `FIXME: comment` to explain future fixes?

Issue #1987, opened 2025-11-03
Locate command fails on multi-line type definitions


Lucccyo:
When the type is defined on multiple lines and the cursor is on its definition, calling `merlinCallCompatible/locate` with either `implementation` or `interface` for `-look-for` fails because the environment is not set up.

PR #2020, opened 2026-01-02
Add reproduction case for #1987

Lucccyo:
This PR reproduces the buggy behaviour of #1987.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,14 @@
 unreleased
 ==========
-  + test suite
-    - Add a test to reproduce [Locate command fails on multi-line type definitions](#1987)
+
+  + merlin library
+    - Add "Other" variant to locate-types result (#2025)
+    - Don't include `option` in locate-types result for optional parameters (#2027)
+    - Fix record field autocompletion (#2028)
+    - Signature help should not loop over the parameters once it is finished (#2023)
+    - Add a reproduction case for #1214, the issue has been resolved before (#2022)
+    - Add reproduction case for #1763 but it is not failing anymore (#2021)
+    - Add a test to reproduce [Locate command fails on multi-line type definitions](#1987) (#2020)
 
 merlin 5.6.1
 ============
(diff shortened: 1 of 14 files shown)


CHANGES.md
```diff
@@ -16,6 +16,8 @@ Sat Dec 20 11:15:42 CET 2025
     - Add a test case illustrating how a snippet produces two unrelated errors in issue #2000. (#2003)
     - Add a test reproducing issue #1983 where `document` command which sometime concatenates consecutive variants and labels (#2005)
     - Signature-help should trigger on unfinished `let ... in` bindings (#2009)
+    - Add a test to reproduce [Locate command fails on multi-line type definitions](#1987)
+
```
voodoos: This is not the correct section of the changelog, since 5.6.1 was released recently we should use a new "unreleased" section.

tests/test-dirs/issue1987.t
```diff
@@ -0,0 +1,15 @@
+  $ cat > test.ml <<EOF
+  > type t =
+  > A | B
+  > let f = Fun.id
+  > EOF
+
+  $ ocamlmerlin single locate -position 3:0 -prefix List.hd -look-for interface -filename test.ml < test.ml | jq .value | jq .pos
```
voodoos: We should always use `$MERLIN` instead of `ocamlmerlin` in tests (here and in your other PRs). It's a wrapper that perform a number of sanitization steps.

tests/test-dirs/issue1987.t
```diff
@@ -0,0 +1,15 @@
...
+  {
+    "line": 76,
+    "col": 4
+  }
+
+-- FIXME : locate should also points to the list interface file
+  $ ocamlmerlin single locate -position 2:0 -prefix List.hd -look-for interface -filename test.ml < test.ml | jq .value
+  "Already at definition point"
```
voodoos: Interesting. I guess this is another example of the more generic issue: Merlin's environment before the end of the first structure item is Empty when it should be the initial env.

This is the most probable cause for that issue, you can add it to the test's comment.

Issue #1763, opened 2024-05-10
Incorrect module type from `type-enclosing`


liam923:
With `test.ml` having contents:
```
include struct
  open struct
    let hidden = 3
  end
  let visible = 4
end
```
I get the following unexpected behavior:
```
$ ocamlmerlin single type-enclosing -position 6:0 < scratch/test.ml | jq '.value[].type'
"sig val hidden : int val visible : int end"
"sig val visible : int end"
```
Although the correct type is included in the returned list, in VSCode with the OCaml Platform extension (I haven't tested other editors), the incorrect `sig val hidden : int val visible : int end` module type is shown.

Note that this error does not occur with the following similar programs:
```
module Foo = struct
  let hidden = 3
end

include struct
  open Foo
  let visible = 4
end
```
```
$ ocamlmerlin single type-enclosing -position 8:0 < scratch/test.ml | jq '.value[].type'
"sig val visible : int end"
```

```
include struct
  module Foo = struct
    let hidden = 3
  end
  open Foo
  let visible = 4
end
```
```
$ ocamlmerlin single type-enclosing -position 7:0 < scratch/test.ml | jq '.value[].type'
"sig module Foo : sig val hidden : int end val visible : int end"
```

Version: `9afb03689bc500fed0f18d612f3b13ad668e2455` (`master` at the time of creating this)

voodoos:
That's surprising, thanks for the report.

PR #2021, opened 2026-01-02
Add reproduction case for #1763

Lucccyo:
This PR aims to reproduce the bug described in [#1763](https://github.com/ocaml/merlin/issues/1763).
However, it appears to be resolved because the error is no longer showing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,13 @@
 unreleased
 ==========
-  + test suite
-    - Add reproduction case for #1763 but it is not failing anymore
+
+  + merlin library
+    - Add "Other" variant to locate-types result (#2025)
+    - Don't include `option` in locate-types result for optional parameters (#2027)
+    - Fix record field autocompletion (#2028)
+    - Signature help should not loop over the parameters once it is finished (#2023)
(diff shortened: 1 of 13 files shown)


tests/test-dirs/issue1763.t
```diff
@@ -0,0 +1,39 @@
...
+  >   let visible = 4
+  > end
+  > EOF
+  "sig val visible : int end"
+
+
+
+
+
+
```
voodoos: That's a lot of newlines, one is enough :-)

CHANGES.md
```diff
@@ -16,6 +16,7 @@ Sat Dec 20 11:15:42 CET 2025
     - Add a test case illustrating how a snippet produces two unrelated errors in issue #2000. (#2003)
     - Add a test reproducing issue #1983 where `document` command which sometime concatenates consecutive variants and labels (#2005)
     - Signature-help should trigger on unfinished `let ... in` bindings (#2009)
+    - Add reproduction case for #1763 but it is not failing anymore
```
voodoos: Same as in the other PRs, you can do the change only in #2020 and then rebase and update you other PRS once it is merged.

Issue #1214, opened 2020-12-02
Locate failing in presence of generalized open


lpw25:
Given `foo.ml`:
```ocaml
module Make (X : sig end) = struct
  module N = struct end
end
```
and `bar.ml`:
```ocaml
module M = struct end
open Foo.Make (M)
module O = N
```
Calling `merlin-locate` on the use of `N` in `bar.ml` will result in the error message:
```
apply: "Needed cmti file of module 'N' to locate 'N' but it is not present"
```

If you give a name to the opened module:
```ocaml
module M = struct end
module P = Foo.Make (M)
open P
module O = N
```
then it works as expected: jumping to the definition of `N` in `foo.ml`.

PR #2022, opened 2026-01-02
Add reproduction case for #1214

Lucccyo:
This PR aims to reproduce the bug described in #1214.
However, it appears to be resolved because the error is no longer showing.


voodoos (approved):
Thanks ! Just need to update the changelog section.

PR #2023, opened 2026-01-09
Signature help should not loop over the parameters once it is finished

Lucccyo:
When a space is entered after a full function call, signature-help loops back and highlights the first parameter again.
Signature-help should not loop over the parameters once all of them are written.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 unreleased
 ==========
+
   + merlin library
+    - Add "Other" variant to locate-types result (#2025)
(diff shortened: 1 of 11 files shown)


CHANGES.md
```diff
@@ -1,3 +1,8 @@
+unreleased
+==========
+  + test suite
+    - Signature help should not loop over the parameters once it is finished (#2023)
```
voodoos: This is not just a new test but also and foremost a change to signature's help behavior in merlin-lib.

PR #2050, opened 2026-03-24
Add LRU to improve the memory usage of the indexer

Lucccyo:
The index links are stored in a Hashtable, which grows without bound as we add new entries. This is using a lot of memory. 
This PR replaces the use of a Hashtable with an LRU cache to bound memory usage, keeping only the most recently used indices and evicting the least recently used entry when the capacity is reached.

### DIFF
--- a/src/index-format/dbllist.ml
+++ b/src/index-format/dbllist.ml
@@ -1,5 +1,5 @@
 type 'a cell =
-  { mutable content : 'a;
+  { content : 'a;
     weight : int;
     mutable prev : 'a cell;
     mutable next : 'a cell
@@ -17,9 +17,9 @@ type stats = {
 type 'a dbll =
   | Nil of int
   | List of
-      { mutable first : 'a cell;
-        mutable last : 'a cell;
-        mutable size : int;
+      { first : 'a cell;
+        last : 'a cell;
+        size : int;
         cap : int;
       }
 
--- a/src/index-format/dbllist.ml
+++ b/src/index-format/dbllist.ml
@@ -69,6 +69,7 @@ let discard t =
   | List l ->
     if l.first == l.last then (
       t.dbll <- Nil l.cap;
+      t.stats.discarded_size <- t.stats.discarded_size + l.last.weight;
       l.last.content)
     else
       let discarded_value = l.last.content in
--- a/src/index-format/dbllist.ml
+++ b/src/index-format/dbllist.ml
@@ -29,7 +29,7 @@ exception Action_on_empty_list of string
 
 let pp_stats t =
   let size = match t.dbll with | Nil _ -> 0 | List l -> l.size in
-  Printf.printf "total_cap \t\t: %d\nsize \t\t: %d\npromote_count \t: %d\nadd_count \t\t: %d\ndiscard_count \t: %d\nadd_size \t\t: %d\ndiscard_size \t: %d\nvolume_conservation \t: %d = %d + %d : %b\n%!"
+  Printf.eprintf "total_cap \t\t: %d\nsize \t\t: %d\npromote_count \t: %d\nadd_count \t\t: %d\ndiscard_count \t: %d\nadd_size \t\t: %d\ndiscard_size \t: %d\nvolume_conservation \t: %d = %d + %d : %b\n%!"
   t.stats.total_cap
   size
   t.stats.promote_count
--- a/src/index-format/dbllist.ml
+++ b/src/index-format/dbllist.ml
@@ -1,0 +1,126 @@
+type 'a cell =
+  { mutable content : 'a;
+    weight : int;
+    mutable prev : 'a cell;
+    mutable next : 'a cell
+  }
+
+type stats = {
+  mutable total_cap : int;
+  mutable promote_count : int;
+  mutable add_count : int;
+  mutable discard_count : int;
+  mutable add_size : int;
+  mutable discarded_size : int;
+}
+
+type 'a dbll =
+  | Nil of int
+  | List of
+      { mutable first : 'a cell;
+        mutable last : 'a cell;
+        mutable size : int;
+        cap : int;
+      }
+
+type 'a t = { mutable dbll : 'a dbll; stats : stats }
+
+exception Action_on_empty_list of string
+
+let pp_stats t =
+  let size = match t.dbll with | Nil _ -> 0 | List l -> l.size in
+  Printf.printf "total_cap \t\t: %d\nsize \t\t: %d\npromote_count \t: %d\nadd_count \t\t: %d\ndiscard_count \t: %d\nadd_size \t\t: %d\ndiscard_size \t: %d\nvolume_conservation \t: %d = %d + %d : %b\n%!"
+  t.stats.total_cap
+  size
+  t.stats.promote_count
+  t.stats.add_count
+  t.stats.discard_count
+  t.stats.add_size
+  t.stats.discarded_size
+  t.stats.add_size t.stats.discarded_size size (t.stats.add_size = t.stats.discarded_size + size)
+
+let create cap =
+  let stats = { total_cap = cap; promote_count = 0; add_count = 0; discard_count = 0; add_size = 0; discarded_size = 0 } in
+  { dbll = Nil cap; stats }
+
+let add_front t (v, w) =
+  t.stats.add_count <- t.stats.add_count + 1;
+  t.stats.add_size <- t.stats.add_size + w;
+  match t.dbll with
+  | Nil cap ->
+    let rec c = { content = v; weight = w; prev = c; next = c } in
+    t.dbll <- List { first = c; last = c; size = w; cap };
+    c
+  | List l ->
+    let rec new_first =
+      { content = v; weight = w; prev = new_first; next = l.first }
+    in
+    l.first.prev <- new_first;
+    t.dbll <- List { first = new_first; last = l.last; size = l.size + w; cap = l.cap };
+    new_first
+
+let discard t =
+  t.stats.discard_count <- t.stats.discard_count + 1;
+  match t.dbll with
+  | Nil _ ->
+    raise
+      (Action_on_empty_list
+         "Unable to discard the last element, the doubly linked list is empty.")
+  | List l ->
+    if l.first == l.last then (
+      t.dbll <- Nil l.cap;
+      l.last.content)
+    else
+      let discarded_value = l.last.content in
+      let discarded_weight = l.last.weight in
+      t.stats.discarded_size <- t.stats.discarded_size + discarded_weight;
+      let new_last = l.last.prev in
+      new_last.next <- new_last;
+      t.dbll <-
+        List
+          { first = l.first;
+            last = new_last;
+            size = Int.max 0 (l.size - discarded_weight);
+            cap = l.cap;
+          };
+      discarded_value
+
+let discard_size t s =
+  let rec iter acc t =
+    match t.dbll with
+    | Nil _ -> acc
+    | List l -> if l.size + s <= l.cap then acc else (
+      iter (discard t :: acc) t)
+  in
+  iter [] t
+
+let promote t c =
+  t.stats.promote_count <- t.stats.promote_count + 1;
+  match t.dbll with
+  | Nil _ ->
+    raise
+      (Action_on_empty_list
+         "Unable to promote a cell, the doubly linked list is empty.")
+  | List l ->
+    if l.first == c then ()
+    else if l.last == c then (
+      let new_last = l.last.prev in
+      new_last.next <- new_last;
+      let new_first = c in
+      new_first.next <- l.first;
+      new_first.prev <- new_first;
+      l.first.prev <- new_first;
+      t.dbll <-
+        List { first = new_first; last = new_last; size = l.size; cap = l.cap })
+    else
+      let voisin_prev = c.prev in
+      let voisin_next = c.next in
+      voisin_prev.next <- voisin_next;
+      voisin_next.prev <- voisin_prev;
+      let new_first = c in
+      new_first.prev <- new_first;
+      new_first.next <- l.first;
+      l.first.prev <- new_first;
+      t.dbll <- List { first = new_first; last = l.last; size = l.size; cap = l.cap }
+
+let get c = c.content
--- a/src/index-format/dbllist.mli
+++ b/src/index-format/dbllist.mli
@@ -1,5 +1,5 @@
 type 'a cell =
-  { mutable content : 'a;
+  { content : 'a;
     weight : int;
     mutable prev : 'a cell;
     mutable next : 'a cell
@@ -17,9 +17,9 @@ type stats = {
 type 'a dbll =
   | Nil of int
   | List of
-      { mutable first : 'a cell;
-        mutable last : 'a cell;
-        mutable size : int;
+      { first : 'a cell;
+        last : 'a cell;
+        size : int;
         cap : int;
       }
 
--- a/src/index-format/dbllist.mli
+++ b/src/index-format/dbllist.mli
@@ -1,0 +1,35 @@
+type 'a cell =
+  { mutable content : 'a;
+    weight : int;
+    mutable prev : 'a cell;
+    mutable next : 'a cell
+  }
+
+type stats = {
+  mutable total_cap : int;
+  mutable promote_count : int;
+  mutable add_count : int;
+  mutable discard_count : int;
+  mutable add_size : int;
+  mutable discarded_size : int;
+}
+
+type 'a dbll =
+  | Nil of int
+  | List of
+      { mutable first : 'a cell;
+        mutable last : 'a cell;
+        mutable size : int;
+        cap : int;
+      }
+
+type 'a t = { mutable dbll : 'a dbll; stats : stats }
+
+exception Action_on_empty_list of string
+
+val pp_stats : 'a t -> unit
+val create : int -> 'a t
+val add_front : 'a t -> 'a * int -> 'a cell
+val discard_size : 'a t -> int -> 'a list
+val promote : 'a t -> 'a cell -> unit
+val get : 'a cell -> 'a
--- a/src/index-format/granular_map.ml
+++ b/src/index-format/granular_map.ml
@@ -193,7 +193,7 @@ module Make (Ord : Map.OrderedType) = struct
       iter f r
 
   let rec iter_in_memory f s =
-    if Granular_marshal.is_on_disk s then
+    if not (Granular_marshal.is_on_disk s) then
       match fetch s with
       | Empty -> ()
       | Node { l; v; d; r; _ } ->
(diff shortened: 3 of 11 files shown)


src/index-format/dbllist.ml
```diff
@@ -0,0 +1,74 @@
+type 'a cell = {
+  mutable content : 'a;
+  mutable prev : 'a cell;
+  mutable next : 'a cell;
+}
+
+type 'a tt = Nil | List of { mutable first : 'a cell; mutable last : 'a cell }
```
xvw: Maybe just cosmetic but why not using `::` and `[]` as constructor names ?

xvw: And I am not found of the name `tt` :)

Lucccyo: For the `::` and `[]`, yes, why not. This isn't clearer to me, but if it is for the general public, let's go. :)

Lucccyo: The name `tt` should definitely be changed :+1:

art-w: nitpick: I'm not a big fan of `::` since `List` isn't the standard list "Cons" operator :P (it points to both ends of the linked list)

src/index-format/lru.ml
```diff
@@ -0,0 +1,52 @@
+module Make (K : sig
```
voodoos: I think we should cite the cachechache inspiration and licensing at the beginning of this module's files. (Not sure if we should do the same for the dllist one which diverged more significantly)

src/index-format/lru.ml
```diff
@@ -0,0 +1,52 @@
...
+
+  type 'a t = {
+    htbl : (K.t * 'a) Dbllist.cell H.t;
+    dbllst : (K.t * 'a) Dbllist.t;
+    cap: int;
+    mutable size : int;
+  }
+
+  let unsafe_create cap = {
+    htbl = H.create 0;
```
voodoos: That's not a great initial value for the hashtbl. Cache cache uses the `cap` value here, which seems like a reasonable choice.

src/index-format/lru.ml
```diff
@@ -0,0 +1,52 @@
...
+end) =
+struct
+  module H = Hashtbl.Make (K)
+  type key = K.t
+
+  type 'a t = {
+    htbl : (K.t * 'a) Dbllist.cell H.t;
+    dbllst : (K.t * 'a) Dbllist.t;
+    cap: int;
+    mutable size : int;
```
voodoos: Do we need this ? It seems to be the same as the Hashtbl's lenght, and you never use it's value anyway.

src/index-format/dbllist.ml
```diff
@@ -0,0 +1,74 @@
...
+      let voisin_prev = c.prev in
+      let voisin_next = c.next in
+      voisin_prev.next <- voisin_next;
+      voisin_next.prev <- voisin_prev;
+      let new_first = c in
+      new_first.prev <- new_first;
+      new_first.next <- l.first;
+      l.first.prev <- new_first;
+      t := List {first = new_first; last = l.last}
+    )
```
voodoos: Is this more complex than the cachecache's version because you don't have the `first` and `last` pointers ?

src/index-format/dbllist.ml
```diff
@@ -0,0 +1,74 @@
...
+      prev = new_first;
+      next = l.first
+    } in
+    l.first.prev <- new_first;
+    t := List {first = new_first; last = l.last};
+    new_first
+
+let discard t =
+  match !t with
+  | Nil -> failwith "unable to discard in an empty dllist."
```
voodoos: Does the Dbllist's API enforces that this case cannot happen ?
If not you should either raise an appropriate exception or return a `_ result`

src/index-format/dbllist.ml
xvw: make sense!

src/index-format/lru.ml
Lucccyo: This is true.

src/index-format/lru.ml
Lucccyo: You are right :+1:

src/index-format/lru.ml
Lucccyo: Yes, but if we test it with a large capacity, it will allocate a very large amount of memory for the hashtable.
As the hashtbl resizes itself, I think it is better to initialize it to a low value. But I agree that `0` is not the best choice, maybe we can seed it to `16`. What do you think @xvw ?

src/index-format/dbllist.ml
Lucccyo: Maybe I didn't get your point, but I have the two pointers `last` and `first`?

src/index-format/dbllist.ml
Lucccyo: yes !

src/index-format/lru.ml
xvw: Maybe @art-w has a better idea about that :)

src/index-format/dbllist.ml
voodoos: Oh right ! But then why is the cc version much "simpler":

```ocaml

let promote l i =
  if i <> l.last then (
    l.t.prev.(l.t.next.(i)) <- l.t.prev.(i);
    if i = l.first then l.first <- l.t.next.(i)
    else l.t.next.(l.t.prev.(i)) <- l.t.next.(i);
    l.t.next.(l.last) <- i;
    l.t.prev.(i) <- l.last;
    l.t.next.(i) <- -1;
    l.last <- i);
  l.last
  ```
  
  Oh I see, it's because the more complex `next` and `prev` values.

Comment (voodoos) [reply src/index-format/dbllist.ml]: So nothing to change here, do you know why that approach was chosen i cache cache ? Is it more performant ?

File: src/index-format/lru.ml (hunk 3 above)
Comment (art-w) [reply src/index-format/lru.ml]: Since this PR is still a draft (it doesn't yet work), the 0 size is fine! (anecdotally the stdlib Hashtbl already has a min size of 16)
In fact the cachecache `cap` size was blocking @Lucccyo from measuring the memory usage of the LRU to confirm that it works, since the memory usage difference included the array/hashtbl sizes of the cachecache LRU (which made it hard to interpret the benchmarks). We can revisit the initial hashtbl size once we have confirmation that the LRU is releasing memory :)

File: src/index-format/lru.ml (hunk 4 above)
Comment (art-w) [reply src/index-format/lru.ml]: We discussed with @Lucccyo that the LRU should not just limit the number of elements stored in the cache, but also take into account their memory footprint (some elements are very light, others are heavier) which we can measure from their on-disk representation. In that context it makes sense to have such a field to store the total "memory size used" :)

File: src/index-format/dbllist.ml (hunk 5 above)
Comment (art-w) [reply src/index-format/dbllist.ml]: @Lucccyo 's  refactoring is related to the confusion when measuring the impact of the LRU, since the memory usage difference measured was actually showing the LRU hashtbl / (`prev`,`next`) arrays size instead of demonstrating that the LRU was really releasing memory. The original `cachecache` arrays should be more efficient in practice, since all the memory is pre-allocated :)

File: src/index-format/lru.ml (hunk 3 above)
Comment (Lucccyo) [reply src/index-format/lru.ml]: Ok :)

File: src/index-format/granular_marshal.ml [hunk 7]
```diff
@@ -136,18 +177,19 @@ let write ?(flags = []) fd root_schema root_value =
...
             (match !original_lnk with
-            | Serialized_reused _ -> ()
+            | Serialized_reused _ | On_disk_ptr _ -> ()
             | In_memory_reused v -> write_child_reused original_lnk schema v
             | _ -> failwith "Granular_marshal.write: duplicate not reused");
             lnk := !original_lnk
-          | In_memory v -> write_child lnk schema v size ~placeholders ~restore
-          | On_disk _ ->
-            write_child lnk schema (fetch lnk) size ~placeholders ~restore)
+          | In_memory v | In_memory_c (v, _) ->
```
art-w: I'm curious if your tests ran into the `In_memory_c` case here? The current code is correct, but if this does happen in practice then `In_memory_c` would benefit from being translated into a pointed-index `On_disk_ptr` directly (since an `In_memory_c` is an in-memory value that was read from an existing file)

src/index-format/granular_marshal.ml
```diff
@@ -82,28 +110,41 @@ let open_store store =
...
-    let v = fetch_loc store loc schema in
-    lnk := In_memory v;
+    let v, size = fetch_loc store loc schema in
+    let discarded = Dbllist.discard_size lru_dbllist size in
+    let cell =
+      Dbllist.add_front lru_dbllist (Cached (lnk, loc, store, schema), size)
+    in
+    List.iter
+      (fun (Cached (link, loc, store, schema)) ->
+        Cache.remove store.cache loc;
```
art-w: This remove doesn't look entirely correct, since we are loosing sharing as we never reintroduce `lnk` into the cache later if it turns out that offset was still useful (so we'll allocate new links for that same value). Does your benchmarks show that this remove is necessary? :)

src/index-format/granular_marshal.ml
Lucccyo: My test doesn't hit the `In_memory_c` case here.

src/index-format/granular_marshal.ml
Lucccyo: Yes I agree.
On the benchmarks, removing the `Cache.remove` multiplies the max memory usage by around 1.1.

src/index-format/granular_marshal.ml
```diff
@@ -1,63 +1,78 @@
 module Cache = Hashtbl.Make (Int)
 
-type store = { filename : string; cache : any_link Cache.t }
+type store = { filename : string; id : int; }
```
art-w: It looks wrong to remove this `cache`, it's pretty essential :)

src/index-format/granular_marshal.ml
```diff
@@ -80,74 +112,111 @@ let open_store store =
...
+          | Serialized { loc } ->
+            (match Cache.find_opt store.cache loc with
+            | Some (Link (type b) ((existing_lnk, type_id') : b link * _)) ->
+              (match Type.Id.provably_equal type_id type_id' with
+              | Some (Equal : (a link, b link) Type.eq) ->
+                lnk := !existing_lnk
+              | None -> lnk := On_disk { store; loc; schema })
+            | None ->
+              Cache.add store.cache loc (Link (lnk, type_id));
+              lnk := On_disk { store; loc; schema })
```
Lucccyo: @art-w What do you think? The cache keeps a trace (loc to link) of the serialized links.

src/index-format/granular_marshal.ml
```diff
@@ -80,74 +112,111 @@ let open_store store =
...
+            schema iter v;
+            child_smalls:= (Value (Obj.magic v)) :: !child_smalls;
+            lnk := Small_from_papa { papa = papa_link; pos = !child_pos };
+            child_pos := !child_pos + 1
+          | Serialized { loc } ->
+            (match Cache.find_opt store.cache loc with
+            | Some (Link (type b) ((existing_lnk, type_id') : b link * _)) ->
+              (match Type.Id.provably_equal type_id type_id' with
+              | Some (Equal : (a link, b link) Type.eq) ->
+                lnk := !existing_lnk
```
art-w: I misspoke the other day when I assumed you could remove all the `Duplicate`, this is a case where we are potentially leaking if we do not use them (the `existing_lnk` is tracked and released by the LRU, but the `lnk` copy isn't)

src/index-format/granular_marshal.ml
```diff
@@ -80,74 +112,111 @@ let open_store store =
...
+              | None -> lnk := On_disk { store; loc; schema })
```
art-w: This shouldn't happen (the original code raised a fatal error)

src/index-format/granular_marshal.ml
```diff
@@ -80,74 +112,111 @@ let open_store store =
...
+  let size_read = pos_in fd - loc in
+  let child_pos = ref 0 in
+  let child_smalls = ref [] in
+  let rec iter =
+    { yield =
+        (fun (type a) (lnk : a link) type_id schema ->
+          match !lnk with
+          | Small v ->
+            schema iter v;
+            child_smalls:= (Value (Obj.magic v)) :: !child_smalls;
```
art-w: Can you run `dune fmt`, and also get rid of the `magic` now? :)

src/index-format/granular_map.ml
```diff
@@ -191,6 +192,15 @@ module Make (Ord : Map.OrderedType) = struct
...
 
+  let rec iter_in_memory f s =
+    if not (Granular_marshal.is_on_disk s) then
+      match fetch s with
+      | Empty -> ()
+      | Node { l; v; d; r; _ } ->
+        iter_in_memory f l;
+        f v d;
+        iter_in_memory f r
+
```
art-w: This function looks like dead code now

src/index-format/dbllist.ml
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type 'a dbll =
+  | Nil of int
+  | List of
+      { mutable first : 'a cell;
+        mutable last : 'a cell;
+        mutable size : int;
+        cap : int;
+      }
```
art-w, line 24: nitpick: it makes sense that the fields are mutable, but the code below doesn't make use of it?

src/index-format/granular_marshal.ml
```diff
@@ -1,63 +1,89 @@
...
+
+and any_value = Value : 'a -> any_value
+
+and cached = Cached : 'a link * int * store * 'a schema -> cached
+
 and 'a link = 'a repr ref
 
 and 'a repr =
   | Small of 'a
+  | Small_from_papa of { papa : papa_link; pos : int }
```
art-w: To rename ^^'

src/index-format/granular_marshal.ml
art-w: yes it looks similar to the previous code :) (which we shouldn't need to update, so to keep the diff short it's best not to introduce minor differences)

src/index-format/granular_marshal.ml
```diff
@@ -80,74 +112,111 @@ let open_store store =
...
+  | Serialized _ | Small _ | On_disk_ptr _ ->
     invalid_arg "Granular_marshal.fetch: serialized"
   | Placeholder -> invalid_arg "Granular_marshal.fetch: during a write"
-  | Duplicate original_lnk ->
-    let v = fetch original_lnk in
-    lnk := In_memory v;
-    v
+  | Small_from_papa { papa; pos } ->
+    let PLink papa = papa in
+    (fun (type a) (papa : a link) ->
```
art-w: Is this necessary? (you may have to add an explicit type annotation to `fetch` since the recursion is irregular, `let rec fetch : type a. a link -> a = fun lnk -> ...`)

src/index-format/dbllist.mli
```diff
@@ -0,0 +1,35 @@
...
+type 'a t = { mutable dbll : 'a dbll; stats : stats }
+
+exception Action_on_empty_list of string
+
+val pp_stats : 'a t -> unit
+val create : int -> 'a t
+val add_front : 'a t -> 'a * int -> 'a cell
+val discard_size : 'a t -> int -> 'a list
+val promote : 'a t -> 'a cell -> unit
+val get : 'a cell -> 'a
```
art-w, line 35: Is `get` really used?

art-w (changes requested):
Almost there! I have some minor requests to clean up the changes :)

src/index-format/granular_marshal.ml
Lucccyo: So I kept the `Duplicate`, but during my tests, they were never used.

src/index-format/dbllist.mli
Lucccyo: Yes, it is, to return the value contained in the cell when we are fetching an `In_cache` value.

Lucccyo:
@art-w I don't know how `Duplicate should behave in the `write`.
What we could want is to call ' write ' recursively on the original link, right?

Lucccyo:
I pushed the review's resolution in a separate commit to make it easier to see the difference between before and after. They are going to be squashed and reorganized at the end.

src/index-format/granular_marshal.ml
```diff
@@ -1,153 +1,243 @@
...
+              lnk := On_disk_ptr { filename; id; loc }
+            | Serialized _ | On_disk_ptr _  ->
+              lnk := !original_lnk
+            | In_memory v ->
+              write_child original_lnk schema v size ~placeholders ~restore;
+              lnk := !original_lnk
+            | In_cache (_v, t, _children) ->
+              let Cached (_, loc, {filename; id; _}, _) = t.content in
+              lnk := On_disk_ptr { filename; id; loc }
+            | _ -> failwith "Granular_marshal.write: duplicate unexpected state")
```
art-w: For `Duplicate`, does the original code with `write_child_reused` work? The only new case should be `In_cache`, which is equivalent to an `On_disk`. (I believe it shouldn't be possible to have `Duplicate (Small ... | In_memory ...)`, but let me know if that's not the case?)

Lucccyo: I suggest going backward and reintroducing the `Duplicate` and `*_reused`, since they are linked, and we cannot remove `*_reused` while keeping the `Duplicate`.

src/index-format/granular_marshal.ml
```diff
@@ -1,153 +1,243 @@
...
-    lnk := In_memory v;
-    v
+  | Duplicate original_lnk -> fetch original_lnk
+  | Small_child { parent; pos } ->
+    let PLink parent = parent in
+      ignore(fetch (Obj.magic parent));
+      (match !parent with
+      | In_cache (_, _, small_poses) ->
+        let Value v = small_poses.(pos) in
+        Obj.magic v
```
Lucccyo, line 185: @art-w How can we return properly `v` without the `Obj.magic` since it is existentially bound, and the typer rejects the case where we annotated `v` with type `a` like:
```
let (Value (v : a)) = small_poses.(pos) in
```
with the error on the part `v : a`:
```
File "src/index-format/granular_marshal.ml", line 192, characters 19-26:
192 |         let (Value (v : a)) = small_poses.(pos) in
                         ^^^^^^^
Error: This pattern matches values of type a
       but a pattern was expected which matches values of type $a1
       Hint: $a1 is an existential type bound by the constructor Value.
```

art-w: This is the part where we use a `Type.Id.t` to convince the type-checker that the value has indeed the right type :)

src/index-format/dbllist.mli
art-w: Really?

voodoos:
Closing in favor of #2079

voodoos:
Now merged as part of #2079

Issue #2019, opened 2025-12-23
`ocamlmerlin single type-enclosing` stopped producing Reason syntax


anmonteiro:
The following cram test produces different results in OCaml versions <= 5.2 vs OCaml >= 5.3.

The merlin versions are [v5.3-502](https://github.com/ocaml/merlin/releases/tag/v5.3-502) and [v5.4-503](https://github.com/ocaml/merlin/releases/tag/v5.4-503).

```shell
  $ cat >dune-project <<EOF
  > (lang dune 3.20)
  > EOF
  $ cat >dune <<EOF
  > (executable (name foo))
  > EOF
  $ cat > foo.re <<EOF
  > let _x = List.map;
  > EOF

  $ dune build
  $ ocamlmerlin single type-enclosing -position 1:15 -verbosity 0 -filename foo.re < foo.re | jq '.value[0]'
  {
    "start": {
      "line": 1,
      "col": 9
    },
    "end": {
      "line": 1,
      "col": 17
    },
    "type": "('a => 'b, list('a)) => list('b)",
    "tail": "no"
  }

```

With OCaml 5.3+, the diff is no longer in Reason syntax:

```diff
@@ -19,7 +19,7 @@
       "line": 1,
       "col": 17
     },
-    "type": "('a => 'b, list('a)) => list('b)",
+    "type": "('a -> 'b) -> 'a list -> 'b list",
     "tail": "no"
   }
```

voodoos:
Thanks for your report @anmonteiro ! I am not able to reproduce your issue with Merlin `main` and OCaml 5.4. I find it suspicious that it would only happen with OCaml 5.3, are you certain your configuration is correct ?

anmonteiro:
I can't reproduce this anymore 🤔 , not sure what was wrong with my setup then.

Thank you for taking a look!

PR #2051, opened 2026-03-24
Add support for pointing indexes in ocaml-index

Tim-ats-d:
This PR adds support for pointing to an index file that already exists on disk.

This mechanism results in significantly smaller index files and fewer write operations.


src/index-format/granular_marshal.ml
```diff
@@ -53,6 +59,10 @@ let read_loc store fd loc schema =
               lnk := On_disk { store; loc; schema };
               Cache.add store.cache loc (Link (lnk, type_id)))
           | In_memory _ | In_memory_reused _ | On_disk _ | Duplicate _ -> ()
+          | On_disk_ptr { filename; loc } ->
+            lnk :=
+              On_disk
+                { store = { filename; cache = Cache.create 0 }; loc; schema }
```
art-w: `Cache.create 0` is problematic here because we can have multiple "pointing indexes" pointing at the same file, but they'll end up with different caches (instead of a shared one). @voodoos suggested looking into `File_cache` to ensure the shared caches get properly cleaned up on file changes (it's used by `Index_cache` atm for a similar reason, and in fact it's possible we don't need `Index_cache` anymore once the granular store Cache is using it)

art-w (changes requested):
@voodoos raised another issue: Since the indexes are not self-contained anymore, it's possible that a pointing index targets another index file that has been removed or updated. If we don't check for that and follow pointers to arbitrary locations, some very bad things could happen!

src/index-format/granular_marshal.ml
Tim-ats-d: I see the problem, my last commit should fix it. I also try to remove `Index_cache` module, but it seems to be used `src/kernel/mocaml.ml`.

src/index-format/granular_marshal.ml
```diff
@@ -157,7 +165,6 @@ let write ?(flags = []) fd root_schema root_value =
             lnk := !original_lnk
           | In_memory v -> write_child lnk schema v size ~placeholders ~restore
           | On_disk { store; loc; _ } ->
-            incr count;
```
voodoos: This change seems unrelated to the rest of this commit.

src/index-format/granular_marshal.ml
voodoos: What about this cache ? (it's really a question, nothing more)

src/index-format/granular_marshal.ml
Tim-ats-d: Yeah, I remove it in my last commit.

src/index-format/granular_marshal.ml
Tim-ats-d: I guess that we can use our new global cache here too :stuck_out_tongue:

src/index-format/granular_marshal.ml
```diff
@@ -24,11 +27,39 @@ let schema_no_sublinks : _ schema = fun _ _ -> ()
 
 let link v = ref (In_memory v)
 
+let is_on_disk lnk =
+  match !lnk with
+  | On_disk _ | On_disk_ptr _ -> false
+  | _ -> true
```
art-w: The name `is_on_disk` suggests the opposite of what this is doing ^^'

src/index-format/granular_marshal.ml
```diff
@@ -24,11 +27,39 @@ let schema_no_sublinks : _ schema = fun _ _ -> ()
...
+let int_of_binstring s =
+  Array.fold_right
+    (fun v acc -> (acc lsl 8) + v)
+    (Array.init ptr_size (fun i -> Char.code s.[i]))
+    0
+
+let fetch_id filename =
+  In_channel.with_open_bin filename (fun fd ->
+      seek_in fd (String.length Config.index_magic_number);
+      int_of_binstring (really_input_string fd ptr_size))
```
art-w: It looks wrong to use `with_open` here, instead of checking an already open file descriptor: we have no guarantee that the file won't change in between the next time we open it to actually read from it. We can assume that the file contents will only be updated by creating a fresh new file (which guarantees a previously opened fd still see the old contents), so if we have checked a file descriptor there's no risk of a race :)

src/index-format/granular_marshal.ml
```diff
@@ -53,6 +84,14 @@ let read_loc store fd loc schema =
               lnk := On_disk { store; loc; schema };
               Cache.add store.cache loc (Link (lnk, type_id)))
           | In_memory _ | In_memory_reused _ | On_disk _ | Duplicate _ -> ()
+          | On_disk_ptr { filename; loc; id } ->
+            let pointed_index_id = fetch_id filename in
+            if pointed_index_id = id then
```
art-w: This is too eager: We would want to delay this check in `fetch` when we have no choice but to read from the file (otherwise we are checking the `id` too early and the file may not have the same id when we re-open it to read it during fetch)

src/index-format/granular_marshal.ml
```diff
@@ -118,25 +157,16 @@ let cache (type a) (module Key : Hashtbl.HashedType with type t = a) =
...
-  String.init ptr_size (fun i -> Char.chr ((v lsr i lsl 3) land 255))
-
-let int_of_binstring s =
-  Array.fold_right
-    (fun v acc -> (acc lsl 8) + v)
-    (Array.init ptr_size (fun i -> Char.code s.[i]))
-    0
-
 let write ?(flags = []) fd root_schema root_value =
+  let id = Random.int ((Float.pow 2. 30. |> Float.to_int) - 1) in
```
art-w: nitpick: `1 lsl 30 - 1` instead of floats, but also it would be nice to have more random bits since we are using 8 bytes to store it anyway :)

src/index-format/index_format.ml
```diff
@@ -133,6 +134,7 @@ let write ~file index =
   Misc.output_to_file_via_temporary ~mode:[ Open_binary ] file
     (fun _temp_file_name oc ->
       output_string oc magic_number;
+      Random.self_init ();
```
art-w: nitpick: it's uncommon to call `self_init` as part of a procedure (instead of at the start of the program execution), but `write` is expected to be called only once during the indexer execution so this is fine. Nevertheless, we could instead use `Random.State.make_self_init` here to produce the unique random identifier we need (we don't need more than one call to random so we don't benefit from resetting the random global state)

art-w (changes requested):
It's looking good! I only have some minor requests to improve the `id` checking logic :)

Tim-ats-d:
Here are some benchmark results from Hyperfine I got running `ocaml-index` on a bunch of ocaml-index files generated from Merlin repo.

Without pointing index:
```
Time (mean ± σ):      2.425 s ±  0.087 s    [User: 1.968 s, System: 0.395 s]
Range (min … max):    2.313 s …  2.553 s    10 runs
```

With pointing-index enabled:
```
Time (mean ± σ):     896.2 ms ±  42.5 ms    [User: 640.6 ms, System: 245.2 ms]
Range (min … max):   840.6 ms … 972.3 ms    10 runs
```

Complete command I used:
```
ocaml-index aggregate dot_merlin_reader.ocaml-index merlin_analysis.ocaml-index merlin_commands.ocaml-index merlin_config.ocaml-index merlin_dot_protocol.ocaml-index merlin_extend.ocaml-index merlin_index_format.ocaml-index merlin_kernel.ocaml-index merlin_specific.ocaml-index merlin_utils.ocaml-index ocaml-index.ocaml-index ocaml_compression.ocaml-index ocaml_parsing.ocaml-index ocaml_preprocess.ocaml-index ocaml_typing.ocaml-index ocaml_utils.ocaml-index ocamlmerlin_server.ocaml-index os_ipc.ocaml-index query_commands.ocaml-index query_protocol.ocaml-index  --root . --rewrite-root
```

src/index-format/granular_marshal.ml
Tim-ats-d: I use an int64 right now https://github.com/ocaml/merlin/pull/2051/commits/8411fc72228db248a22e2174eadd4475dffead98

src/index-format/granular_marshal.ml
Tim-ats-d: Fixed https://github.com/ocaml/merlin/pull/2051/commits/d8c00da4c5d240821bf01d77cfce330ad234308b

src/index-format/index_format.ml
Tim-ats-d: Fixed https://github.com/ocaml/merlin/pull/2051/commits/38b82cde9f9699de9c2b89e3750a9ccdb9e81dab

src/index-format/granular_marshal.ml
```diff
@@ -39,7 +91,8 @@ let read_loc store fd loc schema =
           | Small v ->
             schema iter v;
             lnk := In_memory v
-          | Serialized { loc } -> lnk := On_disk { store; loc; schema }
+          | Serialized { loc } ->
+            lnk := On_disk { store; loc; schema; id = fetch_id store }
```
art-w: Here we expect the `id` to be the same as the `store` passed as argument to `read_loc`, so no need for `fetch_id`

src/index-format/granular_marshal.ml
```diff
@@ -50,36 +103,18 @@ let read_loc store fd loc schema =
                 invalid_arg
                   "Granular_marshal.read_loc: reuse of a different type")
             | exception Not_found ->
-              lnk := On_disk { store; loc; schema };
+              lnk := On_disk { store; loc; schema; id = fetch_id store };
```
art-w: Same, no need for `fetch_id` since it would be an error if the file had a different `id` than the `store`

src/index-format/granular_marshal.ml
```diff
@@ -88,17 +123,21 @@ let fetch_loc store loc schema =
...
-    lnk := In_memory v;
-    v
+  | On_disk { store; loc; schema; id } ->
+    let pointed_index_id = fetch_id store in
+    if pointed_index_id = id then (
+      let v = fetch_loc store loc schema in
+      lnk := In_memory v;
+      v)
+    else
+      failwith "Granular_marshal.read_loc: pointing to an outdated index file"
```
art-w: Should we define an explicit exception for outdated stores, such that we can explicitly catch it to signal to users that they should rebuild? (in which case the exceptions when we try to open a store file should similarly trigger that exception, e.g. when the file is missing or lacking the read permission etc)

src/index-format/granular_marshal.ml
```diff
@@ -118,25 +157,20 @@ let cache (type a) (module Key : Hashtbl.HashedType with type t = a) =
...
-    (fun v acc -> (acc lsl 8) + v)
-    (Array.init ptr_size (fun i -> Char.code s.[i]))
-    0
-
-let write ?(flags = []) fd root_schema root_value =
+let write ?(flags = []) fd root_schema root_value rand_state =
+  let id =
+    let buf = Bytes.create 8 in
+    Bytes.set_int64_be buf 0 (Random.State.int64 rand_state Int64.max_int);
+    Bytes.to_string buf
```
art-w: Is this `binstring_of_int` but for int64? (a 63 bits id should be enough, especially because later we use `int_of_binstring` which may not have the right behavior for an int64)

src/index-format/granular_marshal.ml
```diff
@@ -146,8 +180,9 @@ let write ?(flags = []) fd root_schema root_value =
             | _ -> failwith "Granular_marshal.write: duplicate not reused");
             lnk := !original_lnk
           | In_memory v -> write_child lnk schema v size ~placeholders ~restore
-          | On_disk _ ->
-            write_child lnk schema (fetch lnk) size ~placeholders ~restore)
+          | On_disk { store; loc; _ } ->
+            let id = fetch_id store in
```
art-w: It looks wrong to call `fetch_id` here, both because we already know the expected `id` and because the fetched one could be different (but there's no need to validate ids at this point)

src/index-format/granular_marshal.mli
```diff
@@ -53,11 +56,9 @@ and iter = { yield : 'a. 'a link -> 'a link Type.Id.t -> 'a schema -> unit }
 (** A schema usable when the ['a] value does not contain any links. *)
 val schema_no_sublinks : 'a schema
 
-(** [write oc schema value] writes the [value] in the output channel [oc],
-    creating unmarshalling boundaries on every link in [value] specified
-    by the [schema]. *)
+(** [write oc schema value rand_state] writes the [value] in the output channel [oc], creating unmarshalling boundaries on every link in [value] specified by the [schema]. [rand_state] is used to generate random index ID. *)
 val write :
-  ?flags:Marshal.extern_flags list -> out_channel -> 'a schema -> 'a -> unit
+  ?flags:Marshal.extern_flags list -> out_channel -> 'a schema -> 'a -> Random.State.t -> unit
```
art-w: nitpick: Since we are only using the `Random.State.t` once to generate one id, I feel it would be clearer to pass the id directly to `write` as argument (otherwise someone reading the code would have to check if we have other usages for the random state)

src/index-format/granular_marshal.ml
```diff
@@ -24,11 +27,60 @@ let schema_no_sublinks : _ schema = fun _ _ -> ()
...
+  | Some (store', fd) when store == store' -> fd
+  | Some (_, fd) ->
+    close_in fd;
+    force_open_store store
+  | None -> force_open_store store
+
+let fetch_id store =
+  let fd = open_store store in
+  seek_in fd (String.length Config.index_magic_number);
+  int_of_binstring (really_input_string fd ptr_size)
```
art-w: I don't like this function :P ... because I feel it's orienting the code towards the wrong pattern for checking store ids. The only moment where we need to read and check the id is when we `force_open_store` (at this point we should have the expected `id` to compare with, to validate that the file descriptor is usable). The one exception is the `read` function when we open the root index and don't have an expected id (so we have to use the one read from the file).

art-w (changes requested):
Thanks! I left some additional comments because I'm still not convinced that the store `id` validation is correct... In fact it might be easier to add the `id` field in the `store` type directly (instead of `On_disk`), since it's as relevant as the `filename` to identify the expected index file.
I also did not see the code to report a message to users that their build is out of date? (when we run into an invalid store file)

src/index-format/granular_marshal.ml
Tim-ats-d: I see what you mean, I've removed it and move the check in `force_open_store`.

src/index-format/granular_marshal.ml
Tim-ats-d: Yeah, I've forgotten the fact that an integer is an int63 on 64-bit platform. I now use `binstring_of_int` on a value of type `int`.

src/index-format/granular_marshal.ml
Tim-ats-d: Great idea!

Tim-ats-d: Added here https://github.com/ocaml/merlin/pull/2051/commits/e56a469693204787d0b994901b7283ee83ff5359

Tim-ats-d: and catched here https://github.com/ocaml/merlin/pull/2051/commits/dbde744c88cbc14460ef9bd6b2127f7f95ff2991 with a short error message.

src/ocaml-index/bin/ocaml_index.ml
```diff
@@ -74,43 +74,54 @@ let set_log_level debug verbose =
...
+            (Option.value ~default:"none" root_directory))
+        !input_files
+    | _ -> Printf.printf "Nothing to do.\n%!");
+    exit 0
+  with Granular_marshal.Outdated_store reason ->
+    let msg =
+      match reason with
+      | `Missing_file filename ->
+        Format.asprintf "Missing file \"%s\"." filename
+      | `Index_ids_doesn't_match -> "Index IDs doesn't match."
```
art-w: nitpick: maybe we also want to show the filename for IDs mismatch? (it can help understanding the cause of the error, after a partial rebuild)

src/index-format/granular_marshal.ml
```diff
@@ -20,15 +23,68 @@ and 'a schema = iter -> 'a -> unit
...
+    let required_id = int_of_binstring (really_input_string fd ptr_size) in
+    if required_id = store.id then (
+      last_open_store := Some (store, fd);
+      fd)
+    else raise (Outdated_store `Index_ids_doesn't_match)
+  with Sys_error _ -> raise (Outdated_store (`Missing_file store.filename))
+
+let open_store store =
+  match !last_open_store with
+  | Some (store', fd) when store == store' -> fd
```
art-w: On line 113, we are generating "new" store values so `==` might fail more often than necessary? (even though the two stores are technically equal)

src/index-format/granular_marshal.ml
```diff
@@ -11,6 +13,7 @@ and 'a repr =
   | Serialized of { loc : int }
   | Serialized_reused of { loc : int }
   | On_disk of { store : store; loc : int; schema : 'a schema }
+  | On_disk_ptr of { filename : string; loc : int; id : int }
```
art-w, line 16: (cc @voodoos) Do we know if the filenames are absolute or relative? (I wonder if there's a risk that `ocaml-index` could be called from different folders, such that the pointed filenames are later not found because of the relative paths interpretation)

art-w (approved):
Thanks a lot, this looks great! I have some minor remarks but nothing blocking, in case @voodoos has time for another look :)

src/index-format/granular_marshal.ml
Tim-ats-d: There are relative. We can maybe use Unix.realpath on such filename.

src/index-format/granular_marshal.ml
Tim-ats-d: Is it fine to use structural comparison here? Or it will decrease performance

src/index-format/granular_marshal.ml
```diff
@@ -20,15 +23,68 @@ and 'a schema = iter -> 'a -> unit
 
 and iter = { yield : 'a. 'a link -> 'a link Type.Id.t -> 'a schema -> unit }
 
+exception
+  Outdated_store of [ `Missing_file of string | `Index_ids_doesn't_match ]
```
voodoos: `_don't` or better

```suggestion
  Outdated_store of [ `Missing_file of string | `Index_ids_do_not_match ]
```

src/index-format/granular_marshal.mli
```diff
@@ -53,11 +56,18 @@ and iter = { yield : 'a. 'a link -> 'a link Type.Id.t -> 'a schema -> unit }
 (** A schema usable when the ['a] value does not contain any links. *)
 val schema_no_sublinks : 'a schema
 
-(** [write oc schema value] writes the [value] in the output channel [oc],
-    creating unmarshalling boundaries on every link in [value] specified
-    by the [schema]. *)
+(** Exception raised when attempting to consult an outdated store. *)
+exception
+  Outdated_store of [ `Missing_file of string | `Index_ids_doesn't_match ]
```
voodoos:
```suggestion
  Outdated_store of [ `Missing_file of string | `Index_ids_do_not_match ]
```

src/index-format/granular_marshal.ml
voodoos: The filenames in the locations stored in the indexes are relative unless `--root` is passed to the indexer. 
In `dune` they should always be relative to the root of the build context (like `_build/default`).

src/index-format/granular_marshal.ml
art-w: It's a good question that has a right answer.

Tim-ats-d: I think I just have to compare them by their respective ID :smile:

art-w: Just to be sure, it wouldn't hurt to check the filenames too :)

src/index-format/granular_marshal.ml
art-w: Thanks! Can we test it in `union.t` after the aggregation?

src/index-format/granular_marshal.ml
Tim-ats-d: Done!

src/ocaml-index/bin/ocaml_index.ml
```diff
@@ -74,43 +74,54 @@ let set_log_level debug verbose =
...
+  with Granular_marshal.Outdated_store { filename; reason } ->
+    let msg =
+      match reason with
+      | `Missing_file -> Format.asprintf "Missing file \"%s\"." filename
+      | `Index_ids_do_not_match ->
+        Format.asprintf "Index IDs doesn't match for \"%s\"." filename
+    in
+    Printf.printf
+      "%s\nHint: try to rebuild indexes with dune build @ocaml-index.\n%!" msg;
+    exit 1
```
art-w, line 127: Can we add a test to check the "outdated store" logic? :) (e.g. in `union.t`)

Tim-ats-d: Of course, here are the two tests I've added: https://github.com/ocaml/merlin/pull/2051/commits/9a78a18bc05cf08b03949dcc0ea3a3f030a85ec2

src/index-format/granular_marshal.mli
```diff
@@ -53,11 +56,19 @@ and iter = { yield : 'a. 'a link -> 'a link Type.Id.t -> 'a schema -> unit }
...
+exception
+  Outdated_store of
+    { filename : string; reason : [ `Missing_file | `Index_ids_do_not_match ] }
+
+(** [write oc id schema value] writes the [value] in the output channel [oc], creating unmarshalling boundaries on every link in [value] specified by the [schema]. [id] is used as index UID. *)
 val write :
-  ?flags:Marshal.extern_flags list -> out_channel -> 'a schema -> 'a -> unit
+  ?flags:Marshal.extern_flags list ->
+  out_channel ->
+  int ->
```
voodoos: It would be clearer to use a more descriptive labelled parameter here.

src/ocaml-index/bin/ocaml_index.ml
```diff
@@ -74,43 +74,54 @@ let set_log_level debug verbose =
...
+            (Uid_map.fold
+               (fun _uid locs acc -> acc + Lid_set.cardinal locs)
+               defs 0)
+            (Uid_map.cardinal approximated)
+            (Hashtbl.length cu_shape)
+            (Option.value ~default:"none" root_directory))
+        !input_files
+    | _ -> Printf.printf "Nothing to do.\n%!");
+    exit 0
+  with Granular_marshal.Outdated_store { filename; reason } ->
```
voodoos, line 118: Shouldn't we also guard the index reading in analysis/occurrences.ml against these exceptions ?

tests/test-dirs/occurrences/project-wide/union.t
```diff
@@ -52,3 +52,20 @@ All files should be listed on queries: (except `both.ml`)
...
+An error is expected, a pointer references an index file, but it doesn't exist anymore:
+
+  $ mv test_sig.ocaml-index index-files
+  $ ocaml-index dump project.ocaml-index
+  Missing file "test_sig.ocaml-index".
+  Hint: try to rebuild indexes with dune build @ocaml-index.
+  [1]
+  $ mv index-files test_sig.ocaml-index
+
+An error is expected, a pointer references an index file that is considered outdated because we rebuild it:
```
voodoos:
```suggestion
An error is expected, a pointer references an index file that is considered outdated because we rebuilt it:
```

voodoos (approved):
Thanks, looks fine to me. Just one worry about the new exception handling when reading the indexes for occurrences. Especially since it's a point in time were indexes are much more likely to be out of sync than during aggregation at build time.

Also, it's all or nothing, but for user queries it would be great to still be able to provide partial results if some of the files are out-of-sync. Would that be hard to do ? If so let's just add a comment about it around the place where we read indexes for occurrences.

voodoos:
And this definitely deserves a changelog entry.

src/index-format/granular_marshal.mli
Tim-ats-d: For which parameter? [flags]?

voodoos: (h)int

src/ocaml-index/bin/ocaml_index.ml
Tim-ats-d: Fixed in https://github.com/ocaml/merlin/pull/2051/commits/c3a3bee3b186ffbfd0ba0a57f7348913c323b359

voodoos:
Thanks, can you rebase ? there are some merge conflicts.

CHANGES.md
```diff
@@ -20,6 +20,8 @@ Thu Apr 09 09:59:38 WAT 2026
     - Add reproduction case for #1763 but it is not failing anymore (#2021)
     - Add a test to reproduce [Locate command fails on multi-line type definitions](#1987) (#2020)
     - Add a regression test for issue #2019 (#2030)
+    - Add a regression test for issue #2019 (#2030)
+    - Implemente a pointer to another index file: a new optimization of how ocaml-index builds index files. The resulting index files should take up less disk space (#2051)
```
voodoos, line 24: It's not a test, you duplicated the `Add a regression test for issue #2019 (#2030)` entry and it should be in a new changelog section, not an already released one. I am going to make these changes before merging.

voodoos:
Thanks @Tim-ats-d ! I believe the current CI failures are due to external issues. (well, the fuzzy CI is broken again, we should definitely do something about it.)

PR #2059, opened 2026-04-15
Synchronize `misc.ml` with the compiler

liam923:
The compiler's `utils/misc.ml(i)` module contains a large number of miscellaneous functions. Merlin has a similar file, `src/utils/misc.ml(i)`, that shares a lot of code with the compiler's version, but it makes some small changes and also adds a few functions. Merlin's `misc.ml(i)` and the compiler's `misc.ml(i)` should be kept in sync with each other, just as other shared files should. However, they historically have not been kept in sync. There's a large amount of diff between the two files that is the result of changes to the compiler version not being applied to the Merlin version.

This PR brings these files back in sync by:
1. Repairing the diff so that all diff is intentional.
2. Moving `src/utils/misc.ml` to `src/ocaml/utils/misc.ml` so that the file is kept in sync in the future.

I suggest reviewing this PR by-commit:
- 889d185046e2eefc4f3269edf8c5b46a860be9a8 moves `src/utils/misc.ml(i)` to `src/ocaml/utils/misc.ml(i)`, and it modifies some build rules and module paths to reflect this.
- e899a9f41e9f213f0f770c7b761336f1d3a18bf7 moves `src/utils/format_doc.ml(i)` to `src/ocaml/utils/format_doc.ml(i)`. `Format_doc` is a module that comes from the compiler, and as far as I can tell, is was only in `src/utils` rather than `src/ocaml/utils` to avoid a dependency cycle, as `misc.ml` depends on it.
- 56c71d015651472a7c9727caffcae5df4f34887b replaces the content of Merlin's `misc.ml(i)` with the content of the compiler's `misc.ml(i)` (at compiler revision 82bcb707388fa8df0126d43b1bfd16780ba5df9f). You shouldn't read the diff of this commit - you should just compare the file against the expected.
- aef567d13c18af6f2c19d803988be99a5ed8c581 is the meat of this PR. It is the result of manually comparing Merlin's misc against the compilers and applying all diff that is meaningful.
- ce1c397cc62096331225a76b635647b27046d3bb rewrites a bunch of module paths in the compiler that changed due to the previous commits. This commit purely reduces the diff with the compiler - it simply reverts changes that were made due to the differences between Merlin's misc and the compiler's.


src/ocaml/utils/misc.ml
```diff
@@ -1415,3 +1616,40 @@ module Magic_number = struct
            | Error err -> Error (Unexpected_error err)
            | Ok () -> Ok info
 end
+
+(* Merlin-specific: The below functions are specific to Merlin. *)
+
```
voodoos: Do these functions benefit from being here ? If they are not used by other modules in `src/ocaml` we should probably move them elsewhere.

It's less obvious for the previous changes but maybe we could also extract them out, thus reducing the diff further, but that can be done later.

liam923: No, good point. I moved them to `src/utils/std.ml`. In fact, `modules_in_path` was a duplicate function that existed in both `std.ml` and `misc.ml`

voodoos (approved):
Thanks Liam, looks good to me !

A few remaining things:
1. This change is worth a changelog entry since it changes merlin's library API.
2. And indeed it breaks ocaml-lsp, we should repair it.
3. There is a formatting issue now that you moved some functions out of the vendored compiler sources.

Let me know if you would prefer me to handle some of these points.

Also,that's unrelated, but the Fuzzy CI generating step is broken again with 5.4: 
https://github.com/ocaml/merlin/actions/runs/24471065872/job/71511114904
cc @pitag-ha @Lucccyo do we have plans to make it work again soon, or should I disable it in the meantime ?

liam923:
Thanks for the review. I've done (1) and (3). Would you be able to do (2)? I don't have a good workflow for these sorts of changes, so it would be quite painful for me to do.

voodoos:
I cannot push on your branch, so let's merge and see if the compatibility is good 🙂

PR #2060, opened 2026-04-17
Saved parts testing

trefis:
Some investigations into merlin's saved parts motivated by the review of https://github.com/ocaml/ocaml/pull/14241 .

A very short summary of a couple of long discussions:
- merlin's handling of saved types *in the absence of type errors* is different from the typechecker and seemed potentially incorrect (?)
- turns out if you try to mimick the type checker's behavior, you get unwanted nesting/duplication of the saved-parts, leading to a browse tree blowup (cf. 95ba06e) ; thanks to @let-def for pointing me in this direction, I'd have missed it

Some key points:
- you have to look at the browse tree, because that's where merlin "unfolds" the saved parts, looking at the typedtree straight out of the typechecker wouldn't have shown the issue
- this PR is for illustration purposes ... though if you want to be a bit defensive against future compiler updates, I suggest you reproduce and adapt the latest test (which shows when things blow up) for the other parts of the codebase which currently do not "save parts" on success.
- I suspect the blow up wouldn't produce an observable difference in behavior with regular merlin commands (because most of them deduplicate nodes they receive from enclosing, the nodes they return, etc based on location) ... but it probably maybe could have a negative impact on performances? (notice the uncertainty)

### DIFF
--- a/src/ocaml/typing/typecore.ml
+++ b/src/ocaml/typing/typecore.ml
@@ -3874,8 +3874,7 @@ let rec type_exp ?recarg env sexp =
 
 and type_expect ?recarg env sexp ty_expected_explained =
   Msupport.with_saved_types
-    ~save_part:(fun e -> Partial_expression e)
-    ~warning_attribute:sexp.pexp_attributes
+    ~warning_attribute:sexp.pexp_attributes ?save_part:None
       (fun () ->
         let saved = save_levels () in
         try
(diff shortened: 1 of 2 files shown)


trefis:
Oh, one last thing: typemod tends to pass a `~save_part` argument to `with_saved_types` ... does that lead to a blow up? Would there be an observable difference if it didn't?

I have neither checked nor thought hard about it. Have fun.

xvw (approved):
> this PR is for illustration purposes ... though if you want to be a bit defensive against future compiler updates, I suggest you reproduce and adapt the latest test (which shows when things blow up) for the other parts of the codebase which currently do not "save parts" on success.

We probably want to have a ticket to track this.

voodoos:
There is a weird indentation issue in the test on the macos runner....

PR #2062, opened 2026-04-28
Invalidate typer cache when new cmi file exists

liam923:
This PR fixes a caching bug in `mtyper.ml`. At the moment, the cache will not get invalidated when a new cmi file is created in a directory in the build path (as would happen when dune builds a new module). This PR makes the mtyper cache now get invalidated whenever a new file is created in the build path.

The PR is split into two commits. The first commit adds a test that demonstrates the bug, and the second fixes it.


voodoos (dismissed):
Edit: I posted in the wrong PR 🙈 !

voodoos:
At first glance this looks like a very important fix, the test is.. 🫣

Thanks a lot Liam, I will give it a closer look soon.

It might also be the root cause of the [ocaml-lsp "desync" issues](https://github.com/ocaml/ocaml-lsp/issues/1209).

liam923:
Thanks. Kudos to @JimmyZJX for identifying the bug, making a reproducible test, and coming up with the fix. I just wrapped it up into a PR. :pray:

voodoos:
And now for the funny part: I was not very satisfied with the fix: it felt like a workaround for a deeper issue. Namely with `Env.check_state_consistency`. With the reproduction I was able to investigate a bit more and it turns out that the real culprit is the test condition in:

https://github.com/ocaml/merlin/blob/b3ee1852972560f8ba8dec2430a9211e22ffc7c4/src/ocaml/typing/persistent_env.ml#L262-L266

This means missing modules are not always registered as such, and thus ignored when checking for "env consistency".

And I found this comment thread: https://github.com/ocaml/ocaml/pull/12246/changes#r1352530501
Where @gasche did forebode that exact issue two years ago 🔮

Reading that thread, I suspect that the correct fix would be to do what he suggested:
> (I would be slightly more at ease with caching both with an extra argument to Missing, which we are sure will have no adverse side-effect.)

It should probably be upstreamed afterwards, even if it does not impact the compiler.

Do you want to have a go at it ? If not I will do it later this week.

liam923:
I don't have a very good grasp on how `persistent_env.ml` works, so I'd prefer leaving it for you to take a stab at (but I'm happy to review). But yes I agree, it certainly would be better to improve `Env.check_state_consistency` if possible.

voodoos:
I just pushed an alternative fix. I still need to fix the tests, the `touch -d '+2 seconds' lib` command is not accepted by macos's bsd version of `touch`.

voodoos:
Thanks everyone ! I will merge now and make a release for OCaml 5.4, hopefully the last one. 
This issue was there since OCaml 5.2 support, but I suggest we only release 5.2 and 5.3 fixes if there is a demand for it.

Issue #2063, opened 2026-05-01
Adding project-wide type-aware search for ocaml code patterns (ocamlgrep)


mjambon:
Dear project maintainers,

LexiFi created a tool called [ocamlgrep](https://github.com/LexiFi/ocamlgrep) (formerly known as grep_cmt) and has been using it internally for a while. For example, it allows a query such as `(__ : int list)` to locate all expressions of type `int list` in a project. We would like to make this tool more widely available and easy to use by the OCaml community. I'm looking into the possibility of integrating it into merlin and eventually in IDEs. I would try to do most of the work and would make it easy to maintain.

Ocamlgrep's development would benefit from being part of the merlin code base in the following ways:
- locating the project's cmt files reliably (ocamlgrep currently assumes a dune build using the default context)
- managing compiler-libs variations across OCaml versions (ocamlgrep currently works with a single OCaml version)

One approach would be to move [ocamlgrep's matching code](https://github.com/LexiFi/ocamlgrep/blob/main/lib/Match.mli) to the merlin project. It would be part of merlin-lib and a new merlin command would function like ocamlgrep (takes a user-specified pattern and returns matches in various files). A standalone `ocamlgrep` command could still be provided but it would be a thin wrapper around the merlin command.

Please let me know if this looks like a good idea and if you see any challenges.

xvw:
Yes, when it was first presented, it seemed really interesting. Is it already possible to perform some kind of search with Sherlodoc? By using (for ocaml-eglot) the various commands: https://github.com/tarides/ocaml-eglot#search-for-values

However, `ocamlgrep` offers undeniable advantages (for searching beyond public APIs). Would @nojb agree to integrating `ocamlgrep` directly into Merlin?

nojb:
> Would [@nojb](https://github.com/nojb) agree to integrating `ocamlgrep` directly into Merlin?

Yes! To be clear: this proposal has full support from LexiFi. Indeed, @mjambon is collaborating with LexiFi to drive this forward. And we would all be more than happy to explore any venues or ideas proposed by the Merlin team on this subject.

mjambon:
Here's a pull request that integrates ocamlgrep into merlin: https://github.com/ocaml/merlin/pull/2064
It's big but most of the code was simply copied from the original repo.

Please let me know what I can do to facilitate the code review.

voodoos:
Ah, I had a message I forgot to send yesterday 🤦‍♂️

[...]

PR #2064, opened 2026-05-06
Add ocamlgrep: structural search of OCaml code

mjambon:
This pull request is a tentative integration of the [ocamlgrep project](https://github.com/LexiFi/ocamlgrep) into merlin as [brought up last week](https://github.com/ocaml/merlin/issues/2063).

I used Claude Code to make an initial basic integration and then I guided it to fit better in merlin. Consider that the existing code works but needs to be refined or redone until we're happy with it. It's not considered final in any way. Please treat it as a collection of suggestions given my current understanding of the project which is improving every day.

The main pieces of this PR are:

- an `src/project` library that uses `dune describe workspace` to list the cmt files that are available for scanning. This is a different approach than what was used in the original ocamlgrep project. It is reusable for other project-wide scans.
- the `Expr_search` module (`src/analysis/expr_search.mli`) that takes an OCaml expression (with an optional type annotation and placeholders such as `__` or `__42`) and finds matching expressions of the expected type in a typed tree (from a cmt file). In the original ocamlgrep project, this was the `Match` module.
- the `ocamlgrep` subcommand, which ignores standard input and scans all the available cmt files using `Expr_search`.
- cram tests to check that ocamlgrep finds what it's supposed to.

Originally, the PR included a standalone `ocamlgrep` command that was essentially the same as invoking `ocamlmerlin single ocamlgrep -query ... < /dev/null` but shorter to invoke and defaulting to human-friendly text output. This can be done provided as an external package or added to the merlin repo later.


src/commands/query_json.ml
```diff
@@ -587,3 +595,17 @@ let json_of_response (type a) (query : a t) (response : a) : json =
...
   | Version, version -> `String version
+  | Ocamlgrep _, { findings; warnings } ->
+    let json_of_finding (f : Query_protocol.ocamlgrep_finding) =
+      `Assoc
+        [ ("file", `String f.file);
+          ("line", `Int f.line);
+          ("col_start", `Int f.col_start);
+          ("col_end", `Int f.col_end);
+          ("context", `String f.context)
+        ]
```
mjambon: This needs to change to as to represent multiline matches correctly.

src/frontend/query_protocol.ml
```diff
@@ -136,6 +136,17 @@ type occurrence = { loc : Location.t; is_stale : bool }
 type substitution_result =
   { loc : Location.t; content : string; selection_range : Location.t }
 
+type ocamlgrep_finding =
+  { file : string;
+    line : int;
+    col_start : int;
+    col_end : int;
+    context : string
+  }
```
mjambon, line 148: This needs to change to include the end line. Comments should specify whether the indices are 0-based or 1-based (follow merlin conventions if applicable).

gr-im:
I’m not a maintainer (or even a contributor) to Merlin, but I must admit I was delighted to read the proposal to integrate `ocamlgrep` into Merlin. However, I’m quite surprised because I have a strong feeling that things were done… in the wrong order.

In my opinion, ocamlgrep should expose an API under `merlin-lib` (taking advantage of the project concept, which is duplicated by both Merlin and ocamlgrep) by standardizing paths and locations, and should be invoked: 
- as a binary (`ocamlgrep`)
- and as an analyze phase. 

I’m quite surprised by the dependency lattice that was chosen.

mjambon:
Thanks for the feedback. I was originally hoping that the only module ocamlgrep would bring in would be the `Match` module and that everything related to scanning files would reuse (or add) shared merlin functionality. Let me dig deeper and iterate on this.

src/project/dune_workspace.mli
```diff
@@ -0,0 +1,76 @@
+(* This file is part of the merlin-lib.project library
+   See the attached LICENSE file. *)
+
+(** Structured access to the output of [dune describe workspace]. *)
+
+(** A module described by dune. Each of [impl], [intf], [cmt], [cmti]
+    is encoded by dune as a 0-or-1-element list -- e.g. [(intf ())] or
+    [(intf (lib/foo.mli))]. We mirror that shape with [string list];
+    use {!impl_path}, {!cmt_path} etc. for the convenient
+    [string option] view. *)
```
mjambon: This should use the `option` type directly. The sexp deriver will expect a list of 0 or 1 element, I think.

src/project/dune_workspace.ml
```diff
@@ -0,0 +1,156 @@
...
+let describe ?root () =
+  (* dune parses [--root] positionally: it must come *after* the
+     [describe workspace] subcommand, not before [describe]. *)
+  let base_args = [ "describe"; "workspace"; "--format=csexp"; "--lang"; "0.1" ] in
+  let args =
+    match root with None -> base_args | Some dir -> base_args @ [ "--root"; dir ]
+  in
+  let argv = Array.of_list ("dune" :: args) in
+  let env = Unix.environment () in
+  let ic, oc, ec = Unix.open_process_args_full "dune" argv env in
```
mjambon: Here, make sure to catch the exception we get if `dune` isn't available and print a nice error message (`Unix.Unix_error(Unix.ENOENT, _, _)`).

mjambon:
I made changes but I'm still working on it. I'll let you know when I think it's in a good state.

src/frontend/query_protocol.ml
```diff
@@ -136,6 +136,18 @@ type occurrence = { loc : Location.t; is_stale : bool }
...
   { loc : Location.t; content : string; selection_range : Location.t }
 
+type ocamlgrep_finding =
+  { file : string;
+    start_line : int;
+    start_col : int;
+    end_line : int;
+    end_col : int;
+    lines : string list
+  }
```
mjambon, line 148: This should use the `Location.t` type like everything else.

src/analysis/match.mli
```diff
@@ -0,0 +1,174 @@
...
+   7 |   | None -> None
+   8 |   | Some y -> Some y
+   v}
+
+   For machine-readable JSON output (suitable for editor integration)
+   use the merlin subcommand:
+
+   {v
+   ocamlmerlin single ocamlgrep -query <pattern> < /dev/null
+   v}
```
mjambon: This section used to apply to the `ocamlgrep` standalone command but now we're within merlin and all we have is JSON output. An `ocamlgrep` command should be available in the future but it might come from a package outside merlin, we don't know yet. For now, it's best to not assume and only mention the merlin subcommand.

mjambon:
Status update:

1. I removed the package that was providing an `ocamlgrep` executable. We can add it later or keep it outside of merlin; it would simply depend on `merlin-lib`.
2. The main matching module where a pattern (expression AST) is scanned against a typedtree is now known as `src/analysis/expr_search.mli`.
3. I didn't find general-purpose full-project scanning functionality so I adapted what we had in ocamlgrep and moved it to a new `src/project` folder. It offers iteration over cmt files which is used by the `ocamlgrep` subcommand.
4. Identifying the cmt files for a Dune project is now done by invoking `dune describe workspace`. It produces well-defined output that is now available to access all the files known to a Dune build. It was previously done without invoking Dune by making assumptions about where it stores build products.

Please keep in mind that none of the above was really discussed with the project maintainers and can be removed, changed, etc. I could also split the PR into multiple PRs if it's easier to review. I'm looking forward to meeting with the project maintainers to get guidance on what's desirable and what isn't.

I'm also going to rewrite the PR description by hand which will be more concise but will better reflect my understanding of the issues.

src/analysis/expr_search.ml
```diff
@@ -0,0 +1,462 @@
...
+  match l, p with
+  | Lident "__", _ ->
+      true
+  | Ldot (l0, {txt=s2; _}), Path.Pdot (p0, s1) when s1 = s2 || s2 = "__" ->
+      path_matches_lident l0.txt p0
+  | Lident s2, Path.Pdot (_, s1) when s1 = s2 ->
+      true  (* the longident can be a suffix of the path *)
+  | Lident s, Path.Pident id ->
+      Ident.name id = s
+  | _ ->
```
mjambon: Instead of a catch-all, we should enumerate all the constructors so as to trigger a warning when a future version of compiler-libs adds a new constructor:

```
| _ -> false
```
we would write something like
```
| (Lident _ | Ldot _), (Pident _ | Pdot _) -> false
```

merlin-lib.opam
```diff
@@ -14,6 +14,8 @@ depends: [
   "ocaml" {>="5.4" & <"5.5"}
   "dune" {>= "3.0.0"}
   "csexp" {>= "1.5.1"}
+  "sexplib0"
+  "ppx_sexp_conv"
```
voodoos: We are very conservative when it comes to adding new dependencies to merlin, and depending on `base` is definitely not acceptable.

src/project/dune
```diff
@@ -0,0 +1,27 @@
...
+ (preprocess (pps ppx_sexp_conv)))
+
+; Generate standard_library.ml exposing the OCaml standard library path.
+; The same trick is used in src/kernel/dune. Merlin's vendored Config
+; module strips the upstream Config.standard_library field, so we recover
+; it at build time from %{ocaml-config:standard_library}.
+(rule
+ (targets standard_library.ml)
+ (action (write-file %{targets}
+          "let path = {|%{ocaml-config:standard_library}|}")))
```
voodoos, line 26: Didn't check but this is suspicious, can't we get the information from Merlin_kernel ?

src/project/paths.mli
```diff
@@ -0,0 +1,85 @@
+(* This file is part of merlin.
+   See the attached LICENSE file.
+   Copyright (C) 2026 LexiFi
+
+   Originally written by Nicolás Ojeda Bär (LexiFi);
+   maintained by Martin Jambon (LexiFi). *)
+
```
voodoos: As discussed in the meeting, let's drop the headers for now.

src/project/paths.ml
```diff
@@ -0,0 +1,142 @@
...
+               (fun name -> Filename.check_suffix name ".cmi")
+               (Sys.readdir entry)
+          then
+            res := entry :: !res;
+          walk entry
+        end
+      ) (Sys.readdir dir)
+  in
+  walk paths.build_source_root;
+  List.rev !res
```
voodoos, line 72: Why do you need this in addition to the output of dune describe ? I thought the goal was to avoid walking the tree ourselves.

src/project/paths.ml
```diff
@@ -0,0 +1,142 @@
...
+
+(* Project location is delegated to [Mconfig_dot.find_project_context]
+   so the rule for "what counts as a project root" stays consistent
+   with the rest of merlin: walk up looking for a [.merlin],
+   [dune-project], or [dune-workspace] file.
+
+   The [Mconfig_dot.context] is abstract in the .mli, so we read what
+   we need from the second return value, which is the path of the
+   config file: its directory is the project root, its basename tells
+   us whether the configurator is dune or .merlin.
```
voodoos, line 82: Since Merlin already does it as parts of the pipeline, maybe we could simply add this information to the pipeline's config and reuse it.

src/project/paths.ml
```diff
@@ -0,0 +1,142 @@
...
+  match Mconfig_dot.find_project_context abs_search_root with
+  | None ->
+    Error
+      (Printf.sprintf
+         "Could not find a project rooted at or above %s. \
+          Looks for a dune-project, dune-workspace, or .merlin file."
+         abs_search_root)
+  | Some (_ctx, config_file) ->
+    let project_root = Filename.dirname config_file in
+    match Filename.basename config_file with
```
voodoos, line 104: Could we match on the "configurator" in the context instead ?

src/project/paths.mli
```diff
@@ -0,0 +1,85 @@
...
+        If unspecified, the current directory (.) is assumed.
+        If specified, this path must be a prefix of every file in search
+        results. For example, specifying the search root as foo/../bar should
+        return results such as foo/../bar/baz.ml, not an absolute path,
+        not a normalized path (bar/baz.ml).
+
+        It's also the starting point for locating
+        which Dune project we're in and for locating the compiled data
+        we need to run a pattern search.
+    *)
```
voodoos, line 32: I don't understand why this is needed. The feature is project wide, so can't we just always consider that the search root is the root of the project ?

src/analysis/expr_search.mli
```diff
@@ -0,0 +1,167 @@
...
+    {!Merlin_project.Scan.incremental_search} when partially applied
+    to [query]:
+    {[
+      let search = Match.search expr in
+      Scan.incremental_search paths cmt_files handler search
+    ]} *)
+val search :
+  Parsetree.expression ->
+  Cmt_format.cmt_infos -> source:string -> src_lines:string array ->
+  finding list
```
voodoos, line 160: Is this dead code ?

src/project/scan.mli
```diff
@@ -0,0 +1,84 @@
...
+   maintained by Martin Jambon (LexiFi). *)
+
+(**
+   Build-system-agnostic framework for scanning a project's cmt files.
+
+   This module handles the mechanics of reading cmt files, verifying
+   source digests, mapping compiler-internal paths back to
+   project-relative display paths, and emitting structured events.
+   The actual search logic and the finding type are supplied by the
+   caller.
```
voodoos, line 8: I am not going deeper right now. As we discussed in the meeting, you should reuse the Cmt_cache when appropriate and might be able to reuse source-matching functionalities from the Locate module.

src/frontend/query_commands.ml
```diff
@@ -921,3 +921,38 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a = function
...
+        Merlin_project.Dune_workspace.describe
+          ~root:paths.project_root
+          ()
+      with
+      | Ok ws -> Merlin_project.Dune_workspace.local_cmt_files ws
+      | Error msg -> failwith msg
+    in
+    let expr = Expr_search.parse_query query in
+    Merlin_project.Scan.incremental_search paths cmt_files handle_event
+      (Expr_search.search expr);
```
voodoos: Could incremental_search be a fold instead of forcing the user to accumulate results in references ?

src/frontend/query_commands.ml
```diff
@@ -921,3 +921,38 @@ let dispatch pipeline (type a) : a Query_protocol.t -> a = function
...
+    { Query_protocol.findings = List.rev !findings;
+      warnings = List.rev !warnings
+    }
```
voodoos: There are many way the query can fail, and some of these errors are directly imputable to the user, like query parsing errors. We should not rely on failwith and the generic catch all of merlin for these, but report them as part of the response to the query.

src/commands/new_commands.ml
```diff
@@ -501,6 +501,34 @@ let all_commands =
...
+         where each finding carries the file, a start/end position pair \
+         (using merlin's standard `{line, col}` shape), and the matched \
+         source lines. Ignores the buffer on stdin (this command is \
+         project-wide, not buffer-local). The pattern language is the one \
+         from LexiFi's ocamlgrep (https://github.com/LexiFi/ocamlgrep) -- \
+         see its README for the full syntax reference."
+      ~default:("", None) begin fun buffer (query, root) ->
+      if query = "" then failwith "-query <pattern> is mandatory";
+      run buffer (Query_protocol.Ocamlgrep (query, root))
+      end;
```
voodoos, line 532: We should also update the PROTOCOL.md file

src/analysis/expr_search.mli
mjambon: No. The doc comment needs updating (`Match.search` &rarr; `Expr_search.search`) but otherwise it is current.

src/frontend/query_commands.ml
mjambon: Sure. It will expose a more complicated interface to a user who just wants to do printfs.

src/frontend/query_commands.ml
mjambon: This is what we get in VSCode currently for a malformed search query:

Image

How should user errors be signaled to merlin? I think a simple exception called `User_error of string` would do, while all other exceptions would be treated as internal errors, similar to HTTP's 4xx vs. 5xx error families.

The exception `User_error "Could not parse search expression."` would be shown as just `Could not parse search expression.` instead of `Failure ("Could not parse search expression.")`.

merlin-lib.opam
mjambon: This is for `src/project/dune_workspace.ml`.

The dependency on `ppx_sexp_conv` is the culprit. The runtime lib we really need here is `csexp` which is already a dependency of merlin.

To remove the dependency on `ppx_sexp_conv`, I tried copy-pasting the ml code generated by ppx_sexp_conv but it was ugly so I ended up with a simpler handrolled implementation (which incidentally is what Claude Code had done in the very first version of this PR before I told it to use ppx_sexp_conv instead). I think the result is reasonable now.

src/project/paths.ml
mjambon: This is inherited from the original implementation. I need to figure out the cmi situation and load paths.

src/project/paths.mli
mjambon: This was inherited from the `ocamlgrep` CLI where it's natural to search from the current directory rather than from the project root. I'll clarify or maybe just remove the option.

src/frontend/query_commands.ml
mjambon: What I did is add an `errors` field to the JSON response... but not to the OCaml type which is a `(ocamlgrep_result, string) result`. I'm not convinced that this is a great solution but it bypasses the questions of (1) representing Ok/Error in JSON and (2) a more generic error handling mechanism that would differentiate user errors from internal errors.

Here's what it looks like in VSCode now:
image

src/analysis/expr_search.mli
mjambon: Oh, I think you noticed that `Scan.search` is dead code which is true. This is in another module.

voodoos:
Hi @mjambon, what's the status on this, is it ready for review ? I see that some ci check are failing...

nojb:
> Hi @mjambon, what's the status on this, is it ready for review ? I see that some ci check are failing...

I take the liberty of responding; @mjambon may want to add more detail and also correct me if I'm wrong. Martin explained to me that it turned out that there was not much to be gained from merging this tool and Merlin itself, given that most of the infrastructure of Merlin could not be directly used for this tool given the different foci (single file/buffer vs global search). So Martin reoriented the task to making a standalone tool with, potentially, a direct integration into the various editor (eg VS Code) integrations. This is much simpler than trying to integrate the code into an existing project such as Merlin or OCaml-LSP (which was also considered but similarly it was concluded that there was not much to be gained in terms of synergy).

In short, I think the PR can be closed, but @mjambon will confirm that.

Thanks again @voodoos for the time you took to explain things. It was rather helpful to clarify the various technical points, even if in the end the project ended up going in a different direction.

mjambon:
@voodoos sorry for the communication gap.

To add to what @nojb said, our plan is now to release ocamlgrep as a standalone command that outputs JSON and is consumed directly by the IDE, without going through the LSP.

Thank you for your guidance on how merlin works.

Closing the PR.

voodoos:
> > Hi @mjambon, what's the status on this, is it ready for review ? I see that some ci check are failing...
> 
> I take the liberty of responding; @mjambon may want to add more detail and also correct me if I'm wrong. Martin explained to me that it turned out that there was not much to be gained from merging this tool and Merlin itself, given that most of the infrastructure of Merlin could not be directly used for this tool given the different foci (single file/buffer vs global search). So Martin reoriented the task to making a standalone tool with, potentially, a direct integration into the various editor (eg VS Code) integrations. This is much simpler than trying to integrate the code into an existing project such as Merlin or OCaml-LSP (which was also considered but similarly it was concluded that there was not much to be gained in terms of synergy).
> 
> In short, I think the PR can be closed, but @mjambon will confirm that.
> 
> Thanks again @voodoos for the time you took to explain things. It was rather helpful to clarify the various technical points, even if in the end the project ended up going in a different direction.

> To add to what @nojb said, our plan is now to release ocamlgrep as a standalone command that outputs JSON and is consumed directly by the IDE, without going through the LSP.

Oh I see. It's true that the benefits are not as large as we could expect today, but I do think having it in Merlin and have a custom LSP query for it would have the advantage of providing a standard way for editor devs to add support for it. And the cmt cache could turn out to be a nice gain in the end... Anyway, thanks for getting back to us, let's see how it goes and we will continue to think about it on our side too.  Let's keep in touch 🙂

Issue #1796, opened 2024-07-08
Jump to definition with let-punning


liam923:
Consider the program:
```ocaml
let (let+) x f = List.map f x

let f x =
  let+ x in
  x
```
Locating `x` on line 4 gives:
```
$ ocamlmerlin single locate -position 4:7 -filename test.ml < test.ml | jq .value
"Already at definition point"
```

In my opinion, for this command it would be better to treat `x` as the variable bound on line 3 rather than the one on line 4. I believe that if a user is querying the definition point of `x` on line 4, they likely want the definition that isn't on that line.

voodoos:
Yes I agree, there is a similar frustrating thing happening when asking for the definition of punned record arguments (but in that case it is less clear which is the correct answer).

It suspect that the interesting node is given a ghost location, so Merlin chooses the other one.

panglesd:
This can be closed! (by #2066)

PR #2066, opened 2026-05-11
Fix locate in let punnings

panglesd:
This fixes an issue with the locate query reported in #1796.

(Sorry this is quite verbose for such a small PR but it is also here to help me be corrected if I have assumed something wrong)

In a punned let binding, both the pattern and the expression have the same location. None of them is ghost. I think that somehow breaks the assumption that "sibling nodes have disjoint locations".

During a locate query, the first thing is to load the environment at the point of the cursor. The lookup of "nodes at cursor position" will suppose the invariant mentioned above to return a list all nodes the cursor is in (of which we take the first one, the "deepest").

Since two siblings (the pattern and expression) have the same location, the order in which they are considered matter: when the first one is found, we never look at the other siblings, and "enter" it to continue the search.

In the current implementation, patterns are looked first, and we end up with the environment at the pattern to do our "locate". And so, the unhelpful "Already at definition point".

This PR fixes this by folding let operators in a different order, making the expressions side be entered first.

Having the behaviour depend on the order of the fold is slightly brittle, but I think this PR is still fine to solve #1796 without refactoring a huge part to solve a bigger underlying issue ("siblings might have the same location"). But I'm new to the codebase so I might miss something!
(I'm happy to implement a better fix, but then I'd like to discuss with more experienced Merlin disciple first)

### DIFF
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -845,7 +845,10 @@ let infer_namespace ?namespaces ~pos lid browse is_label =
       `Error `Missing_labels_namespace)
   | None -> (
     match
-      (Context.inspect_browse_tree ~cursor:pos lid [ browse ], is_label)
+      ( Context.inspect_browse_tree
+          ~disambiguate:Mbrowse.Tie_breaker.prefer_expression ~cursor:pos lid
+          [ browse ],
+        is_label )
     with
     | None, _ ->
       log ~title:"from_string" "already at origin, doing nothing";
--- a/src/analysis/locate.ml
+++ b/src/analysis/locate.ml
@@ -844,12 +844,11 @@ let infer_namespace ?namespaces ~pos lid browse is_label =
         "input is clearly a label, but the given namespaces don't cover that";
       `Error `Missing_labels_namespace)
   | None -> (
-    match
-      ( Context.inspect_browse_tree
-          ~disambiguate:Mbrowse.Tie_breaker.prefer_expression ~cursor:pos lid
-          [ browse ],
-        is_label )
-    with
+    let context =
+      let disambiguate = Mbrowse.Tie_breaker.prefer_expression in
+      Context.inspect_browse_tree ~disambiguate ~cursor:pos lid [ browse ]
+    in
+    match (context, is_label) with
     | None, _ ->
       log ~title:"from_string" "already at origin, doing nothing";
       `Error `At_origin
--- a/src/kernel/mbrowse.ml
+++ b/src/kernel/mbrowse.ml
@@ -69,7 +69,6 @@ let node_loc node = approximate_loc Browse_raw.node_real_loc node
 let node_merlin_loc node = approximate_loc Browse_raw.node_merlin_loc node
 
 let leaf_node = List.hd
-let leaf_loc t = node_loc (snd (leaf_node t))
 
 let drop_leaf t =
   match t with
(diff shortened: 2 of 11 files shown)


xvw (approved):
Neat!

tests/test-dirs/punning.t
```diff
@@ -0,0 +1,53 @@
...
+
+  $ cat > let_punning.ml <<EOF
+  > let (let+) x f = f x
+  > 
+  > let f x =
+  >   let+ x in
+  >   x
+  > EOF
+
+FIXME: Should locate to the `x` in `f x`, not say that we are at the definition
```
voodoos: This is not a fixme anymore, tight ? 
```suggestion
Should locate to the `x` in `f x`, not say that we are at the definition
```

tests/test-dirs/punning.t
```diff
@@ -0,0 +1,53 @@
...
+  $ $MERLIN single locate -position 4:7 -filename let_punning.ml < let_punning.ml | jq .value
+  {
+    "file": "$TESTCASE_ROOT/let_punning.ml",
+    "pos": {
+      "line": 3,
+      "col": 6
+    }
+  }
+
```
voodoos, line 17: Of course this change impacts other queries too, and some might expect the new value to be considered, not the punned one. I can think of occurrences, but maybe others too ?

Fixing Locate at that costs feels like a fair payment, but ideally we could fix one without breaking the other.

```suggestion

FIXME Since #2066 this shows the occurrences of the punned value (lines 3 and
4), not of the pattern (lines 4 and 5). Unlike for the locate query this does
not feel like the more natural choice. We should think of a more flexible fix at
some point.
  $ $MERLIN single occurrences -identifier-at 4:7 -filename let_punning.ml < let_punning.ml | jq .value
  [
    {
      "start": {
        "line": 3,
        "col": 6
      },
      "end": {
        "line": 3,
        "col": 7
      },
      "stale": false
    },
    {
      "start": {
        "line": 4,
        "col": 7
      },
      "end": {
        "line": 4,
        "col": 8
      },
      "stale": false
    }
  ]

```

voodoos (commented):
I have mixed feeling because this makes the behavior of [occurrences](https://github.com/ocaml/merlin/pull/2066#discussion_r3288944660) (and maybe other queries? we should think about it!) less intuitive. I do think locate is the most important of them so this is an improvement, just one that is not entirely satisfying.

voodoos:
@panglesd found a much more annoying regression, this also changes the type returned by type-enclosing on the punned value. He his going to try a better fix.

panglesd:
Thanks for the review! Yes, this is annoying especially for type-enclosing: We want the type of the pattern to be displayed.

I've changed the PR in the following way:

- Added a test for type enclosing before any fix
- Kept the first implementation to show it broke the added test
- Revert the first implementation
- Made another implementation, where one can pass a "tie breaker" function to use when two nodes have the same location.

This way, we can select the "expression" node in case of locate query, and the pattern node otherwise. The "oneliner" fix is not anymore a oneliner, but I think it is more correct and hopefully the tie-breaking mechanism can be useful in other contexts.

src/kernel/mbrowse.ml
```diff
@@ -102,15 +101,30 @@ let select_leafs pos root =
   (try traverse root with Exit -> ());
   !branches
 
-let compare_locations pos l1 l2 =
+module Tie_breaker = struct
+  type tie_break = [ `Prefer_first | `Prefer_second ]
```
voodoos: 🏓

src/kernel/mbrowse.mli
```diff
@@ -39,16 +39,41 @@ val drop_leaf : t -> t option
 
 (* Navigate through tree *)
 
+module Tie_breaker : sig
+  (** Some nodes in the typedtree may share the same, non-ghost, location, while
+      being different. One example of this is let-punning:
+
```
voodoos, line 45: For science, can we have more than two overlapped nodes ?

src/analysis/locate.ml
```diff
@@ -845,7 +845,10 @@ let infer_namespace ?namespaces ~pos lid browse is_label =
       `Error `Missing_labels_namespace)
   | None -> (
     match
-      (Context.inspect_browse_tree ~cursor:pos lid [ browse ], is_label)
+      ( Context.inspect_browse_tree
+          ~disambiguate:Mbrowse.Tie_breaker.prefer_expression ~cursor:pos lid
+          [ browse ],
+        is_label )
```
voodoos: Nitpick: maybe this would be more readable by defining disambiguate before the match expression ? Or even the complete fst of the pair.

voodoos (approved):
Thanks, that looks much better !

The next step is to check which other places could benefit from this (aka record labels and labelled arguments).

src/kernel/mbrowse.ml
xvw: Did we really need polymorphic variant here?
Is that just for avoiding repeating the path `Mbrowse.Tie_breaking...` ?

panglesd: No I don't think we need it, and the path is probably not needed anyway with type-directed disambiguation of constructors. I think I started with it before I defined the type and it stuck.
I will turn it into regular variants

src/kernel/mbrowse.mli
panglesd: I... don't think so, unless you take ppx rewriting into account.

But maybe it can, I don't know. For instance, I think defining an object defines internally: an object, a type and a class type? Whose locations might coincide? (completely outside of my comfort zone).

Anyway, the tie-breaker compares nodes two-by-two, carrying the winner to the next match, until there is no challenger left. That is less expressive than taking all overlapping nodes, but I think it is expressive enough.

voodoos (approved):
Thanks ! Just need to disable the test on windows or sanitize jq output to only print the basename 🫠

panglesd:
To fix the Windows CI I omitted the path entirely, anyway there is only one file in this test!

Issue #1927, opened 2025-04-28
Signature help does not work in the presence of type aliases


ddickstein:
Signature help appears for `M.f`:
```ocaml
module M : sig
  val f : int -> unit
end = struct
  let f (_ : int) = ()
end

let () = M.f 5
```

Signature help does not appear for `M.f`:
```ocaml
type t = int -> unit

module M : sig
  val f : t
end = struct
  let f (_ : int) = ()
end

let () = M.f 5
```

PR #2067, opened 2026-05-11
Fix signature-help with type aliases

lyrm:
Fixes issue #1927.

## What changes
`signature help` now expands types once to decide whether a value is a function:

```ocaml
type t = int -> unit
let f : t = fun _ -> ()
let _ = f  (* before: nothing; after: f : int -> unit *)
```

Same when there is an alias exposed mid-call: if the call supplies more arguments than the visible arrows allow, the alias is expanded.

```ocaml
type t = int -> unit
let f (_ : int) : t = …
let _ = f 1 2  (* f : int -> int -> unit *)
```

## What have not changed

```ocaml
type t = int * int
let f (_ : t) = ()         
let _ = f  (* f : t -> unit, not f : int * int -> unit *)
```
```ocaml
type t = int -> unit
let f (_ : int) : t = unit 
let _ = f  (* f : int -> t` *)
```

## What is still broken? (or is it on purpose?)

```ocaml
let f ~x = x + 1     
let _ = f  (* nothing *)
```

Even so
```ocaml
let f ~x = x + 1     
let _ = f  ~  (* f: x:int -> int *)
```
and 
```ocaml
type t = x:int -> int
let f : t = fun ~x -> x + 1     
let _ = f  ~  (* f: x:int -> int *)
```
both work.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+Unreleased
+============
+  + merlin library
+    - Fix signature-help with type aliases (#2067, fixes #1927)
+
 merlin 5.7.1
 ============
 Thu Apr 30 14:15:42 CEST 2026
--- a/src/analysis/signature_help.ml
+++ b/src/analysis/signature_help.ml
@@ -47,7 +47,6 @@ let pp_type env ppf ty =
 let rec type_is_arrow ty =
   match Types.get_desc ty with
   | Tarrow _ -> true
-  | Tlink ty -> type_is_arrow ty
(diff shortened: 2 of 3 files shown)


xvw (approved):
The implementation seems reasonnable (and the test suite convincing). Thanks a lot.

voodoos (approved):
Thanks, looks reasonable to me.

voodoos:
@lyrm can you rebase and add a change entry ?

voodoos:
Thanks !

Issue #1404, opened 2021-11-17
discrepancies between locate and occurrences


mattiase:
Consider the code fragment `x+y`.
With the cursor on `+`, `locate` will find the definition of the operator `+`, whereas `occurrences` will find all occurrences of `x`.
With the cursor on `y`, `locate` picks `y` but `occurrences` picks `+`.

I'd say `locate` is right; ie, prefer a thing starting at the position to one ending there.

voodoos:
ack. I added a test reproducing the issue.

voodoos:
Fixed by #1757

PR #2073, opened 2026-05-26
Remove FIXME line and add 2 additional tests.

lyrm:
This is related to issue #1404, which was already fixed. This PR just removes the FIXME line on the tests and adds 2 extra tests.

Issue #1404 can be closed.


voodoos:
Thanks !

PR #2075, opened 2026-05-26
Add test for polymorphic variant completion (fix #298)

### LINKED ISSUE #298: Polymorphic variant name completion
seems unsupported. Related to discussion in #282.

panglesd:
Unless I misunderstand what completion on polymorphic variant is, issue #298 can be closed already, and this demonstrates it by adding a test.


voodoos:
Thanks !

PR #2078, opened 2026-06-12
Decouple no-dune.t test from dune internal paths

shonfeder:
Dune doesn't provide the layout of its install directories as part of its public API. These internals are changing in dune 3.24, as a result of https://github.com/ocaml/dune/pull/14432,  so this path will change. Using `dune describe location` uses dunes stable API to find the path for the needed binaries.

The need for this update was identified in the revdeps tests on the pre-releas for dune 3.24 in https://github.com/ocaml/opam-repository/pull/29996#issuecomment-4651407043 .

Thanks to @Alizter for diagnosing the affect of the breaking change here.

### DIFF
--- a/tests/test-dirs/config/no-dune.t
+++ b/tests/test-dirs/config/no-dune.t
@@ -1,6 +1,6 @@
   $ mkdir bin
-  $ cp ../../../../install/default/bin/ocamlmerlin bin/ocamlmerlin
-  $ cp ../../../../install/default/bin/ocamlmerlin-server bin/ocamlmerlin-server
+  $ cp $(dune describe location ocamlmerlin 2>&1) bin/ocamlmerlin
+  $ cp $(dune describe location ocamlmerlin-server 2>&1) bin/ocamlmerlin-server
 
   $ cat >dune-project <<EOF
   > (lang dune 2.0)
--- a/tests/test-dirs/config/no-dune.t
+++ b/tests/test-dirs/config/no-dune.t
@@ -1,6 +1,6 @@
   $ mkdir bin
-  $ cp $(dune describe location ocamlmerlin 2>&1) bin/ocamlmerlin
-  $ cp $(dune describe location ocamlmerlin-server 2>&1) bin/ocamlmerlin-server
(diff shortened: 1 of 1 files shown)


tests/test-dirs/config/no-dune.t
```diff
@@ -1,6 +1,6 @@
   $ mkdir bin
-  $ cp ../../../../install/default/bin/ocamlmerlin bin/ocamlmerlin
-  $ cp ../../../../install/default/bin/ocamlmerlin-server bin/ocamlmerlin-server
```
shonfeder, line 3: `dune describe location` outputs to stderr (for no good reason I can think of). This should be fixed at some point, but this will continue to work when when it is.

tests/test-dirs/config/no-dune.t
```diff
@@ -1,6 +1,6 @@
...
+  $ cp $(dune describe location ocamlmerlin 2>&1) bin/ocamlmerlin
+  $ cp $(dune describe location ocamlmerlin-server 2>&1) bin/ocamlmerlin-server
```
Alizter: Another way to do this, which will likely work better on Windows, is to do 

```suggestion
  $ cp $(which ocamlmerlin) bin/ocamlmerlin
  $ cp $(which ocamlmerlin-server) bin/ocamlmerlin-server
```
There is a `(cram)` stanza in `test/` that adds this dependency anyway, so it should appear in `PATH`. You just need to find it and copy it out to the `bin/` dir for this test.

shonfeder: Yes, I initially did this with `which`. However, I hoped that using dune may be a bit less permissive (i.e. avoid reporting locations from the system, to give better isolation). I just tried tho and see that this is not true: it will also return the location for things installed by the system.

I'll leave it the maintainers to decide on their preferred course here.

shonfeder: Switched this to `which` after letting it percolate, since `dune describe` doesn't bring any advantages here and it's simpler. Thanks :)

voodoos (approved):
Thanks a lot @shonfeder and @Alizter !

PR #2079, opened 2026-06-15
Add a LRU cache to limit indexing memory usage.

voodoos:
- Change the way small values are handled by storing them along their parent.
- Make filenames relative to the current working directory of the indexer.
- Disable related-uids compression

This work is based on #2050

### DIFF
--- a/src/analysis/occurrences.ml
+++ b/src/analysis/occurrences.ml
@@ -224,30 +224,26 @@ let get_external_locs ~(config : Mconfig.t) ~current_buffer_path uid :
 let lookup_related_uids_in_indexes ~(config : Mconfig.t) uid =
   let title = "lookup_related_uids_in_indexes" in
   let open Index_format in
-  let store = ref (Uid_map.empty ()) in
-  let related_uids =
-    List.fold_left ~init:(Uid_map.empty ()) config.merlin.index_files
-      ~f:(fun acc index_file ->
+  let store, related_uids =
+    List.fold_left
+      ~init:(Uid_map.empty (), Uid_map.empty ())
+      config.merlin.index_files
+      ~f:(fun (store, acc) index_file ->
         try
           let index = Index_cache.read index_file in
-          store := Union_find.merge !store index.related_uids_store;
-          Uid_map.union
-            (fun _ a b ->
-              let store', v = Union_find.union !store a b in
-              store := store';
-              Some v)
-            index.related_uids acc
+          Union_find.merge_union store index.related_uids
+            index.related_uids_store acc
         with
         | Index_format.Not_an_index _
         | Sys_error _
         | Granular_marshal.Outdated_store _
         ->
           log ~title "Could not load index %s" index_file;
-          acc)
+          (store, acc))
   in
   Uid_map.find_opt uid related_uids
   |> Option.value_map ~default:[] ~f:(fun x ->
-      x |> Union_find.get !store |> Uid_set.to_list)
+      x |> Union_find.get store |> Uid_set.to_list)
 
 let find_linked_uids ~config ~scope ~name uid =
   let title = "find_linked_uids" in
--- a/src/analysis/occurrences.ml
+++ b/src/analysis/occurrences.ml
@@ -224,13 +224,18 @@ let get_external_locs ~(config : Mconfig.t) ~current_buffer_path uid :
 let lookup_related_uids_in_indexes ~(config : Mconfig.t) uid =
   let title = "lookup_related_uids_in_indexes" in
   let open Index_format in
+  let store = ref (Uid_map.empty ()) in
   let related_uids =
     List.fold_left ~init:(Uid_map.empty ()) config.merlin.index_files
       ~f:(fun acc index_file ->
         try
           let index = Index_cache.read index_file in
+          store := Union_find.merge !store index.related_uids_store;
           Uid_map.union
-            (fun _ a b -> Some (Union_find.union a b))
+            (fun _ a b ->
+              let store', v = Union_find.union !store a b in
+              store := store';
+              Some v)
             index.related_uids acc
         with
         | Index_format.Not_an_index _
@@ -242,7 +247,7 @@ let lookup_related_uids_in_indexes ~(config : Mconfig.t) uid =
   in
   Uid_map.find_opt uid related_uids
   |> Option.value_map ~default:[] ~f:(fun x ->
-      x |> Union_find.get |> Uid_set.to_list)
+      x |> Union_find.get !store |> Uid_set.to_list)
 
 let find_linked_uids ~config ~scope ~name uid =
   let title = "find_linked_uids" in
--- a/src/analysis/occurrences.ml
+++ b/src/analysis/occurrences.ml
@@ -243,7 +243,7 @@ let lookup_related_uids_in_indexes ~(config : Mconfig.t) uid =
   in
   Uid_map.find_opt uid related_uids
   |> Option.value_map ~default:[] ~f:(fun x ->
-      x |> Union_find.get store |> Uid_set.to_list)
+      x |> Union_find.get store |> Uid_set.elements)
 
 let find_linked_uids ~config ~scope ~name uid =
   let title = "find_linked_uids" in
--- a/src/index-format/dbllist.ml
+++ b/src/index-format/dbllist.ml
@@ -1,0 +1,145 @@
+type 'a cell =
+  { content : 'a; weight : int; mutable prev : 'a cell; mutable next : 'a cell }
+
+type stats =
+  { mutable total_cap : int;
+    mutable promote_count : int;
+    mutable add_count : int;
+    mutable discard_count : int;
+    mutable add_size : int;
+    mutable discarded_size : int
+  }
+
+type 'a dbll =
+  | Nil of int
+  | List of { first : 'a cell; last : 'a cell; size : int; cap : int }
+
+type 'a t = { mutable dbll : 'a dbll; stats : stats }
+
+exception Action_on_empty_list of string
+
+let pp_stats t =
+  let size =
+    match t.dbll with
+    | Nil _ -> 0
+    | List l -> l.size
+  in
+  Printf.eprintf
+    "total_cap \t\t: %d\n\
+     size \t\t: %d\n\
+     promote_count \t: %d\n\
+     add_count \t\t: %d\n\
+     discard_count \t: %d\n\
+     add_size \t\t: %d\n\
+     discard_size \t: %d\n\
+     volume_conservation \t: %d = %d + %d : %b\n\
+     %!"
+    t.stats.total_cap size t.stats.promote_count t.stats.add_count
+    t.stats.discard_count t.stats.add_size t.stats.discarded_size
+    t.stats.add_size t.stats.discarded_size size
+    (t.stats.add_size = t.stats.discarded_size + size)
+
+let create cap =
+  let stats =
+    { total_cap = cap;
+      promote_count = 0;
+      add_count = 0;
+      discard_count = 0;
+      add_size = 0;
+      discarded_size = 0
+    }
+  in
+  { dbll = Nil cap; stats }
+
+let add_front t (v, w) =
+  t.stats.add_count <- t.stats.add_count + 1;
+  t.stats.add_size <- t.stats.add_size + w;
+  match t.dbll with
+  | Nil cap ->
+    let rec c = { content = v; weight = w; prev = c; next = c } in
+    t.dbll <- List { first = c; last = c; size = w; cap };
+    c
+  | List l ->
+    let rec new_first =
+      { content = v; weight = w; prev = new_first; next = l.first }
+    in
+    l.first.prev <- new_first;
+    t.dbll <-
+      List { first = new_first; last = l.last; size = l.size + w; cap = l.cap };
+    new_first
+
+let discard t =
+  t.stats.discard_count <- t.stats.discard_count + 1;
+  match t.dbll with
+  | Nil _ ->
+    raise
+      (Action_on_empty_list
+         "Unable to discard the last element, the doubly linked list is empty.")
+  | List l ->
+    if l.first == l.last then (
+      t.dbll <- Nil l.cap;
+      t.stats.discarded_size <- t.stats.discarded_size + l.last.weight;
+      l.last.content)
+    else
+      let discarded_value = l.last.content in
+      let discarded_weight = l.last.weight in
+      t.stats.discarded_size <- t.stats.discarded_size + discarded_weight;
+      let new_last = l.last.prev in
+      (* TODO Should we explicitely disconnect last's pointers ? *)
+      new_last.next <- new_last;
+      (* TODO Int.max 0 (l.size - discarded_weight) does seems useless.
+         We could use an assert to check it. *)
+      (* Unlinking the discaded cell is not strictly necessary but not doing it
+         could lead to memory leaks if the user of the cache keeps a reference
+         to the cell. *)
+      l.last.next <- l.last;
+      l.last.prev <- l.last;
+      t.dbll <-
+        List
+          { first = l.first;
+            last = new_last;
+            size = Int.max 0 (l.size - discarded_weight);
+            cap = l.cap
+          };
+      discarded_value
+
+let discard_size t s =
+  (* this is fold not iter *)
+  let rec iter acc t =
+    match t.dbll with
+    | Nil _ -> acc
+    | List l -> if l.size + s <= l.cap then acc else iter (discard t :: acc) t
+  in
+  iter [] t
+
+let promote t c =
+  t.stats.promote_count <- t.stats.promote_count + 1;
+  match t.dbll with
+  | Nil _ ->
+    raise
+      (Action_on_empty_list
+         "Unable to promote a cell, the doubly linked list is empty.")
+  | List l ->
+    if l.first == c then ()
+    else if l.last == c then (
+      let new_last = l.last.prev in
+      new_last.next <- new_last;
+      let new_first = c in
+      new_first.next <- l.first;
+      new_first.prev <- new_first;
+      l.first.prev <- new_first;
+      t.dbll <-
+        List { first = new_first; last = new_last; size = l.size; cap = l.cap })
+    else
+      let voisin_prev = c.prev in
+      let voisin_next = c.next in
+      voisin_prev.next <- voisin_next;
+      voisin_next.prev <- voisin_prev;
+      let new_first = c in
+      new_first.prev <- new_first;
+      new_first.next <- l.first;
+      l.first.prev <- new_first;
+      t.dbll <-
+        List { first = new_first; last = l.last; size = l.size; cap = l.cap }
+
+let get c = c.content
--- a/src/index-format/granular_marshal.ml
+++ b/src/index-format/granular_marshal.ml
@@ -63,7 +63,7 @@ and 'a repr =
   (*
    * In-memory realm
    *)
-  | On_disk of { store : store; loc : int; schema : 'a schema }
+  | On_disk of { store : store; loc : int; schema : 'a schema option }
       (** {i in-memory} A value that can be read from the disk. *)
   | On_disk_small of
       { parent : parent_link; (* Either the parent or an On_disk_ptr *)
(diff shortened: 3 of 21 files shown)


voodoos:
cc @art-w if you want to have a quick 🍔 look ?

src/index-format/granular_marshal.ml
```diff
@@ -92,137 +191,399 @@ let open_store store =
...
+        | Duplicate _ -> (* These are already "clean" *) ())
+  }
+
+let on_cache_discard (Cached (link, loc, store, schema)) =
+  (* This also free the smalls that are stored in the link *)
+  match !schema with
+  | Some schema -> link := On_disk { store; loc; schema }
+  | None ->
+    link :=
+      On_disk_ptr { filename = store.filename; id = store.id; loc; pos = None }
```
art-w: Would it simplify the code overall if `On_disk` had a `schema option`? (to avoid having to default to `On_disk_ptr`)

src/index-format/granular_marshal.ml
```diff
@@ -4,43 +4,148 @@ type store = { filename : string; id : int; cache : cache }
...
+          optional `pos` field is used to target small values. *)
+  (*
+   * In-memory realm
+   *)
   | On_disk of { store : store; loc : int; schema : 'a schema }
-  | On_disk_ptr of { filename : string; loc : int; id : int }
+      (** {i in-memory} A value that can be read from the disk. *)
+  | On_disk_small of
+      { store : store;
+        loc : int;
```
art-w: nitpick: the `store/loc` should be the retrievable from the `parent`

src/index-format/granular_marshal.ml
```diff
@@ -92,137 +191,399 @@ let open_store store =
...
+  add_to_cache v lnk ~loc store ~size small_children None;
+  small_children
+
+(** Fetch the parent of a small value in order to read its smalls. If the parent
+  has not yet been loaded in memory it will be read from the disk and kept dirty
+  because its schema is unknown.*)
+let fetch_parent : parent_link -> any_value array =
+ fun (PLink parent_link) ->
+  match !parent_link with
+  | In_cache (_, _, _, smalls) -> smalls
```
art-w: Should we also LRU promote the parent?

src/index-format/granular_marshal.ml
```diff
@@ -92,137 +191,399 @@ let open_store store =
...
-              Cache.add store.cache loc (Link (lnk, type_id)))
-          | In_memory _ | In_memory_reused _ | On_disk _ | Duplicate _ -> ()
-          | On_disk_ptr { filename; loc; id } ->
-            let store = { filename; id; cache = Cache_cache.read filename } in
-            lnk := On_disk { store; loc; schema }
-          | Placeholder -> invalid_arg "Granular_marshal.read_loc: Placeholder")
-    }
+let resolve_filename store ~filename =
+  if Filename.is_relative filename then
+    Filename.concat (Filename.dirname store.filename) filename
```
art-w, line 190: nitpick: I believe `filename` could contain `../../` (introduced in `relativize` below), which could mess up `relativize` when we write later? (I guess the result wouldn't be wrong but would have `../../path/../otherpath`)

src/analysis/occurrences.ml
```diff
@@ -223,13 +223,18 @@ let get_external_locs ~(config : Mconfig.t) ~current_buffer_path uid :
...
       ~f:(fun acc index_file ->
         try
           let index = Index_cache.read index_file in
+          store := Union_find.merge !store index.related_uids_store;
           Uid_map.union
-            (fun _ a b -> Some (Union_find.union a b))
+            (fun _ a b ->
+              let store', v = Union_find.union !store a b in
+              store := store';
+              Some v)
```
art-w: It looks like the same `merge + union` code is duplicated in `index.ml`, perhaps this should live in `Union_find`

src/index-format/dbllist.ml
```diff
@@ -0,0 +1,145 @@
...
+      let discarded_weight = l.last.weight in
+      t.stats.discarded_size <- t.stats.discarded_size + discarded_weight;
+      let new_last = l.last.prev in
+      (* TODO Should we explicitely disconnect last's pointers ? *)
+      new_last.next <- new_last;
+      (* TODO Int.max 0 (l.size - discarded_weight) does seems useless.
+         We could use an assert to check it. *)
+      (* Unlinking the discaded cell is not strictly necessary but not doing it
+         could lead to memory leaks if the user of the cache keeps a reference
+         to the cell. *)
```
art-w, line 94: There are some TODOs here with nice ideas :) (but not essentials)

src/index-format/granular_marshal.ml
```diff
@@ -4,43 +4,148 @@ type store = { filename : string; id : int; cache : cache }
...
+        small_pos : int;
+        small_schema : 'a schema
+      }
+      (** {i in-memory} A small value whose parent can be read from the disk. *)
   | In_memory of 'a
+      (** {i in-memory} A value that has been created in memory. *)
   | In_memory_reused of 'a
+      (** {i in-memory} A value that has been created in memory and is used
+          multiple times. *)
+  | In_cache of 'a * value_status * cached Dbllist.cell * any_value array
```
art-w: (I suspect that `In_cache` should just hold onto everything that's currently in `cached`, so that we don't need to `Dbllist.get` everywhere... The `Dllist.cell` can (recursively) hold a pointer to its link)

src/index-format/granular_marshal.ml
```diff
@@ -92,137 +191,399 @@ let open_store store =
...
+          match Cache.find_opt store.cache loc with
+          | Some (Link (type b) ((lnk', Some type_id') : b link * _)) -> (
+            match Type.Id.provably_equal type_id type_id' with
+            | Some (Equal : (a link, b link) Type.eq) ->
+              lnk := Duplicate (normalize lnk')
+            | None ->
+              invalid_arg "Granular_marshal.read_loc: reuse of a different type"
+            )
+          | Some _ ->
+            invalid_arg "Granular_marshal.read_loc: reuse of a different type"
```
art-w: Isn't that case matching "we know that loc but didn't know the schema at the time it was cached"? (in which case we want to enrich the existing `lnk'` with a type_id and clean schema)

src/index-format/granular_marshal.ml
```diff
@@ -92,137 +191,399 @@ let open_store store =
...
+          lnk :=
+            On_disk_small
+              { store;
+                loc;
+                parent;
+                small_type_id = type_id;
+                small_schema = schema;
+                small_pos = pos
+              }
+        | Serialized { loc } -> lnk := On_disk { store; loc; schema }
```
art-w: `Serialized` should use the same cache logic as `Serialized_reused`: before we could rely on deduplicate to flag links worth caching, but now we have both the LRU which could trigger multiple reads to the same loc and the smalls pointing to parents (which are potentially not marked as `Serialized_reused`)

src/index-format/granular_marshal.ml
voodoos: Done in [edc4fe2](https://github.com/ocaml/merlin/pull/2079/commits/edc4fe27780aca784a1b9536f6668073e5a596d8)

src/index-format/granular_marshal.ml
voodoos: Sounds... right ?

Done in db488542

src/index-format/granular_marshal.ml
voodoos: Done in 4e03ff14
The main advantage is that now `On_disk_ptr` is only appearing in the "on disk" realm, not "in memory" anymore. Such links are only created at write-time.

src/analysis/occurrences.ml
voodoos: Done in ffe6b847

src/index-format/granular_marshal.ml
voodoos: Hum, right, I did that in 544c9737

src/index-format/granular_marshal.ml
voodoos: Done in f990efd3

voodoos:
Thanks a lot for the review @art-w. I think I addressed all the important comments. These are probably worth a re-review 🙂

src/index-format/granular_marshal.ml
voodoos: Done in de52e53a

voodoos:
I ran two benchmarks with a LRU large enough to store the entire date.

One that builds Merlin's `@ocaml-index`: in this build, there are no aggregations, only things built In_memory and written to the disk.

| Command | Mean [s] | Min [s] | Max [s] | 
|:---|---:|---:|---:|
| Before pointing indexes | 10.570 ± 0.175 | 10.366 | 10.935 |
| With pointing indexes | 10.533 ± 0.392 | 10.151 | 11.528 | 
| With pointing + LRU | 10.951 ± 0.347 | 10.596 | 11.619 | 

One that aggregates all of Merlin indexes: 

| Command | Mean [ms] | Min [ms] | Max [ms] | 
|:---|---:|---:|---:|
| Before pointing indexes | 768.5 ± 43.1 | 707.9 | 829.1 |
| With pointing indexes | 229.4 ± 10.3 | 217.2 | 250.3 | 
| With pointing + LRU | 354.9 ± 19.2 | 340.4 | 406.1 | 

So we loose a bit of the improvements that pointing indexes brought, but we are still much faster at aggregating than before this work. 

The most likely explanation for the slowdown is the move to a functional union-find, which is a necessary evil.

I think these results are quite good overall, with an aggregation that is twice as fast as before.

voodoos:
And index sizes:

| Version | total index size |
|---|---|
| Before pointing indexes | 25M |
| With pointing indexes | 25M | 
| With pointing + LRU | 27M | 

These are the index built by `dune @ocaml-index` so no aggregates which expect that pointing indexes has no impact.

And the size of the aggregate of all these indexes:

| Version | total index size |
|---|---|
| Before pointing indexes | 25M |
| With pointing indexes | 4.9M | 
| With pointing + LRU | 4.1M |

src/index-format/union_find.ml
```diff
@@ -1,40 +1,88 @@
...
   | Link _ -> assert false
+
+let union ~f store x y =
+  let store, x = find_and_compress store x in
+  let store, y = find_and_compress store y in
+  if Uid.equal x y then (store, x)
+  else
+    match (Uid_map.find x store, Uid_map.find y store) with
+    | ( Root { value = value_x; rank = rank_x },
+        Root { value = value_y; rank = rank_y } ) ->
```
art-w, line 45: nitpick: Maybe `find_and_compress` could return the root it found instead of requiring a second `find`

src/ocaml-index/bin/ocaml_index.ml
```diff
@@ -14,6 +14,7 @@ let root = ref ""
 let rewrite_root = ref false
 let store_shapes = ref false
 let do_not_use_cmt_loadpath = ref false
+let cache_size = ref 1_000_000
```
art-w: That looks small, it should be okay to bump it with a couple more zeros :P

src/index-format/index_format.ml
```diff
@@ -2,20 +2,42 @@ exception Not_an_index of string
...
+    (store, Granular_marshal.link root)
+
+  let get store t = Union_find.get store (Granular_marshal.fetch t)
+
+  let union store a b =
+    let open Granular_marshal in
+    let store, root =
+      Union_find.union store ~f:Uid_set.union (fetch a) (fetch b)
+    in
+    (store, link root)
```
art-w: I believe you could arbitrarily return `a` or `b`, instead of creating a new `root` link

src/index-format/index_format.ml
```diff
@@ -55,7 +79,10 @@ let index_schema (iter : Granular_marshal.iter) index =
     index.approximated;
   Uid_map.schema type_ufmap iter
     (fun iter _ v -> Union_find.schema iter v)
-    index.related_uids
+    index.related_uids;
+  Uid_map.schema type_ufstore iter
+    (fun _iter _uid _content -> ())
```
art-w: Don't we need to schema iter on the `Uid_set.t` stored in `_content`?

src/index-format/union_find.ml
```diff
@@ -1,40 +1,88 @@
...
+    if Uid_map.mem uid store then store
+    else
+      match Uid_map.find (find s2 uid) s2 with
+      | Root { value; _ } -> fst (new_root store uid value)
+      | Link _ -> assert false
+  in
+  Uid_map.fold
+    (fun uid content store ->
+      match content with
+      | Root _ -> ensure store uid
```
art-w, line 79: If `ensure` terminates early because `Uid_map.mem uid s1` and there are no `Link` to that `Root` in `s2`, then I believe we need to compute the `Uid_set.union` of s2 with s1? (otherwise `store` will only have the `s1` set?)

(or is it the case that if we only have a root-with-no-link then the uid set is the singleton `{uid}` which is guaranteed to be in the s1 set anyway?)

src/index-format/granular_marshal.ml
```diff
@@ -4,43 +4,154 @@ type store = { filename : string; id : int; cache : cache }
...
-  | On_disk of { store : store; loc : int; schema : 'a schema }
-  | On_disk_ptr of { filename : string; loc : int; id : int }
+      (** {i on-disk} A pointer to serialized value that is used multiple times.
+          Allow for better file compression and perofrmance. *)
+  | Small of int
+      (** {i on-disk} A "small value" placeholder. Contains the index of this
+          small's actual value in the array stored by its parent value. *)
+  | Serialized_small of { loc : int; pos : int }
+      (** {i on-disk} A pointer to an already serialized small value. *)
+  | On_disk_ptr of { filename : string; loc : int; id : int; pos : int option }
```
art-w: Out of curiosity, is the 2mb increase in filesize caused by the extra `pos: int option`? (in which case we could consider splitting in another case `On_disk_small_ptr` like for `On_disk`/`On_disk_small`)

src/index-format/granular_marshal.ml
```diff
@@ -92,137 +197,397 @@ let open_store store =
...
+              lnk := Duplicate (normalize lnk')
+            | None ->
+              invalid_arg "Granular_marshal.read_loc: reuse of a different type"
+            )
+          | Some (Link (lnk', None)) ->
+            let lnk' = Obj.magic lnk' in
+            upgrade_reused_link_schema lnk' store schema;
+            Cache.replace store.cache loc (Link (lnk', Some type_id));
+            lnk := Duplicate (normalize lnk')
+          | _ -> lnk := On_disk { store; loc; schema = Some schema })
```
art-w: It looks similar to code above ("find in cache -> mark as duplicate or store in cache"), but here it's missing the `Cache.replace store.cache loc (Link ...)` to remember the entry for the next time

src/index-format/granular_marshal.ml
```diff
@@ -92,137 +197,397 @@ let open_store store =
...
     }
+  and output_and_mark (V v) (small_children : any_val_link list) =
+    let new_smalls =
+      (* Some smalls might have been already serialized by another value *)
+      List.filter
+        (fun (Vlink (_v, lnk)) ->
+          match !lnk with
+          | On_disk_ptr { pos = Some _; _ } ->
+            (* This small has already been serialized by another owner *) false
+          | _ -> true)
```
art-w, line 501: Should we also skip the `Serialized_small` which would indicate that small was adopted by another parent?

src/index-format/granular_marshal.ml
```diff
@@ -4,43 +4,154 @@ type store = { filename : string; id : int; cache : cache }
...
+      (** {i in-memory} A value that has been created in memory. *)
   | In_memory_reused of 'a
+      (** {i in-memory} A value that has been created in memory and is used
+          multiple times. *)
+  | In_cache of
+      { value : 'a;
+        status : value_status;
+        store : store;
+        loc : int;
+        schema : 'a schema option ref;
```
art-w: I might have missed something, but is the `schema` used? (it's a bit redundant with the `status`)

src/index-format/granular_marshal.ml
```diff
@@ -4,43 +4,154 @@ type store = { filename : string; id : int; cache : cache }
...
+  their schema known.
+*)
 and 'a repr =
-  | Small of 'a
+  (*
+   * On-disk realm
+   *)
   | Serialized of { loc : int }
+      (** {i on-disk} A pointer to a serialized value in the file. *)
   | Serialized_reused of { loc : int }
```
art-w: nitpick: I suspect we could remove `Serialized_reused` in favor of always using `Serialized`, it seems like the distinction is not used anymore

art-w (commented):
Thanks! It's looking good, I only have some minor questions/suggestions

src/ocaml-index/bin/ocaml_index.ml
voodoos: Ah yes, in fact I already did that in a local commit. I changed the unit of the flag to kb and use a 1gb default:
a08ac5a2a3ae31e6136afa13ea061aa7c2b1d160

src/index-format/union_find.ml
voodoos: Done in [b187c12](https://github.com/ocaml/merlin/pull/2079/commits/b187c122923db0534bb1da1e4516e40043382c2e)

src/index-format/index_format.ml
voodoos: Done in dfa6001588622ea17bdc3c6944b4fc0b8955332a

src/index-format/index_format.ml
voodoos: Currently the `Uid_set` is not granular: `module Uid_set = Shape.Uid.Set`.
Should we make it granular ? I did that in d8aa258d6aca3331fb056044c7a28ac3c1d50dd0

It required removing some fast-paths in Union_find.union that were relying on polymorphic comparison. If it turns out to be important we can add an equal function to granular set.

src/index-format/union_find.ml
voodoos: I think we don't even have "roots with no link" since we always register at least a pair of uids during the indexing...

src/index-format/granular_marshal.ml
voodoos: Good cache (ahah). I fixed it and factored it out: 1d5887046d6bdb2ef34a90188944b4ac7e7b7f75

src/index-format/granular_marshal.ml
voodoos: Seems fair because they were written once, not sure if that does happen however ? I made the change (which does not impact file size)

src/index-format/granular_marshal.ml
voodoos: It is used when unloading stuff from the cache and rewrite the link to an "on_disk".

```ocaml
let add_to_cache value lnk ~loc store ~size small_values schema =
  let discarded = Dbllist.discard_size (get_lru ()) size in
  let status = if Option.is_none schema then Dirty_unknown_schema else Clean in
  let cell = Dbllist.add_front (get_lru ()) (Cached lnk, size) in
  List.iter
    (fun (Cached link) ->
      (* This also free the smalls that are stored in the link *)
      match !link with
      | In_cache { loc; store; schema; _ } ->
        link := On_disk { store; loc; schema = !schema }
      | _ -> assert false)
    discarded;
  lnk :=
    In_cache
      { value; status; loc; store; schema = ref schema; cell; small_values }
 ```
 
 And indeed, when we "clean" things we don't update that schema:
 ```ocaml
   | In_cache ({ value; status = Dirty_unknown_schema; _ } as c) ->
    (* If we already have the value in cache we must clean it. *)
    schema (disk_to_memory_iter store (PLink lnk)) value;
    lnk := In_cache { c with status = Clean }
 ```
    
  I guess we should ? And then we could use the presence / absence of a schema as a marker for cleanliness ?

src/index-format/granular_marshal.ml
voodoos: I just tried and this has no impact on the file size... Do you have a preference between the two representations ?

src/index-format/granular_marshal.ml
voodoos: I did that in afcf26da8b99afa59e2156ea10d3d959b0e8d873

voodoos:
My changes following the latest review do not have any significant performance impact.

src/index-format/granular_marshal.ml
art-w: No I'm fine with either if it doesn't impact the file size :) Thanks for testing!

voodoos: Let's keep the new version then

src/index-format/union_find.ml
voodoos: Fixed in [db3ab43](https://github.com/ocaml/merlin/pull/2079/commits/db3ab435a0e59582ac3637847a90dd80a9d92f60)

src/index-format/granular_marshal.ml
voodoos: And I removed it in 12def0fb5a5f7eb800cc6615fed45452604032e2

voodoos:
Ok, I think I addressed all your remaining comments @art-w !

art-w (approved):
Thanks, great work!

voodoos:
Thank you for the thorough review @art-w. Merging now.

Issue #2006, opened 2025-11-21
CI failure


PizieDust:
When the CI runs, it fails with the log:

```sh
Run opam exec -- dune build @check
Entering directory '/home/runner/work/merlin/merlin'
File "irmin/duniverse/ppxlib/astlib/dune", lines 10-13, characters 0-94:
10 | (rule
11 |  (targets ast-version)
12 |  (action
13 |   (run %{ocaml} %{dep:config/gen.ml} %{ocaml_version})))
(cd _build/default/irmin/duniverse/ppxlib/astlib && /home/runner/work/merlin/merlin/_opam/bin/ocaml config/gen.ml 5.4.0)
Unknown OCaml version 5.4.0
File "irmin/duniverse/ppxlib/ast/dune", line 16, characters 5-58:
16 |      (run %{exe:pp/pp.exe} %{ocaml_version} %{input-file}))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(cd _build/.sandbox/75ac29a3e2509a443064b0b7f56d5e15/default && irmin/duniverse/ppxlib/ast/pp/pp.exe 5.4.0 irmin/duniverse/ppxlib/ast/versions.mli) > _build/.sandbox/75ac29a3e2509a443064b0b7f56d5e15/default/irmin/duniverse/ppxlib/ast/versions.pp.mli
Unknown OCaml version 5.4.0
File "irmin/duniverse/ppxlib/ast/dune", line 16, characters 5-58:
16 |      (run %{exe:pp/pp.exe} %{ocaml_version} %{input-file}))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(cd _build/.sandbox/1d3f17d59c00557c14acb4bdd25f714e/default && irmin/duniverse/ppxlib/ast/pp/pp.exe 5.4.0 irmin/duniverse/ppxlib/ast/versions.ml) > _build/.sandbox/1d3f17d59c00557c14acb4bdd25f714e/default/irmin/duniverse/ppxlib/ast/versions.pp.ml
Unknown OCaml version 5.4.0
Leaving directory '/home/runner/work/merlin/merlin'
Error: Process completed with exit code 1.
```

PizieDust:
https://github.com/ocaml/merlin/blob/main/.github/workflows/fuzzy-ci.yml#L42-L49

[...]

PR #2086, opened 2026-07-06
Fix fuzzy CI

Tim-ats-d:
- Update from OCaml 5.3 to 5.5
- Update Merl-an CLI usage
- Regenerate Irmin lock file using this method:
```bash
$ opam switch create ocaml-base-compiler.5.5.0
$ opam switch ocaml-base-compiler.5.5.0
$ git clone https://github.com/mirage/irmin
$ git checkout 7fa4b043a97944635cc100ae2e7dd85f73d8a4ce # The release commit hash of the release of Irmin version 3.10
$ opam install opam-monorepo
$ opam repository add dune-universe git+https://github.com/dune-universe/opam-overlays.git
$ opam-monorepo lock irmin irmin-pack irmin-tezos --ocaml-version=5.5.0 
```
Merging this PR should close that issue https://github.com/ocaml/merlin/issues/2006

### DIFF
--- a/.github/workflows/fuzzy-ci.yml
+++ b/.github/workflows/fuzzy-ci.yml
@@ -43,7 +43,7 @@ env:
 
   # Pinned Merl-an version commit hash
   MERL_AN_SHA: a61cb2b6c2f57baf00b88a74f67b332c459575ef
-  
+
   # The compiler version used on the respective branches. It also needs to form part of Irmin's build cache key.
   # Bump either of these whenever the compiler version is bumped on either of the two branches.
   merge_branch_COMPILER_VERSION: ocaml-base-compiler.5.5.0
@@ -130,6 +130,7 @@ jobs:
         if: steps.irmin-cache.outputs.cache-hit != 'true'
         run: |
           cp ".github/fuzzy-ci-helpers/irmin.$IRMIN_VERSION.opam.locked" irmin/irmin.opam.locked
+          cat irmin/irmin.opam.locked
 
       - name: Install opam monorepo
         if: steps.irmin-cache.outputs.cache-hit != 'true'
--- a/.github/workflows/fuzzy-ci.yml
+++ b/.github/workflows/fuzzy-ci.yml
@@ -41,6 +41,9 @@ env:
   # Irmin version and merl-an version need to be consistent for reproducibility (Irmin is used as the test code base to test `ocamlmerlin` on)
   IRMIN_VERSION: 3.10.0
 
+  # Pinned Merl-an version commit hash
+  MERL_AN_SHA: a61cb2b6c2f57baf00b88a74f67b332c459575ef
+  
   # The compiler version used on the respective branches. It also needs to form part of Irmin's build cache key.
   # Bump either of these whenever the compiler version is bumped on either of the two branches.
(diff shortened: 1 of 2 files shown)


.github/workflows/fuzzy-ci.yml
```diff
@@ -40,13 +40,11 @@ env:
 
   # Irmin version and merl-an version need to be consistent for reproducibility (Irmin is used as the test code base to test `ocamlmerlin` on)
   IRMIN_VERSION: 3.10.0
-  # TODO: Release merl-an and install a certain version instead of pinning it to a certain commit
-  MERL_AN_SHA: 344a079bf502f27e9269498db9ed5369753f2ddb
```
voodoos, line 44: I don't think removing in the pin to a specific commit is wise, we should bump it to master's HEAD instead

Tim-ats-d: This should be fixed now

Tim-ats-d:
Thanks for the review

Tim-ats-d:
Why does it fail now? :thinking: The [ppxlib version 0.38](https://github.com/ocaml-ppx/ppxlib/releases/tag/0.38.0) seems to support OCaml 5.5 and the ppxlib version has been fixed accordingly in the lock file.

voodoos:
> Why does it fail now? :thinking:

Because that's what CIs are for 😂

voodoos:
> Why does it fail now? 🤔 The [ppxlib version 0.38](https://github.com/ocaml-ppx/ppxlib/releases/tag/0.38.0) seems to support OCaml 5.5 and the ppxlib version has been fixed accordingly in the lock file.

Can you add prints right before building irmin to check the ppxlib version ?

voodoos:
image

🤷‍♂️

.github/workflows/fuzzy-ci.yml
```diff
@@ -128,7 +129,7 @@ jobs:
       - name: Get Irmin's lock files
         if: steps.irmin-cache.outputs.cache-hit != 'true'
         run: |
-          cp .github/fuzzy-ci-helpers/irmin.3.10.0.opam.locked irmin/irmin.opam.locked
+          cp .github/fuzzy-ci-helpers/irmin.$IRMIN_VERSION.opam.locked irmin/irmin.opam.locked
```
voodoos: I suspect the variable expansion doesn't work here, you might need quotes `".github/fuzzy-ci-helpers/irmin.$IRMIN_VERSION.opam.locked"`

Tim-ats-d: I've quoted all the IRMIN_VERSION variable expansion and it has no effect

Tim-ats-d:
Maybe the Opam monorepo has not been upgraded to OCaml 5.5?

voodoos:
> Maybe the Opam monorepo has not been upgraded to OCaml 5.5?

Didn't you say it used to work on your machine / fork ?

voodoos:
My new print shows the the irmin.opam.locked file we get is not the one we expect.

voodoos:
Ah I think I get it ! We checkout the main branch before copying the lock file 😅
I am going to push the lockfile first

voodoos:
🙌

Tim-ats-d:
Nice, good catch!

PR #2089, opened 2026-07-07
Fix issue 1748: locate on a open with a module shadowing

lyrm:
This is a fix for issue #1748.

### Minimal project to trigger the bug
We need 4 files: `bar.ml`, `import.ml`, `foo.ml` and `foo.mli`.

```ocaml
(* bar.ml *)
module Import = struct
  type hello = string
end

(* import.ml *)
include Bar     (* the module `Import` now contains a submodule `Import` *)

(* foo.ml  *)
open! Import

(* foo.mli *)
open! Import
```

### The bug
Calling `locate` on `Import` jumps to
- `import.ml` when called from `foo.ml`
- `bar.ml` when called from `foo.mli`

Both should behave the same and jump to `import.ml`.

The problem only occurs with shadowing: in the `.mli`, the environment used to answer the query is the one *after* the `open`, which already contains the opened `Import`, so the bare name resolves to the submodule instead of the original module. (Rename either `import.ml` or `Bar.Import` and the issue disappears.)

The same bug affects the `document` command (regression test added in `document/issue1748.t`).

## The fix
We want the query answered in the environment *right before* the `open`. That environment was not carried on the `Browse_raw.Open_description` node (only the post-open one was inherited), so the cursor landed on a node with the wrong environment. The fix makes `Open_description` carry the pre-open environment, mirroring what structure opens already get through their `Module_expr` child. This corrects `locate`, `document`, and any other command that resolves a name at the `open`.

## Use of AI
I used Claude to understand the issue and help me fix it, and I reviewed everything carefully.

### DIFF
--- a/tests/test-dirs/locate/issue1748.t
+++ b/tests/test-dirs/locate/issue1748.t
@@ -1,0 +1,67 @@
+Reproduction for https://github.com/ocaml/merlin/issues/1748
+
+When a module `Import` exists (here `import.ml`) and it includes another module
+that itself contains a nested module named `Import` (here `Bar.Import`), calling
+locate on `Import` from an `open Import` should jump to `import.ml` and not to 
+`Bar.Import`.
+
+  $ cat >dune-project <<'EOF'
+  > (lang dune 3.0)
+  > EOF
+
+  $ cat >dune <<'EOF'
+  > (library
+  >  (name jump_to_wrong_import))
+  > EOF
+
+  $ cat >bar.ml <<'EOF'
+  > module Import = struct
+  >   type hello = string
+  > end
+  > EOF
+
+  $ cat >import.ml <<'EOF'
+  > include Bar
+  > EOF
+
+  $ cat >foo.ml <<'EOF'
+  > open! Import
+  > EOF
+
+  $ cat >foo.mli <<'EOF'
+  > open! Import
+  > EOF
+
+  $ dune build @check 2>&1 | head -n 5
+
+Locating `Import` from `foo.ml` correctly jumps to `import.ml`:
+
+  $ $MERLIN single locate -position 1:10 -look-for implementation \
+  > -filename foo.ml <foo.ml
+  {
+    "class": "return",
+    "value": {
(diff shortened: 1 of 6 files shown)


voodoos (approved):
Thanks, seems good to me !

We usually try to have tests promoted with their incorrect output before the fix commit which then also includes the fixed tests' diffs. No need to redo it here though.

tests/test-dirs/locate/issue1748.t
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+Locating `Import` from `foo.mli` should also jump to `import.ml`:
+
+  $ $MERLIN single locate -position 1:10 -look-for implementation \
+  > -filename foo.mli <foo.mli
+  {
+    "class": "return",
+    "value": {
+      "file": "$TESTCASE_ROOT/import.ml",
```
voodoos, line 60: Any test priting `$TESTCASE_ROOT` will fail on windows, you should disable them in the appropriate dune files.

voodoos (changes requested):

tests/test-dirs/locate/issue1748.t
lyrm: The dune file for tests in `test-dirs/locate` already contains:
```
(cram
 (applies_to :whole_subtree)
 (enabled_if
  (<> %{os_type} Win32)))
```

Isn't that enough?

panglesd:
What happens in the case of `include`s instead of `open`?

```ocaml
(* bar.ml *)
module Import = struct
  type hello = string
end

(* import.ml *)
include Bar     (* the module `Import` now contains a submodule `Import` *)

(* foo.ml  *)
include Import

(* foo.mli *)
include Import
```
From a quick look it should have the same problem?

lyrm:
Good point, let me check!

lyrm:
So the answer is no. An `include` in a .mli would be represented as an `Include_description` containing a `module_type`, which carries the appropriate environment. Thanks for the question!

voodoos (approved):

PR #2092, opened 2026-07-09
Fix `document` returning neighboring entries' doc comments

liam923:
`document` on a variant constructor or record field sometimes returns neighboring entries' docs instead of just its own. For example, in:
```ocaml
type t =
  | A of int (** A *)
  | B of int (** B *)
  | C of int (** C *)
```
asking for the docs for `B` when in the same buffer as the declaration with give `A\nB\nC`.

There's already some test cases that cover this, and the first commit adds some additional interesting ones.

Two fixes were needed:
1. `after_only` in `get_doc` was often `false` in cases where it should be `true`.
2. A comment on the line after the declaration used to be unconditionally associated with the declaration. This was because there used to be a mechanism to check whether there was a declaration betweent the decl in question and the comment, but at some point this mechanism fell by the wayside. `Location.last_location` is always `None` nowadays, meaning that the `nextloc` passed into `associate_comment` is always `None`. This PR strips out the `last_location` mechanism in favor of requiring there to be only whitespace between the decl and the next-line comment.

This PR also does a refactor of `associate_comment`, which had some dead code in it and was hard to read. That's factored out into it's own commit; it may be easier to review that commit separately.


src/analysis/ocamldoc.mli
```diff
@@ -0,0 +1,24 @@
...
+(*                                                                        *)
+(*                  Liam Stevenson, Jane Street, New York                 *)
+(*                                                                        *)
+(*   Copyright 2026 Jane Street Group LLC                                 *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
```
voodoos, line 13: We should probably reuse the same header as the implementation, or nothing.

voodoos (commented):
Thanks, I will have a look next week !

src/analysis/ocamldoc.mli
liam923: done

src/analysis/ocamldoc.ml
```diff
@@ -12,49 +12,62 @@
...
-      and cend = cloc.Location.loc_end.Lexing.pos_lnum in
-      let processed =
-        (* It seems 4.02.3 remove ** from doc comment string, but not from
-         * locations.  We can recognize doc comment by checking how the two
-         * differ. *)
-        cloc.Location.loc_end.Lexing.pos_cnum
-        - cloc.Location.loc_start.Lexing.pos_cnum
-        = String.length comment + 5
+let associate_comments ~source ~after_only comments (loc : Location.t) =
+  (* https://ocaml.org/manual/5.5/doccomments.html#s:doc-comments *)
```
xvw, line 16: nitpick: I don't think this particular comment (unlike the others) adds much.

xvw (approved):
Wow, the implementation is much clearer and easier to follow (especially thanks to the explanatory comments). Thank you very much!

voodoos (approved):

PR #2098, opened 2026-07-17
lexer: skip string literals except in interpolation

ikostia:
# Summary

See commit da0ab7d376946f3e3e9131c0963cbd138f9616fa with a test demonstrating the bug this PR fixes, but TL/DR is that ` single document` on the position of a `/` in a string literal `"/usr/bin/home"` would return the division opeartor docs. This is not desired, of course.

It is important to note that deciding that's a correct fix here is tricky:
1. the lexer runs on a pre-ppx source and ocaml itself does not have any string interpolation, so `%{name}` in `"hello %{name}"` is just a bit of a string literal, as per the compiler
2. however, the PPXes do exist and are common, so from a practical perspective it makes sense to try and treat `%{name}` as an interpolation for the purposes of documentation/hover etc

I chose to prioritize 2, so this PR handles interpolation inside string literals and skips only non-interpolated bits of string literals.

# Testing
Existing tests pass + failure demonstration test now tests the correct behavior

### DIFF
--- a/src/kernel/mreader_lexer.ml
+++ b/src/kernel/mreader_lexer.ml
@@ -290,9 +290,8 @@ let reconstruct_identifier_from_tokens tokens pos =
     List.map ~f:fmt (List.filter ~f:before_pos acc)
 
 let reconstruct_identifier config source pos =
-  let state = Lexer_ident.make_state () in
   let rec lex acc lexbuf =
-    let token = Lexer_ident.token state lexbuf in
+    let token = Lexer_ident.token lexbuf in
     let item = (token, lexbuf.Lexing.lex_start_p, lexbuf.Lexing.lex_curr_p) in
     match token with
     | EOF -> item :: acc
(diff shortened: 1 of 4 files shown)


voodoos:
Hi @ikostia, sorry for not coming back to this earlier. We definitely need to fix the unwanted results on type enclosing requests.

In fact, our custom lexing and parsing of identifier is a recurrent source of issues and surprising behaviours. I recently started experimenting with alternative pproaches, but that work is not yet ready and might still take som time. We can try smaller fixes before, and skipping string literals in the lexer does not seem unreasonnable. 

However I don't think introducing more complexity to these already troublesome parts of merlin, to support specific ppxes in an ad-hoc way, is a good thing to do. I wonder if there would be a mean for said ppxes to generate nodes with appropriate locations for merlin to use instead... it migtht be a worse hack, but at least it will be confined to these ppxes.

Issue #2093, opened 2026-07-12
Type-enclosing resolves `*` in tuple types as integer multiplication


smorimoto:
## Summary

When the cursor is placed on the `*` that separates components of a tuple type, Merlin's `type-enclosing` and `document` queries resolve the token as the Stdlib integer multiplication operator. This happens even though the cursor is in a core-type context.

Clients using Merlin, including `ocaml-lsp`, consequently display a misleading hover containing `int -> int -> int` and the documentation for integer multiplication.

This was originally reported as [ocamllabs/vscode-ocaml-platform#1478](https://github.com/ocamllabs/vscode-ocaml-platform/issues/1478).

## Environment

- OCaml 5.5.0
- Merlin 5.8-505
- ocaml-lsp-server 1.27.0 (used to verify the client-visible behaviour)

The relevant logic is also unchanged on the current `main` branch.

## Minimal reproduction

Create `test.mli` with the following contents:

```ocaml
val last_2 : 'a list -> ('a * 'a) option
```

The `*` starts at zero-based column 28. Query the first enclosing type:

```sh
ocamlmerlin single type-enclosing \
  -filename test.mli \
  -position 1:28 \
  -verbosity 0 \
  -index 0 < test.mli | jq '.value[0]'
```

Current result:

```json
{
  "start": { "line": 1, "col": 28 },
  "end": { "line": 1, "col": 29 },
  "type": "int -> int -> int",
  "tail": "no"
}
```

The next enclosing result is the correct tuple type:

```sh
ocamlmerlin single type-enclosing \
  -filename test.mli \
  -position 1:28 \
  -verbosity 0 \
  -index 1 < test.mli | jq '.value[1]'
```

```json
{
  "start": { "line": 1, "col": 25 },
  "end": { "line": 1, "col": 32 },
  "type": "'a * 'a",
  "tail": "no"
}
```

The documentation query at the same position also resolves `*` as the value operator:

```sh
ocamlmerlin single document \
  -filename test.mli \
  -position 1:28 < test.mli | jq -r '.value'
```

```text
Integer multiplication.
    Left-associative operator, see {!Ocaml_operators} for more information.
```

## Expected behaviour

- `type-enclosing` should not return the value-level multiplication operator for a tuple separator in a type context.
- The smallest meaningful enclosing type at this position is `'a * 'a` (columns 25–32).
- `document` should not return the integer multiplication documentation at this position.

## Client-visible behaviour

[...]

PR #2102, opened 2026-07-20
Add a test reproducing #2093

PizieDust:
This PR adds a test reproduction for https://github.com/ocaml/merlin/issues/2093


voodoos (commented):
Thanks ! Can you add a comment in the test saying that this is a FIXME and that the current promoted result is wrong ?

voodoos (approved):

PR #2103, opened 2026-07-20
Destruct open record patterns into closed ones

### LINKED ISSUE #436: Destruct record fields
I was trying to destruct records fields and noticed it wasn't supported. More specifically, with my cursor on the `_` character I `:MerlinDestruct`ed and got `Nothing to do`.

```
type r = {a:int; b:int; c:int; d:int;}
let f {a;_} = a + b + c + d
```
[Note: obviously the actual example is a bit different.]

I expected the pattern to complete. It could either go as `b=_;c=_;d=_;` or (even better) with punning `b;c;d;`.

Does it make sense? Anyone else would find this helpful?

panglesd:
Fixes #436 

Whenever you have a complete match, and your cursor is on an open record pattern, destruct will close the pattern. For instance:

```ocaml
type t = { a : int; b : int; c : int}

let () = match (x : t) with
  | { b = _; _ } -> ()
```

when destructing the second open record pattern, is turned into:

```ocaml
type t = { a : int; b : int; c : int}

let () = match (x : t) with
  | { b = _; a; c } -> ()
```

### DIFF
--- a/src/analysis/destruct.ml
+++ b/src/analysis/destruct.ml
@@ -334,9 +334,10 @@ let rec get_every_pattern loc = function
 let rec destructible patt =
   let open Typedtree in
   match patt.pat_desc with
-  | Tpat_any | Tpat_var _ -> true
+  | Tpat_any | Tpat_var _ -> `Var
+  | Tpat_record (record, Open) -> `Open_record record
   | Tpat_alias (p, _, _, _, _) -> destructible p
-  | _ -> false
+  | _ -> `Not_destructible
 
 let is_package ty =
   match ty.Types.desc with
@@ -717,11 +718,12 @@ let refine_complete_match (type a) parents (patt : a Typedtree.general_pattern)
     config source (patterns : Typedtree.pattern list) =
   match Typedtree.classify_pattern patt with
   | Computation -> raise (Not_allowed "computation pattern")
-  | Value ->
-    if not (destructible patt) then raise Nothing_to_do
-    else
+  | Value -> (
+    match destructible patt with
+    | `Not_destructible -> raise Nothing_to_do
+    | `Var ->
       let ty = patt.Typedtree.pat_type in
-      begin match gen_patterns patt.Typedtree.pat_env ty with
+      begin match gen_patterns patt.pat_env ty with
       | [] -> assert false
       | [ more_precise_pattern ] ->
         (* If only one pattern is generated, then we're only refining the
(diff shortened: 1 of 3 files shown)


voodoos (approved):
It's not a fix, it's a feature !

I am a bit surprised to not see the original example from the issue in your test file... it it also working as expected ?

panglesd:
I went with my own test because it tested a bit more: the provided label is not the first one and is not punned.

But I added it to the test file (before the feature to see the effect of the fix). As well as a test for module qualification.

tests/test-dirs/destruct/destruct-open-records.t
```diff
@@ -0,0 +1,159 @@
...
+        "start": {
+          "line": 4,
+          "col": 6
+        },
+        "end": {
+          "line": 4,
+          "col": 13
+        }
+      },
+      "{ T.a = a; b; c; d }"
```
voodoos, line 156:
```suggestion
      "{ T.a; b; c; d }"
```

🥲

xvw: Yes, but I guess that this is something that can be done by ocamlformat?

src/analysis/destruct.ml
```diff
@@ -334,9 +334,10 @@ let rec get_every_pattern loc = function
 let rec destructible patt =
   let open Typedtree in
   match patt.pat_desc with
-  | Tpat_any | Tpat_var _ -> true
+  | Tpat_any | Tpat_var _ -> `Var
```
xvw, line 337: I know it is nitpicking but since polymorphic variant are not exposed outside of the module, why not having plain regular variant :)

tests/test-dirs/destruct/destruct-open-records.t
panglesd: Yes, ocamlformat does it! Otherwise, I think `PprintAst` is responsible for that. I suggest to do it in another PR if we want to fix that.

src/analysis/destruct.ml
panglesd: I can do that but since plain regular variant are not exposed outside of the module, why not having polymorphic variant :)

More seriously, I see this use of polymorphic variant as "variant type inference". I could explicitly declare the variant type, just as I can explicitly declare the type of a function. But I don't do the latter, so why do the former?

That being said, I'm happy doing the change!

voodoos:
In the tradition of merlin, and the absence of commit from PE in more than twelve hours, let's embrace weaker type disciplines and merge this PR as it is. Thanks !

src/analysis/destruct.ml
xvw: Although this is a minor detail, polymorphic variants result in very slight performance drops (which are likely negligible).

PR #2104, opened 2026-07-21
occurrences: fix a bug where out-of-sync files were never reported

ikostia:
Summary
=======

As part of its [occurrences search](https://github.com/ikostia/merlin/blob/main/src/analysis/occurrences.mli#L3) Merlin computes  the [`occurences_status` ](https://github.com/ikostia/merlin/blob/main/src/frontend/query_protocol.ml#L132), which is intended to give the callers some information about whether the returned list of occurrences is complete or not (see [how `ocaml-lsp` does that](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/ocaml_lsp_server.ml#L464)).

The status itself is [computed](https://github.com/ikostia/merlin/blob/main/src/analysis/occurrences.ml#L391-L395) based on a set of out-of-sync files, which are in turn [computed by the `get_external_locs` function](https://github.com/ikostia/merlin/blob/main/src/analysis/occurrences.ml#L329-L345). The function [calls](https://github.com/ikostia/merlin/blob/main/src/analysis/occurrences.ml#L222) `Stat_check.get_outdated_files stats`, which just [folds `stats.cache`](https://github.com/ikostia/merlin/blob/main/src/analysis/occurrences.ml#L124-L127). The `cache`  is [populated by `Stat_check.check`](https://github.com/ikostia/merlin/blob/main/src/analysis/occurrences.ml#L151-L159), called from [here](https://github.com/ikostia/merlin/blob/main/src/analysis/occurrences.ml#L211).

The bug arises from the fact that the check is populated in the first tuple element, and it is read in the second tuple element. While the tuple evaluation order is not specified, in practice it is right-to-left, meaning that the cache is always empty at the time of out-of-sync-files computation. Therefore the fix is to evaluate the left tuple element in an explicitly preceeding `let` binding.

Testing
======
A new test, demonstrating the bug and the fix.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+unreleased
+==========
+
+  + merlin library
+    - occurrences: fix files modified since the index was built never being
+      reported as out-of-sync.
+
 merlin 5.8
 ==========
 Tue Jun 23 12:15:42 CEST 2026
--- a/src/analysis/occurrences.ml
+++ b/src/analysis/occurrences.ml
@@ -190,7 +190,10 @@ let get_external_locs ~(config : Mconfig.t) ~current_buffer_path uid :
       in
       Option.map external_locs ~f:(fun (index, locs) ->
           let stats = Stat_check.create ~cache_size:128 index in
-          ( Occurrence_set.of_filtered_lid_set locs ~f:(fun lid ->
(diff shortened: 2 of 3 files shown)


voodoos (approved):

voodoos:
Thanks !

PR #2107, opened 2026-07-22
Test quality of life: Add some tooling to better test selection ranges

panglesd:
I cannot read location by eye.

Being able to easily see the output of a query makes us more likely to catch any problem.

### DIFF
--- a/tests/test-dirs/extract_ranges/extract_ranges.ml
+++ b/tests/test-dirs/extract_ranges/extract_ranges.ml
@@ -1,0 +1,67 @@
+let ( let> ) x f = x f
+
+type pos = { line : int; col : int }
+type range = { start : pos; end_ : pos }
+
+let pos_of_json (j : Yojson.Safe.t) =
+  match j with
+  | `Assoc [ ("line", `Int line); ("col", `Int col) ]
+  | `Assoc [ ("col", `Int col); ("line", `Int line) ] ->
+    { line = line - 1; col }
+  | _ -> failwith "wrong pos, expected an object with two fields"
+
+let range_of_json (j : Yojson.Safe.t) =
+  match j with
+  | `Assoc [ ("start", pos_start); ("end", pos_end) ]
+  | `Assoc [ ("end", pos_end); ("start", pos_start) ] ->
+    let start = pos_of_json pos_start in
+    let end_ = pos_of_json pos_end in
+    { start; end_ }
+  | _ -> failwith "wrong range, expected an object with two fields"
+
+let extract range s =
+  s
+  |> Array.mapi (fun i s ->
+      if i < range.start.line then None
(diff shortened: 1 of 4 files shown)


tests/test-dirs/extract_ranges/extract_ranges.ml
```diff
@@ -0,0 +1,67 @@
...
+  | `Assoc [ ("start", pos_start); ("end", pos_end) ]
+  | `Assoc [ ("end", pos_end); ("start", pos_start) ] ->
+    let start = pos_of_json pos_start in
+    let end_ = pos_of_json pos_end in
+    { start; end_ }
+  | _ -> failwith "wrong range, expected an object with two fields"
+
+let extract range s =
+  s
+  |> Array.mapi (fun i s ->
```
voodoos, line 24: That's a bit wasteful right? We could start the iteration at the expected line... and stop at the last one. I mean, we know statically the lines we need, I think it would actually make this function more readable ?

But I mean, it works, and we really don't care in this setting.

tests/test-dirs/extract_ranges/extract_ranges.ml
```diff
@@ -0,0 +1,67 @@
...
+      if i < range.start.line then None
+      else if i = range.start.line && range.start.line = range.end_.line then
+        let spaces = String.init range.start.col (fun _ -> ' ') in
+        let extracted =
+          String.sub s range.start.col (range.end_.col - range.start.col)
+        in
+        Some (spaces ^ extracted)
```
voodoos: I wonder it it would be worth it to materialize the eventual whitespaces at the beginning and end of the selected ranges. Having some would likely be an issue.

voodoos (approved):
That's cool ! Weird extract function though.

tests/test-dirs/extract_ranges/extract_ranges.ml
panglesd: Yes, this originates from before where I was adding empty lines before/after so it would always have the same height, so when you scrolled very fast you could animate the ranges appearing, like a flipbook! :)
But the compact form felt better.

tests/test-dirs/extract_ranges/extract_ranges.ml
panglesd: Very good point. I'll switch to another character than space.

panglesd: I used `···` around the included range. (It felt more readable to use only three dots rather than the full line filled with dots, and was more symmetric with the separation at the end).

voodoos: I was thinking of replacing the eventual leading and trailing whitespaces (with `•`s for example), but your version works too.

PR #2109, opened 2026-07-25
test: cover stale load paths across typer states

rgrinberg:
This extends the existing `cmi-appears.t` regression test to cover two typechecker states that share the same build directory.

Both states first cache `Bar` as missing. After `bar.cmi` appears, querying the first state refreshes the shared directory-content cache. The second state then incorrectly treats its stale local load-path snapshot as current and continues to report `Unbound module Bar`.

This reproduces ocaml/ocaml-lsp#1209 and demonstrates the multi-state case not covered by #2062. The new assertion currently fails with `Unbound module Bar`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@ unreleased
       submodule no longer hides the opened module (fixes #1748)
     - Fix occurrences staleness detection when the server is not running at the
       project's source root. (#2097)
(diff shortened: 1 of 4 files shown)


voodoos:
Thanks @rgrinberg, that is very helpful ! I am going to look for a fix.

PR #2110, opened 2026-07-27
Fix index staleness in presence of ppxes

liam923:
The `occurrences` query's ability to detect staleness in the presence of ppxes is broken. To determine staleness, ocaml-index records the digest of the source file. However, when the file fed into ocaml-index is a file like `foo.pp.ml`, ocaml-index incorrectly uses the digest of `foo.pp.ml` rather than `foo.ml`. (Note that we want to show an occurrence as stale if `foo.ml` has changed, not if `foo.pp.ml` has changed.)

This PR is split into two commits. The first commit adds a test that demonstrates the bug, and the second fixes it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@ unreleased
       submodule no longer hides the opened module (fixes #1748)
     - Fix occurrences staleness detection when the server is not running at the
       project's source root. (#2097)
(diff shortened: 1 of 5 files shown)


voodoos (approved):
Thanks @liam923 !

PR #2115, opened 2026-08-05
occurrences: fix a bug with out-of-sync file detection when src files don't change sizes

ikostia:
# Summary

Prior to this change,`find_linked_uids` did not report out-of-sync files (unlike `get_external_locs`), despite observing them (`log ~title "Found clashing idents %S <> %S. Ignoring UID %a."`). This PR exposes this notion of a name clash (which may happen when a rename has just happened, but the file hasn't been rebuilt yet) in the return type of `find_linked_uids`: now the fn returns both the related UIDs *and* the files with clashes. The consumer fn `locs_of` just joins `clashing_files` with the out-of-sync files returned by `get_external_locs`.

# Testing
`r-stale-after-rename.t` demonstrates how incompleteness-related log messages show up when something has been renamed by not rebuilt.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,13 @@
+unreleased
+==========
+
+  + merlin library
+    - occurrences: when looking for the occurrences of an identifier that was
+      renamed since the project was last built, report the files whose build
+      artifacts are out-of-sync with the sources in the status of the query,
+      instead of silently returning incomplete results. Clients performing a
+      rename can now detect that the occurrences of these files are missing.
+
 merlin 5.8.1
 ============
 Fri Jul 31 11:31:42 CEST 2026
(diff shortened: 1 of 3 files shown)


panglesd (commented):
Sorry for the delayed review!

It seems to me the root cause of the problem is that the check in `Stat_check` is just a check of the size. Checking the names match seems a workaround to partially fixing this issue. I would be happier with an implementation of what is suggested in the comment:

```ocaml
        let equal =
          (* This is fast but approximative. A better option would be to check
               [mtime] and then [source_digest] if the times differ. *)
          Int.equal stats.st_size size
        in
```

(but I might be lacking context on why the current solution was chosen!)

An example of the issue of the current fix, is that is does not detect staleness when nor the name nor the size change. For instance, the same test with, instead of a `foo -> bar` rename, going from `foo + 1` to `1 + foo` fails to detect the staleness. Demo: https://github.com/panglesd/merlin/commit/d95e485f4c377ec22ecc24712b83ab4c8c5eb62a

Issue #1162, opened 2020-08-28
Feature request: destruct to "let" instead of "match" when there is only one case


emillon:
Hi,

I have a tiny request regarding Destruct: in the case there's only one branch (for example in the case of a record type), I think it would make sense to emit a "let" instead of a single-branch "match".

Here's what I mean. For the corresponding code (and cursor):

```ocaml
type t = {x:int; y:int}

let f (r:t) =
  r|
```

Merlin does:

```ocaml
match r with | { x; y } -> (??)
```

I'm suggesting that in does instead:

```ocaml
let { x; y } = r in (??)
```

What do you think?

Thanks!

cc @voodoos

trefis:
Sounds reasonable to me.

trefis:
Actually, I think I'd rather have this as a separate rewriting that'd toggle between match and let when there's only one clause.
Currently in your example, once I'm at that step:
```ocaml
match r with | { x; y } -> (??)
```
and call destruct again, on `x` or `y`, I get a second clause, something like this:
```ocaml
match r with
| { x = 0; y }
| { x = _; y } -> (??)
```

So it seems simpler (and more regular) to me to always generate matches from destruct.

emillon:
That's a good point, so I agree that it's better to start as a separate rewrite.

To explore the problem, I wonder if it would be possible to keep the "best" syntax at each step, e.g.:

    r|

->

    let {x|; y} = r in
    (??)

->

    match r with
    | {x = 0; y}
    | {x = _; y} -> (??)

As you showed, there are cases where a single branch match gets rewritten into a multi-branch match. Are there features in merlin that would do the opposite? Something that collapses identical branches together maybe?

PR #2117, opened 2026-08-17
Allow destruction in let bindings' pattern

panglesd:
Before, it was impossible to obliterate a pattern inside a let-binding (unless the let-binding was turned into a match internally of course)

```ocaml
type t = { a : int; b : int}
let v a b = { a; b }

(* Calling destruct on the _ below would fail *)
let _ : t = v 1 2
```

As far as I understand, there is no reason for this limitation, it was just not implemented. This PR fixes that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+unreleased
+==========
+
+  + merlin library
+    - destruct: allow to destruct let-bindings's patterns (#2117)
+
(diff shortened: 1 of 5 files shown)


panglesd:
Related to: #1162

voodoos (approved):
Thanks @panglesd, that's cool ! The changes are surprisingly contained, and feel reasonable.

PR #2118, opened 2026-08-24
Fix locate when cursor is on another declaration

aryemans:
Fixes #2033.

When `locate` is invoked with an explicit identifier while the cursor is on a different declaration, context detection currently treats the cursor position as the definition point and returns `Already at definition point`.

This changes declaration context detection so that it only reports being at the origin when the declaration under the cursor has the same name as the requested identifier. Otherwise, lookup continues normally.

The fix covers value and type declarations and adds a regression test for the reported case.

Tests:
- `dune runtest tests/test-dirs/locate/issue2033.t`
- `dune build @all-locate-tests`
- `dune runtest`

### DIFF
--- a/src/analysis/context.ml
+++ b/src/analysis/context.ml
@@ -131,22 +131,34 @@ let inspect_browse_tree ?disambiguate ~cursor lid browse : t option =
     let node = Browse_tree.of_browse enclosings in
     log ~title:"inspect_context" "current enclosing node is: %s"
       (string_of_node node.Browse_tree.t_node);
-    match node.Browse_tree.t_node with
-    | Pattern p -> inspect_pattern ~cursor ~lid p
-    | Value_description _
-    | Type_declaration _
-    | Extension_constructor _
-    | Module_binding_name _
-    | Module_declaration_name _
-    | Label_declaration _
-    | Constructor_declaration _ -> None
-    | Module_expr _ | Open_description _ -> Some Module_path
-    | Module_type _ -> Some Module_type
-    | Core_type { ctyp_desc = Ttyp_package _; _ } -> Some Module_type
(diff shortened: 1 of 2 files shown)


panglesd (commented):
Thanks for the PR!

I feel the logic should be in `locate.ml`, as `inspect_browse_tree` can be used in other contexts (currently only `type_enclosing`) where an equality between the lid and declaration name could be interpreted differently.
I feel the semantics of "`Some Unknown`" is "we are probably in a context represented by a `Context.t` but leaf_node is not deep enough in the typedtree to know which context. `None` is "we are in no context represented by `Context.t`".
So I think `inspect_browse_tree` should keep this semantic, and the logic be put in `locate.ml` near the call to `inspect_browse_tree`. If we are looking for a lid different from the one at the cursor, we return the same namespaces as an unknown context, otherwise we use the context.

src/analysis/context.ml
```diff
@@ -131,22 +131,34 @@ let inspect_browse_tree ?disambiguate ~cursor lid browse : t option =
...
+      match node.Browse_tree.t_node with
+      | Value_description vd -> Some vd.val_name.txt
+      | Type_declaration td -> Some td.typ_name.txt
+      | _ -> None
+    in
+
+    match declared_name with
+    | Some name when name = Longident.last lid -> None
+    | Some _ -> Some Unknown
+    | None -> (
```
voodoos, line 145: The `inspect_browse_tree` function should always return information about the context around the cursor, it is not the correct place to have this kind of special cases.

Context detection should be kept as generic as possible, but we can have it convey more information if that is need downstream.

voodoos (changes requested):
I agree with @panglesd that context detection is not an appropriate place for that fix.

PR #2119, opened 2026-09-07
Clarify intention in an ocaml-index test.

Tim-ats-d:
cc @voodoos


voodoos (approved):

voodoos:
Thanks ! For context: in the test we were overwriting one of the input indexes during aggregation, which is not correct anymore with pointing indexes.

PR #2120, opened 2026-09-07
Fix "merlin_loc" in constrained expression.

panglesd:
In Merlin there is a notion of "extended loc", that extends the normal loc in different scenarios. For instance, the loc of x is extended to the following ranges:
- To account for the space between "in" and the expression:
   ```ocaml
   let _ = 1 in[ x]
   ```
   (And similarly in many cases, as ifthenelse, while, ::, ...)

- To account for the lack of a node for "extra constraints":
   ```ocaml
   let a = [( x : int)]
   ```
   And similarly for coercion, pattern constraints and coercion

When in the two situation at the same time, merlin currently drops the first case:
```ocaml
let _ = 1 in [(x : int)]
```
instead of 
```ocaml
let _ = 1 in[ (x : int)]
```

This PR fixes it. It also introduce a small test helper to see where your cursor is, because line and character counting is something I am happy to automate.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@ unreleased
 
   + merlin library
     - destruct: allow to destruct let-bindings's patterns (#2117)
+    - Fix detection of expression at cursor in case of type-constrained
+      expression (#2120)
 
 merlin 5.8.1
 ============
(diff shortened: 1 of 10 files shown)


voodoos (approved):
Looks good to me

voodoos:
Need a changelog entry

PR #2121, opened 2026-09-07
Improve enclosing ("Selection Range" in the LSP jargon)

panglesd:
Made with @PizieDust.

Drastically improve the enclosing request.

## Going down-right

It used to follow exactly the `Mbrowse.node` hierarchy. But we don't always want to select "whole nodes", as for instance in:

```ocaml
let () =
  let x = 1 in
  let y = 1 in
  let z = 1 in
  let t = 1 in
  x, y, z, t
```
we don't want to go from
```ocaml
let () =
  let x = [1] in
  let y = 1 in
  let z = 1 in
  let t = 1 in
  x, y, z, t
```
to:
```ocaml
let () =
  [let x = 1 in
  let y = 1 in
  let z = 1 in
  let t = 1 in
  x, y, z, t]
```
although this is the parent expression. It would be downright bad!

Instead, the new implementation goes "down-right" on expression nodes when it comes from the left: 
```ocaml
let () =
  [let x = 1 in]
  let y = 1 in
  let z = 1 in
  let t = 1 in
  x, y, z, t
```
then
```ocaml
let () =
  [let x = 1 in
  let y = 1 in]
  let z = 1 in
  let t = 1 in
  x, y, z, t
```
then
```ocaml
let () =
  [let x = 1 in
  let y = 1 in
  let z = 1 in]
  let t = 1 in
  x, y, z, t
```
...

And I say down-right, because the (virtual) AST looks like that:

downright

## Better locations

This PR improves locations by adding a merlin loc to sequences (`expr1 ; expr2`). It is also careful with which kind of locations to consider.

## Care for constraints

The following expression `((((x : int) : int) : int) : int)` is a single expression node... the constraints (and their location) being in the `exp_extra` field of the expression
This PR makes different enclosing for each layer.

# Conclusion

I can clean the history, but it should be ok to review it all at once (not commit by commit) since we went in several wrong directions, but at then end it is not too big.

:guitar:


panglesd:
There is an issue with `begin ... end` and `( ... )` as in the typedtree, apart from locations, there is no distinction between:
```ocaml
let () = 
  f ();
  ( f ();
    f () )
```
and
```ocaml
let () = 
  f ();
  f ();
  f ()
```

As a consequence, we might include `(` without the closing one, eg as in:

```ocaml
let () = 
  [f ();
  ( f ();]
    f () )
```

I'll investigate, but I'm not sure how accessible those locations are... I would say this PR is already mergeable as is. I'll include the problem as a FIXME in the test.

panglesd:
I've reworked the history to make it easier to review, since the last commits introduced a few "special cases".

The first commit is the core of the rewrite of `enclosing.ml`. It is hopefully relatively generic and easy to review. The commit also contains many tests, some of which are `FIXME`s, mainly due to corner cases in the locations:
- Sequences are missing their `merlin_loc`
- Parenthesized expression/pattern/coretype only have their parenthesized location

The next commits fix those FIXME one by one, and each of them is fairly small.

panglesd:
CI is green, this is ready for review!

panglesd:
For the last 4 commits: I made an AI review, added the relevant findings as tests, and fixed them.

PizieDust:
Thanks @panglesd 
I tested it and everything seems to work well. Thank you for the tests.

src/analysis/enclosing.ml
```diff
@@ -0,0 +1,201 @@
+open Std
```
voodoos, line 26: Let's make a MLI with a nice description of the intent for this module, and how it would benefit from a rework using the parsetree directly.

voodoos (commented):
This looks like a good improvement indeed. 

It is, as always, a bit saddening to see the accumulation of fixes needed to circumvent the inherent limitations of working on the Typedtree and not the (non-ppxed-)Parsetree.

But let's keep that bigger change for next time.

Can you add an mli for the new module you're introducing ? 

Also there are now two competing definitions of what is an "enclosing". In the `type-enclosing` command (and `enclosing` before this PR) an enclosing is a node from the Browse_tree. This PR introduces a more complex definition for the `enclosing` command, which introduces more intermediate nodes. We are probably goign to keep the command naming as it is today, but then we should document clearly these behaviour differences in the protocol docs: `doc/dev/PROTOCOL.md`.

src/analysis/enclosing.ml
panglesd: Do I need to add a copyright header?

panglesd: I assumed not (I did not add one for the implementation either)

panglesd:
Thanks for the review!

I added an mli and some docs (and rebased on main).

I agree that using the parsetree would remove several of the corner cases of using the typedtree, which often remove some information that should not be needed at this stage of the compilation.

I tried to leave a list of the things the switch to the parsetree would simplify, to encourage the rewrite. But I think this can be merged before. I would be happy to tackle the parsetree work needed for that though.

PR #2127, opened 2026-09-16
Add `PPX_DEPS` to merlin configuration to allow `USE_PPX_CACHE` to work when ppx have deps

panglesd:
Everything is in the long title.

Merlin caches PPX output: if the parsetree did not change, and the ppx/ppx options did not change, it can reuse the parsetree output by the PPX last time instead of rerunning it.

However, some ppx have dependencies, which were not taken into account by the cache mechanism.

This PR adds:
- A `PPX_DEPS` field to .merlin files
- That are then taken into account when computing the key for ppx cache.


panglesd:
Note that I am going to do the dune side soon, and it might make sense to wait for it to be done and tested in dune condition before merging or even reviewing this.

voodoos:
There might be more than one ppx though... should we try to handle these cases ?

panglesd:
Dune work can be tracked in [this branch](https://github.com/panglesd/dune/tree/enable-ppx-cache) (close to completion, in the sense that it compiles but I haven't tested it yet :slightly_smiling_face:)

I think that as a first step we can consider that there is only one PPX, any PPX dependency invalidates the whole ppx phase. I think it is already reasonably accurate, since ppxlib merges all registered ppx in one. But I'm not sure! I'll investigate that.

panglesd:
I was able to test the feature in a real-life scenario (in my editor!), using the [dune branch](https://github.com/panglesd/dune/tree/enable-ppx-cache) mentioned above.

That uncovered a preexisting bug in the ppx cache (where `Filename.concat` was used, but contrary to what I would expect coming from `Fpath`, it does not check whether if the second path is absolute or relative...). Fixed in the last commit (I'll see if there is a more idiomatic way to do this in other part of the codebase...)

And, now it works!

panglesd:
Rebased on top of #2131.

panglesd:
Hopefully, CI is green and this is ready for (re-)review!

voodoos (approved):
Looks good to me, thanks !

PR #2128, opened 2026-09-17
Fix outdated merl-an references in CI and Dockerfile.

Tim-ats-d:
We were still pointing to Xavier's fork.

cc @voodoos


panglesd (approved):
Checked, looks good to me.

voodoos:
Thanks !

PR #2129, opened 2026-09-22
test: demonstrate over-sharing with reader cache

Alizter:
Demonstrate an issue where the reader cache fails to be invalidated when file contents are the same but "kind" differs.

Found when reviewing 
- https://github.com/ocaml/dune/pull/16460


panglesd:
Thanks a lot! The issue can also be triggered when a Merlin config that affects parsing is changed. I'll open a PR to fix that, on top of yours since I don't have the rights to merge it right away.

panglesd:
Merged in #2131 !

PR #2131, opened 2026-09-23
Fix oversharing cache

### LINKED ISSUE #2130: Reader cache not invalidated for different parser kind
Reproduction case in 
- #2129 

The reader cache is only keyed by the file contents and the completion position. This means that you can make an invalid `a.ml` file, get an error and then any other file or even mli will not report an error, if it has the same contents.

panglesd:
Use the merlin config in the parsetree cache fingerprint, so that when the config change, the cache is invalidated. Not everything from the config affects the parsing, from what I saw only fields like `reader`, `extension`, `extension_to_reader`, and `warnings` do affect it. But it felt safer and easier maintenance to use the whole config as fingerprint.

Fix #2130

Based on #2129, thanks @Alizter for reporting and adding the test.


panglesd:
Hmm let me investigate the red CI

panglesd:
Okay, I always forget you need to add `MERLIN_TEST=all` to run all tests. Promoted with a mention.

Let me know if you think I should only fingerprint the config relevant for parsing (in which case, I should check a bit better exactly which one are used). (Since the over-approximation on options used for parsing fingerprint is why the test had to be updated.

voodoos (approved):
I thinks it's all right to use Cache.key for now. We can always refine it later if needed.

PR #2133, opened 2026-09-23
Repro for Issue 1980 (#2017 rebased)

panglesd:
#2017 rebased and with added clarity in location reading


voodoos:
This test should be disabled on windows

panglesd:
Oups! Sorry you had to repeat the same comment.

panglesd:
(I found `dune`'s way of avoiding the rebase conflicts on changelog entries quite nice. It also avoids silently merging a changelog entry into an old release changelog. But merlin's volume of PR probably does not justify it)

